PR #3, opened 2025-05-31
add ci config

likewhatevs:
Add a ci config.

Some traits of this which IMO make it nice:
1) It retries wrt/ deps.
2) Deps are minimal.
3) Caching is used.

Runtime is about 40 seconds once cache populates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,28 +8,26 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     steps:
-      - name: Install Deps.
-        uses: cytopia/shell-command-retry-action@v0.1.6
-        with:
-          retries: 10
-          pause: 20
-          command: |
-            export DEBIAN_FRONTEND=noninteractive &&
-            sudo apt-get install -y rustup linux-tools-common libelf-dev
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+    push:
+    pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Install Deps.
+        uses: cytopia/shell-command-retry-action@v0.1.6
```
josefbacik, line 12: I'd rather not do retry, especially for build, it'll just slow down when there's an actual problem. People can always re-kick if they believe there was a transient error.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build
+        uses: cytopia/shell-command-retry-action@v0.1.6
+        with:
+          retries: 3
+          pause: 3
+          command: |
+            cargo build --release
```
josefbacik, line 34: I've started adding tests, so a cargo test step would be lovely as well.

josefbacik (changes requested):

PR #15, opened 2025-07-15
systing: add process command line

yaakov-stein:
This adds the command line arguments for processes by introducing the `ProcessTree`, which syncs with the `ProcessDescriptor` based on `pid`. Although `sysinfo` does have this capabilities, using it made the trace unreadable as the majority of the trace was filled with `systing`. In general, it's state wasn't needed and `fb_procfs` was lighter.

This feature was recently added to `scxtop trace`, so I figured I'd make a PR for it here as well.

Here are some examples of what the command line will look like in the UI:
Screenshot 2025-07-15 at 6 37 21 PM
Screenshot 2025-07-15 at 6 37 50 PM
Screenshot 2025-07-15 at 6 38 02 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
+name = "ahash"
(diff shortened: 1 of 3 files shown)


yaakov-stein:
> Please fix the clippy errors.

Looks like they're coming from a different file (`build.rs`) that wasn't changed by this PR. It should be fixed with #16.

PR #122, opened 2026-06-03
Bound stack recorder memory under short-lived-process churn

josefbacik:
## Problem

Fleet machines running systing alongside CI (thousands of parallel short-lived compiles) OOM the 4 GiB cgroup limit. Reproduced with a churn workload (64 spawners, ~37k processes in a 30s trace) under a hard `memory.max=4GiB`: the old binary peaked at **3.84 GiB unreclaimable** — at the edge of the limit; slightly heavier churn kills it.

## Root causes

Measured via per-phase `memory.stat` breakdown of the recording cgroup:

1. **Recording phase**: the `(Stack, tgid) → stack_id` dedup map holds every unique stack's full address vectors in memory for the whole trace. ASLR makes user addresses distinct per process, so under churn most samples are unique (~1.2M unique stacks for 1.8M samples) and the map grows unboundedly.
2. **Symbolization phase** (the actual peak): the `(addr, tgid)` string cache stored a formatted hex string per address even for exited processes — where symbolization can never succeed — and blazesym retains parsed DWARF for every live binary it touches with no eviction (GiBs when debug-heavy binaries like rustc/rust-analyzer are running).

## Changes

- Keep only a 128-bit content hash (random per-run keys) per unique stack in memory; spill stack contents to an unlinked tempfile in the output dir and stream them back at symbolization. Periodic flush (every 256 records) plus an in-memory pending window means an IO error loses nothing: durable records replay from disk, the rest fall back to memory — no dangling stack ids.
- Two-pass symbolization: dead-process stacks stream straight from the spill (raw-hex user frames + shared kernel cache, freed immediately); live-process stacks are grouped per tgid with a per-group user cache. A memory valve checks anon RSS against a budget derived from the cgroup limit and rebuilds the symbolizer between/within groups, with a growth floor to prevent re-parse thrashing.
- Stack parquet rows flush every 20k records instead of 200k (each row carries ~25 frame-name strings).

## Results (A/B, 64-spawner churn, hard 4 GiB limit)

| | old | new |
|---|---|---|
| peak unreclaimable (anon+kernel) | 3.84 GiB | **1.94 GiB** |
| cgroup peak | 3.96 GiB | 2.12 GiB |
| runtime | 78s | 68s |
| missed events | 22,584 stack | 22,340 stack |

Output validation: zero dangling `stack_sample` references, zero duplicate stack ids, identical table shapes and frame counts; Python stack symbolization unaffected.

## Testing

- 345 unit tests pass (3 new: spill round-trip, periodic-flush durability accounting, dedup-hash properties)
- `trace_validation` integration suite: 31/31 pass against the final binary
- Adversarial review of the spill format, error paths, and borrow lifetimes; findings (flush-failure data loss, write-error blast radius, valve thrash, fixed-key hashing) all addressed

Patch version bumped 1.8.1 → 1.8.2 (no schema change).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3333,7 +3333,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.8.1"
+version = "1.8.2"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "systing"
-version = "1.8.1"
+version = "1.8.2"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/duckdb.rs
+++ b/src/duckdb.rs
@@ -15,8 +15,9 @@ use crate::parquet_paths::ParquetPaths;
 /// DuckDB sizes its buffer pool from physical RAM, which inside a container is
 /// the host's RAM, not the cgroup limit. On large hosts this makes the
 /// parquet→duckdb import OOM the container. We read the cgroup limit ourselves
-/// so we can cap DuckDB explicitly.
-fn detect_cgroup_memory_limit() -> Option<u64> {
+/// so we can cap DuckDB explicitly. Also used by the stack symbolizer to size
+/// its memory valve.
+pub(crate) fn detect_cgroup_memory_limit() -> Option<u64> {
     // A limit at or above this is "effectively unlimited" (cgroup v1 reports a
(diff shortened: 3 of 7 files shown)


josefbacik:
Second commit extends the same treatment to the memory recorder: `--memory`/`--memory-alloc` interned a stack per malloc site per process into its own in-memory `(Stack, tgid)` map, with the same unbounded growth under churn. The dedup+spill machinery is now a shared `StackInterner` used by both recorders; the trace-end handoff moves the whole interner (disjoint id ranges, so no collisions) and the stack recorder replays every spill through the two-pass symbolization.

Verified: 12k alloc + 2.6k fault rows with zero dangling stack refs and symbolized live-process alloc stacks; 346 unit tests and the memory_recorder integration suite (5/5) pass.

PR #137, opened 2026-07-02
Add contextual labels and gVisor island bridging for unresolvable frames

### LINKED ISSUE #136: Symbolization: user frames from gVisor (runsc) sandboxed processes mostly render as raw hex
Stack samples from processes running inside gVisor (runsc) sandboxes largely fail user-frame
symbolization: application frames render as bare `0x...` hex in both trace and `--continuous`
modes. Host-side processes on the same machine symbolize fine, and the same binaries symbolize
fine when run unsandboxed. On a gVisor-based Kubernetes cluster we observe the majority of
sandbox-attributed CPU samples with mostly-unresolved stacks; kernel frames are unaffected.

This issue maps out exactly where the failures come from (verified against gvisor @ HEAD and on
a live sandbox with `runsc release-20260622.0`, systrap platform) and proposes a phased fix.
The picture is more nuanced than "guest memory is anonymous" — several distinct classes with
different fixes:

## How a gVisor guest looks to the host-side symbolizer

Under the systrap platform (the default), guest code executes natively in *stub* processes
whose address spaces mirror the guest's: the Sentry mmaps guest ranges at the guest virtual
addresses (`pkg/sentry/platform/systrap/subprocess.go` `MapFile`, `MAP_FIXED|MAP_SHARED`). So:

- Sampled user addresses in stub threads ARE guest VAs, and frame-pointer unwinding works.
- `bpf_get_current_pid_tgid()` yields the stub's host tgid; symbolization correctly targets the
  stub's `/proc/<tgid>/maps`.
- Guest process names never appear in host comms: nothing calls PR_SET_NAME host-side, stubs
  fork without exec'ing, so all sandbox CPU aggregates under the runtime's own identity —
  comm `exe` with current releases (the sandbox re-execs `/proc/self/exe`; verified on
  release-20260622.0), `runsc` with some older/vendor builds. Worth knowing when reading
  per-comm profiles.

What backs the stub's mappings is the crux. Actual maps of a stub running a static busybox
guest (runsc release-20260622.0):

```
64000-69000    r-xs 3ff26000 00:01 4    /memfd:runsc-memory (deleted)   <- usertrap trampolines
400000-4c2000  r-xs 00000000 00:13 426  /root/bin/guestbox              <- guest text, file-backed
4c2000-4c3000  r-xs 3ff1f000 00:01 4    /memfd:runsc-memory (deleted)   <- 4KiB patched island
4c3000-4cd000  r-xs 000c3000 00:13 426  /root/bin/guestbox
4cd000-4ce000  r-xs 3ff20000 00:01 4    /memfd:runsc-memory (deleted)   <- island
4ce000-4d5000  r-xs 000ce000 00:13 426  /root/bin/guestbox
...
7fa398efc000-7fa398eff000 r-xp 00000000 00:00 0                         <- stub/sysmsg text (anon)
```

1. **Guest ELF text is mostly host-file-backed.** The gofer donates host FDs for regular files;
   private file mappings only leave the host file on write (`pkg/sentry/mm/pma.go` COW). The
   symbolic path (`/root/bin/guestbox`) is relative to the gofer's private mount namespace and
   useless on the host — but blazesym's default `map_files: true` opens the mapping through
   `/proc/<pid>/map_files/<range>`, which works regardless of namespaces (verified: returns the
   real ELF). **So plain guest text frames already resolve today** while the stub is alive.

2. **systrap's syscall patching punches unresolvable holes exactly where samples land.**
   `pkg/sentry/platform/systrap/usertrap` lazily rewrites executed syscall sites in guest text;
   the write COW-breaks precisely those 4KiB pages into the Sentry's `runsc-memory` memfd
   (executable vmas copy only the required pages). Result: a file/memfd mosaic where the memfd
   islands sit at *hot syscall sites* — over-represented among CPU samples of syscall-heavy
   workloads. Island addresses still correspond to functions of the original binary; only the
   backing changed. (Contrast: the ptrace platform shows a single contiguous file-backed text
   mapping — no islands.)
   Hazard note: an island's `map_files` entry opens the pool memfd itself; its offset-0 bytes
   can coincidentally parse as an ELF header (e.g. a cached copy of some binary's first page).
   The maps entries' large memfd offsets then fail `file_offset_to_virt_offset`, so today this
   degrades to "unknown" — but a small-offset island could in principle resolve to a *wrong*
   symbol.

3. **Sandbox runtime execution is structurally unsymbolizable and unlabeled.** Every guest
   syscall/fault executes trampoline + sysmsg handler code on the stub thread, in memfd/anon
   regions (the guest-side `/proc/<pid>/maps` even labels the trampoline region `[usertrap]`).
   For syscall-heavy guests a large share of "guest" CPU samples land here and can never
   resolve to an application binary — today they render as meaningless hex, indistinguishable
   from missing symbols. That's not a symbolization failure so much as *unattributed sandbox
   overhead*, and labeling it would make gVisor's CPU cost directly measurable in profiles.

4. **Exited stubs**: sandboxed workloads that die before end-of-trace hit the existing
   dead-process path (user frames rendered as hex unconditionally). Sandboxing platforms often
   run short-lived episodic work, so this class is over-represented there — but it's a general
   systing gap, probably best tracked separately.

Two adjacent classes worth naming because they dominate real gVisor-heavy profiles and
compose with the above:

5. **JIT code**: for JIT runtimes in sandboxes (node/V8, JVM, CPython 3.14's copy-and-patch
   JIT), the engine's own text resolves fine per (1) — but JIT output lives in anonymous exec
   pages and renders as hex. Not gVisor-specific, but gVisor makes the standard mitigations
   harder: perf-map files (`/tmp/perf-<pid>.map`, which blazesym would consult) are written
   inside the *guest* filesystem under the *guest* pid, so the host-side symbolizer never
   finds them. In practice JIT pages can be the single largest unresolved term on such hosts.
   Deserves its own issue (jitdump/perf-map translation + labeling); noted here for scope.

6. **The Sentry itself may not resolve**: release runsc binaries can ship without an ELF
   symbol table, in which case the (correctly identified) `runsc` module yields no symbols —
   Go's `.gopclntab` is present but blazesym doesn't consume it. Module-level labeling (below)
   at least attributes these frames to the sandbox runtime.

## Proposed fixes (phased)

**P0 — classify instead of bare hex.** One mechanism covers all the classes above: when
symbolization fails, attach the best-known context instead of rendering `0x...`:
- Detect gVisor tgids cheaply (`/proc/<tgid>/exe` → runsc binary, and/or maps containing
  `memfd:runsc-memory`; stub text mappings are also distinctively `r-xs`). Frames in
  stub/trampoline regions → `[gvisor:runtime]`; frames in the runsc module without symbols →
  `[gvisor:sentry]`; otherwise-unresolvable guest frames → `[gvisor:guest] 0x...`.
- Generic labels from the same pass, not gVisor-specific: anon-exec frames in a process whose
  maps contain a known JIT runtime → `[jit:node]`/`[jit:jvm]`/`[jit:python]`; exited-process
  frames → `[exited] 0x...`; module-known-symbol-missing → `unknown (<module>)`.
- Surfacing blazesym's failure `Reason` (currently discarded in `symbolize_stack_frames`)
  gives most of the discrimination for free: "no mapping" / "unparseable member" / "bad file
  offset" / "symbol not found" are exactly the class boundaries.
This turns an opaque hex wall into measurable categories — notably, `[gvisor:runtime]` +
`[gvisor:sentry]` make the sandbox's own CPU overhead directly visible in profiles, which is
arguably more valuable than recovering any individual frame.

**P1 — bridge the patched islands.** For flagged tgids, parse maps once; for each executable
memfd/anon island whose file-backed neighbors reference the same file with congruent offsets
(`R.offset == L.offset + (R.start - L.start)`), symbolize island addresses against the
neighbor's file: `Source::Elf` + `Input::FileOffset(addr - L.start + L.offset)`, opening via
the neighbor's `map_files` link. Conservative (both neighbors must agree), pure host-side, no
gVisor cooperation, and it also sidesteps the wrong-symbol hazard above. Longer-term this
could live in blazesym as a maps-coalescing feature; doing it in systing first keeps it
self-contained against the pinned version.

**P2 — guest-maps-based symbolization** for what remains (wholly-memfd text, e.g. anything
copied up in the Sentry's overlay; JIT regions via guest perf-map files; plus guest comm/pid
attribution). gVisor already assembles everything needed — `containerManager.ProcfsDump`
returns per-process guest maps (path, offset, perms), comm, exe and start time
(`runsc/boot/procfs/dump.go`) — but no stock CLI exposes it; `runsc exec <ctr> cat
/proc/<pid>/maps` works today (at the cost of running a process inside the sandbox, which
profilers shouldn't do routinely). Sketch: enumerate sandboxes from the runsc state dir,
pull guest maps per process, correlate host stub tgid ↔ guest process (stub start-time ≈ guest
exec time; fall back to matching sampled VAs against guest mm ranges — fork-siblings collide
but symbolize identically), resolve guest paths against the container rootfs (OCI bundle
`root.path`) or by build-id, symbolize by VA identity. A tiny upstream gVisor change adding
`runsc debug --procfs` would make this clean (the RPC exists; only CLI plumbing is missing).
This is also the unlock for pystacks inside sandboxes: discovery just needs the guest maps and
binaries; the BPF walker already reads the right addresses since stub VAs are guest VAs.

**Non-goals here:** VM-based runtimes (Firecracker, Kata, QEMU-class) — guests there are
architecturally invisible to host BPF; symbolization for those belongs to an in-guest agent,
and the host side can at most label the VMM threads. The dead-process gap (4) and the JIT
story (5) generalize beyond sandboxes — each is worth its own issue (maps snapshot at
first-sample time; jitdump/perf-map support incl. guest-fs translation).

Happy to send a prototype PR for P0+P1.

## Repro

Any Linux host with runsc installed (release-20260622.0 used here):

```
mkdir -p /tmp/ctr/rootfs/bin && cp $(which busybox) /tmp/ctr/rootfs/bin/busybox
ln -s busybox /tmp/ctr/rootfs/bin/sh
cd /tmp/ctr && runsc spec -- /bin/sh -c 'i=0; while true; do i=$((i+1)); done'
sed -i 's/"terminal": true/"terminal": false/' config.json
runsc --network=none --ignore-cgroups run -bundle /tmp/ctr demo &
sleep 3
# the app stub is the runsc-descendant whose maps contain the guest binary:
for p in $(pgrep -f runsc-sandbox); do echo == $p; cat /proc/$p/maps; done
ps -eo pid,ppid,comm | awk '$3=="exe"'   # stubs/sandbox all show comm "exe"
```

Then profile the spinning guest with systing and observe the sampled leaf frames render as hex
while the same busybox loop unsandboxed resolves.

josefbacik:
Contextual labels for frames that fail symbolization, and recovery of the text pages gVisor's
syscall patching breaks out of guest binaries. Folds in and closes #136 (design + analysis
below, per review preference for PR-carried design).

## Problem

Stack samples from processes running inside gVisor (runsc) sandboxes largely fail user-frame
symbolization: application frames render as bare `0x...` hex in both trace and `--continuous`
modes. Host-side processes on the same machine symbolize fine, and the same binaries symbolize
fine when run unsandboxed. On a gVisor-based Kubernetes cluster we observe the majority of
sandbox-attributed CPU samples with mostly-unresolved stacks; kernel frames are unaffected.

The picture is more nuanced than "guest memory is anonymous" — there are several distinct
failure classes with different fixes. This PR implements the first two phases (labels +
island bridging); the analysis below covers the whole taxonomy so the rest is on record.

## How a gVisor guest looks to the host-side symbolizer

Verified against gvisor @ HEAD and on a live sandbox with `runsc release-20260622.0`
(systrap platform). Under systrap, guest code executes natively in *stub* processes whose
address spaces mirror the guest's: the Sentry mmaps guest ranges at the guest virtual
addresses (`pkg/sentry/platform/systrap/subprocess.go` `MapFile`, `MAP_FIXED|MAP_SHARED`). So
sampled user addresses in stub threads ARE guest VAs, frame-pointer unwinding works, and
`bpf_get_current_pid_tgid()` correctly targets the stub's `/proc/<tgid>/maps`. Guest process
names never appear in host comms (nothing calls PR_SET_NAME host-side; stubs fork without
exec), so all sandbox CPU aggregates under the runtime's identity — comm `exe` with current
releases, `runsc` with some older/vendor builds.

What backs the stub's mappings is the crux. Actual maps of a stub running a static busybox
guest:

```
64000-69000    r-xs 3ff26000 00:01 4    /memfd:runsc-memory (deleted)   <- usertrap trampolines
400000-4c2000  r-xs 00000000 00:13 426  /root/bin/guestbox              <- guest text, file-backed
4c2000-4c3000  r-xs 3ff1f000 00:01 4    /memfd:runsc-memory (deleted)   <- 4KiB patched island
4c3000-4cd000  r-xs 000c3000 00:13 426  /root/bin/guestbox
4cd000-4ce000  r-xs 3ff20000 00:01 4    /memfd:runsc-memory (deleted)   <- island
4ce000-4d5000  r-xs 000ce000 00:13 426  /root/bin/guestbox
...
7fa398efc000-7fa398eff000 r-xp 00000000 00:00 0                         <- stub/sysmsg text (anon)
```

1. **Guest ELF text is mostly host-file-backed.** The gofer donates host FDs for regular
   files; private file mappings only leave the host file on write (`pkg/sentry/mm/pma.go`
   COW). The symbolic path (`/root/bin/guestbox`) is relative to the gofer's private mount
   namespace and useless on the host — but blazesym's default `map_files: true` opens the
   mapping through `/proc/<pid>/map_files/<range>`, which works regardless of namespaces
   (verified: returns the real ELF). So plain guest text frames already resolve today while
   the stub is alive.

2. **systrap's syscall patching punches unresolvable holes exactly where samples land.**
   `pkg/sentry/platform/systrap/usertrap` lazily rewrites executed syscall sites in guest
   text; the write COW-breaks precisely those 4KiB pages into the Sentry's `runsc-memory`
   memfd (executable vmas copy only the required pages). Result: a file/memfd mosaic where
   the memfd islands sit at *hot syscall sites* — over-represented among CPU samples of
   syscall-heavy workloads. Island addresses still correspond to functions of the original
   binary; only the backing changed. (Contrast: the ptrace platform shows a single contiguous
   file-backed text mapping — no islands.) Hazard: an island's own `map_files` entry opens
   the pool memfd, whose offset-0 bytes can coincidentally parse as an ELF header; the maps
   entries' large pool offsets then fail offset translation, so today this degrades to
   "unknown" — but a small-offset island could in principle resolve to a *wrong* symbol.

3. **Sandbox runtime execution is structurally unsymbolizable and unlabeled.** Every guest
   syscall/fault executes trampoline + sysmsg handler code on the stub thread, in memfd/anon
   mappings (the guest-side `/proc/<pid>/maps` even labels the trampoline region
   `[usertrap]`). For syscall-heavy guests a large share of "guest" CPU samples land here and
   can never resolve to an application binary — that's not a symbolization failure so much as
   *unattributed sandbox overhead*, and labeling it makes gVisor's CPU cost directly
   measurable in profiles.

4. **Exited stubs** hit the existing dead-process path (user frames rendered as hex
   unconditionally). Sandboxing platforms often run short-lived episodic work, so this class
   is over-represented there — but it's a general gap, tracked separately.

5. **JIT code**: for JIT runtimes in sandboxes (node/V8, JVM, CPython 3.14's copy-and-patch
   JIT), the engine's own text resolves per (1) — but JIT output lives in anonymous exec
   pages and renders as hex. Not gVisor-specific, but gVisor makes the standard mitigations
   harder: perf-map files are written inside the *guest* filesystem under the *guest* pid, so
   the host-side symbolizer never finds them. In practice JIT pages can be the single largest
   unresolved term on such hosts. Deserves its own follow-up (jitdump/perf-map translation).

6. **The Sentry itself may not resolve**: release runsc binaries can ship without an ELF
   symbol table, in which case the (correctly identified) `runsc` module yields no symbols —
   Go's `.gopclntab` is present but blazesym doesn't consume it. Module-level labeling below
   at least attributes these frames to the sandbox runtime.

## What this PR does

**New `sandbox_maps` module** — parses `/proc/<tgid>/maps` once per live process at
symbolization time and detects gVisor processes (`/proc/<tgid>/exe` → runsc, or
`runsc-memory`/`systrap-memory` pool memfds present).

**Island bridging** — executable `runsc-memory` islands whose file-backed neighbor mappings
agree (same file, address-adjacent, offset-congruent) are symbolized against the original ELF
through the *neighbor's* `map_files` link with `Input::FileOffset`. This recovers the pages
from class 2 — exactly the hot syscall-site leaves — and the conservative both-neighbors rule
means a pool page between unrelated mappings never bridges, which also sidesteps the
wrong-symbol hazard above.

**Labels** — when symbolization fails, render the most specific context known instead of bare
hex, reusing the existing `unknown (<module>) <addr>` miss shape (it's how kernel misses
render today, so downstream parsers already handle it):

| frame location | rendered as |
|---|---|
| stub/sysmsg/trampoline execution, Sentry-created mappings | `unknown ([gvisor:runtime]) <0x...>` |
| guest pool pages (COW'd data, unbridgeable) | `unknown ([gvisor:guest]) <0x...>` |
| anon-exec pages in a process with a JIT runtime mapped | `unknown ([jit:node]) <0x...>` etc. |
| module known, symbol lookup failed (stripped binaries) | `unknown (<module>) <0x...>` |
| process exited before end-of-trace symbolization | `unknown ([exited]) <0x...>` |

`[gvisor:runtime]` frames (plus stripped-`runsc` module frames) make the sandbox's own CPU
overhead directly measurable, which for gVisor-heavy hosts is arguably worth more than any
individual recovered frame.

The classification leans on mechanism, not position: all guest-created memory is served from
the `runsc-memory` pool; `systrap-memory` is sysmsg machinery; pure host-anonymous mappings
in a sandbox process can only come from the stub/Sentry itself (a guest cannot create host
mappings outside the pool); the usertrap trampoline table is pool-backed but injected below
the guest image. Details in the module docs.

`--no-frame-labels` reverts every label to the historical bare hex. Bridging stays on
unconditionally — it only ever adds real symbols and is a no-op outside gVisor processes.

## Future phases (not in this PR, recorded here since #136 closes)

**Guest-maps-based symbolization** for what remains (wholly-memfd text, JIT regions via guest
perf-map files, guest comm/pid attribution): gVisor already assembles everything needed —
`containerManager.ProcfsDump` returns per-process guest maps (path, offset, perms), comm, exe
and start time (`runsc/boot/procfs/dump.go`) — but no stock CLI exposes it; `runsc exec <ctr>
cat /proc/<pid>/maps` works today at the cost of running a process inside the sandbox, which
profilers shouldn't do routinely. Sketch: enumerate sandboxes from the runsc state dir, pull
guest maps per process, correlate host stub tgid ↔ guest process (stub start-time ≈ guest
exec time; fall back to matching sampled VAs against guest mm ranges — fork-siblings collide
but symbolize identically), resolve guest paths against the container rootfs (OCI bundle
`root.path`) or by build-id, and symbolize by VA identity. A tiny upstream gVisor change
adding `runsc debug --procfs` would make this clean. This is also the unlock for pystacks
inside sandboxes (discovery needs the guest maps and binaries; the BPF walker already reads
the right addresses since stub VAs are guest VAs).

Out of scope entirely: VM-based runtimes (Firecracker, Kata, QEMU-class) — guests there are
architecturally invisible to host BPF; symbolization belongs to an in-guest agent, and the
host side can at most label the VMM threads. The dead-process gap (class 4) and JIT story
(class 5) generalize beyond sandboxes and warrant their own changes (maps snapshot at
first-sample time; jitdump/perf-map support incl. guest-fs translation).

## Testing

- Unit tests for parsing, gVisor detection, bridging (offset-congruence, adjacency, and
  different-file negative cases) and every label class. The primary fixture is the maps
  excerpt above, captured from a live runsc `release-20260622.0` systrap stub.
- An end-to-end test drives the real `finish()`-time symbolization over live `/proc`: a stack
  from the test's own process must resolve (or at minimum carry a module label), a stack from
  an impossible tgid takes the dead-process path and must render `unknown ([exited]) <...>`,
  and the `--no-frame-labels` path must reproduce the historical bare-hex output.
- `cargo test` (372), `cargo fmt --check`, `cargo clippy --no-default-features -- -D warnings`
  all clean.
- Not yet validated against a live gVisor sandbox end-to-end (my dev environment can't run
  BPF). Manual validation path on any Linux host with runsc installed:

```
mkdir -p /tmp/ctr/rootfs/bin && cp $(which busybox) /tmp/ctr/rootfs/bin/busybox
ln -s busybox /tmp/ctr/rootfs/bin/sh
cd /tmp/ctr && runsc spec -- /bin/sh -c 'i=0; while true; do i=$((i+1)); done'
sed -i 's/"terminal": true/"terminal": false/' config.json
runsc --network=none --ignore-cgroups run -bundle /tmp/ctr demo &
sleep 3
# the app stub is the runsc-descendant whose maps contain the guest binary:
for p in $(pgrep -f runsc-sandbox); do echo == $p; cat /proc/$p/maps; done
```

Then profile the spinning guest with systing: before this change the sampled leaves render as
hex; after it, leaves in patched pages symbolize to busybox functions and stub-region samples
render as `unknown ([gvisor:runtime]) <...>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,7 +3430,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.3"
+version = "1.10.4"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "systing"
-version = "1.10.3"
+version = "1.10.4"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,6 +50,7 @@ pub mod perfetto;
 pub mod pystacks;
 pub mod record;
 pub mod ringbuf;
+pub mod sandbox_maps;
 pub mod sched;
 pub mod session_recorder;
 pub mod stack_recorder;
--- a/src/main.rs
+++ b/src/main.rs
@@ -65,6 +65,10 @@ struct Command {
     /// Enable debuginfod for enhanced symbol resolution (requires DEBUGINFOD_URLS environment variable)
     #[arg(long)]
     enable_debuginfod: bool,
+    /// Render frames that fail symbolization as bare hex instead of contextual
+    /// labels like "unknown ([gvisor:runtime]) <0x...>" or "unknown ([jit:node]) <0x...>"
+    #[arg(long)]
+    no_frame_labels: bool,
     /// Disable scheduler event tracing (sched_* tracepoints and scheduler event recorder)
     #[arg(long)]
     no_sched: bool,
@@ -182,6 +186,7 @@ impl From<Command> for Config {
             pystacks_pids: Vec::new(), // CLI doesn't expose this yet, uses discovery
             pystacks_debug: cmd.pystacks_debug,
             enable_debuginfod: cmd.enable_debuginfod,
+            no_frame_labels: cmd.no_frame_labels,
             no_sched: cmd.no_sched,
             syscalls: cmd.syscalls,
             markers: cmd.markers,
--- a/src/sandbox_maps.rs
+++ b/src/sandbox_maps.rs
@@ -1,0 +1,501 @@
+//! Host-side analysis of a process's memory map for symbolizing sandboxed
+//! (gVisor/runsc) workloads and labeling otherwise-unresolvable frames.
+//!
+//! Under gVisor's systrap platform, guest code executes in *stub* processes
+//! whose address spaces mirror the guest's. Guest ELF text is mostly backed by
+//! gofer-donated host file descriptors (symbolizable through
+//! `/proc/<pid>/map_files/`), but three things still defeat a
+//! `/proc/<pid>/maps`-driven symbolizer:
+//!
+//! 1. systrap's syscall patching COW-breaks the 4KiB text pages containing
+//!    executed syscall sites into the Sentry's `runsc-memory` memfd, leaving
+//!    memfd "islands" inside file-backed text. Island addresses still
+//!    correspond to functions of the original file, so they can be recovered
+//!    by re-deriving the file offset from the file-backed neighbors
+//!    ("bridging", see [`ProcessMaps::bridge_for`]).
+//! 2. The sysmsg/trampoline code every guest syscall executes lives in
+//!    memfd/anonymous mappings outside the guest image range — those frames
+//!    are gVisor runtime overhead, not application code.
+//! 3. Anonymous guest ranges (JIT output, copied-up pages) are backed by the
+//!    memory-pool memfd whose offsets have no relation to any file.
+//!
+//! This module parses `/proc/<pid>/maps` once per process at symbolization
+//! time and answers, per address: "can this be bridged to a file?" and if
+//! not, "what is the most specific label we can attach instead of raw hex?".
+//! The labels reuse systing's existing miss format (`unknown (<module>)
+//! <0xADDR>`), so downstream consumers see a module-shaped string:
+//! `[gvisor:runtime]`, `[jit:<runtime>]`, or the real module basename when
+//! only the symbol lookup failed.
+
+use std::fs;
+use std::path::PathBuf;
+
+/// What backs one `/proc/<pid>/maps` entry, as far as classification cares.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Backing {
+    /// Regular file with an absolute path (may be namespace-relative and
+    /// unopenable by path; `map_files` still works).
+    File(PathBuf),
+    /// A memfd (`/memfd:<name> (deleted)`); gVisor's memory pools land here.
+    Memfd(String),
+    /// No backing object at all (pure anonymous).
+    Anon,
+    /// Bracketed pseudo-entries (`[vdso]`, `[stack]`, ...) and
+    /// `anon_inode:...` style components.
+    Special(String),
+}
+
+#[derive(Debug, Clone)]
+struct MapEntry {
+    start: u64,
+    end: u64,
+    exec: bool,
+    /// File offset field of the maps line (memfd offsets are pool offsets
+    /// and meaningless for ELF math; file offsets are real).
+    offset: u64,
+    backing: Backing,
+}
+
+/// Result of a bridge lookup: symbolize `file_offset` against the ELF that
+/// is reachable through `map_files_path` (a namespace-immune magic link of a
+/// *file-backed* neighbor entry, not of the island itself).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BridgedAddr {
+    pub map_files_path: PathBuf,
+    pub module_name: String,
+    pub file_offset: u64,
+}
+
+/// Per-process map analysis. Build once per tgid with [`ProcessMaps::load`]
+/// (or [`ProcessMaps::parse`] for tests), query per unresolved address.
+#[derive(Debug, Default)]
+pub struct ProcessMaps {
+    tgid: i32,
+    entries: Vec<MapEntry>,
+    /// Any gVisor memory-pool memfd present (`runsc-memory`,
+    /// `systrap-memory`).
+    gvisor_memfd: bool,
+    /// `/proc/<tgid>/exe` resolves to a runsc binary.
+    gvisor_exe: bool,
+    /// Basename of a known JIT runtime found among file-backed entries.
+    jit_runtime: Option<&'static str>,
+    /// Address span of file-backed guest mappings ([lo, hi)); executable
+    /// pool/anon mappings outside this span are runtime scaffolding, inside
+    /// are guest anon (JIT or data made executable).
+    file_span: Option<(u64, u64)>,
+}
+
+/// memfd names used by gVisor for application/stub memory. Entries here mark
+/// a process as sandbox-related and their mappings as pool-backed.
+const GVISOR_MEMFDS: &[&str] = &["runsc-memory", "systrap-memory"];
+
+/// Module basenames whose presence marks a process as running a JIT runtime;
+/// anonymous executable guest pages in such a process are labeled
+/// `[jit:<tag>]`. Deliberately short and conservative.
+const JIT_RUNTIMES: &[(&str, &str)] = &[
+    ("node", "node"),
+    ("libnode", "node"),
+    ("libjvm", "jvm"),
+    ("libpython3", "python"),
+    ("libv8", "v8"),
+];
+
+impl ProcessMaps {
+    /// Read and analyze `/proc/<tgid>/maps`. Returns `None` when the process
+    /// is gone or unreadable — callers fall back to plain hex rendering.
+    pub fn load(tgid: i32) -> Option<Self> {
+        let maps = fs::read_to_string(format!("/proc/{tgid}/maps")).ok()?;
+        let exe = fs::read_link(format!("/proc/{tgid}/exe"))
+            .ok()
+            .and_then(|p| {
+                p.file_name()
+                    .and_then(|f| f.to_str())
+                    .map(|s| s.to_string())
+            })
+            .unwrap_or_default();
+        Some(Self::parse(tgid, &maps, &exe))
+    }
+
+    /// Analyze pre-read maps content. `exe_basename` is the resolved
+    /// basename of `/proc/<tgid>/exe` ("" when unknown).
+    pub fn parse(tgid: i32, maps: &str, exe_basename: &str) -> Self {
+        let mut pm = ProcessMaps {
+            tgid,
+            // The sandbox re-execs /proc/self/exe so its comm is "exe", but
+            // the exe link still resolves to the runsc binary; stubs inherit
+            // it across fork.
+            gvisor_exe: exe_basename.starts_with("runsc"),
+            ..Default::default()
+        };
+
+        for line in maps.lines() {
+            let Some(entry) = parse_maps_line(line) else {
+                continue;
+            };
+            match &entry.backing {
+                Backing::Memfd(name) if GVISOR_MEMFDS.iter().any(|m| name == m) => {
+                    pm.gvisor_memfd = true;
+                }
+                Backing::File(path) => {
+                    if let Some(base) = path.file_name().and_then(|f| f.to_str()) {
+                        for (prefix, tag) in JIT_RUNTIMES {
+                            if base.starts_with(prefix) {
+                                pm.jit_runtime = Some(tag);
+                            }
+                        }
+                    }
+                    let (lo, hi) = pm.file_span.unwrap_or((u64::MAX, 0));
+                    pm.file_span = Some((lo.min(entry.start), hi.max(entry.end)));
+                }
+                _ => {}
+            }
+            pm.entries.push(entry);
+        }
+        pm
+    }
+
+    /// Whether this process belongs to a gVisor sandbox (Sentry, gofer, or a
+    /// stub mirroring a guest address space).
+    pub fn is_gvisor(&self) -> bool {
+        self.gvisor_exe || self.gvisor_memfd
+    }
+
+    fn entry_for(&self, addr: u64) -> Option<&MapEntry> {
+        // Entries are in address order as read from /proc.
+        self.entries
+            .iter()
+            .find(|e| e.start <= addr && addr < e.end)
+    }
+
+    /// If `addr` falls in a pool-memfd island whose file-backed neighbors
+    /// agree on the original file, return where to symbolize it instead.
+    ///
+    /// The island's own bytes are a patched copy, but its *addresses* still
+    /// belong to the original binary, so `addr`'s file offset is derived
+    /// from the left neighbor: `addr - L.start + L.offset`. Conservative
+    /// rule: both neighbors must reference the same file with congruent
+    /// offsets (`R.offset - L.offset == R.start - L.start`), so a pool page
+    /// that merely happens to sit between two unrelated files never bridges.
+    pub fn bridge_for(&self, addr: u64) -> Option<BridgedAddr> {
+        let idx = self
+            .entries
+            .iter()
+            .position(|e| e.start <= addr && addr < e.end)?;
+        let island = &self.entries[idx];
+        if !matches!(island.backing, Backing::Memfd(_)) || !island.exec {
+            return None;
+        }
+        let left = self.entries.get(idx.checked_sub(1)?)?;
+        let right = self.entries.get(idx + 1)?;
+        let (Backing::File(lpath), Backing::File(rpath)) = (&left.backing, &right.backing) else {
+            return None;
+        };
+        if lpath != rpath || !left.exec {
+            return None;
+        }
+        // Neighbors must be address-adjacent to the island and offset-congruent.
+        if left.end != island.start || island.end != right.start {
+            return None;
+        }
+        if right.offset.wrapping_sub(left.offset) != right.start.wrapping_sub(left.start) {
+            return None;
+        }
+        let module_name = lpath
+            .file_name()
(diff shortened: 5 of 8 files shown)


josefbacik:
Added the integration test: `tests/gvisor_record.rs`, runs via `./scripts/run-integration-tests.sh gvisor_record`.

What it does: builds a self-contained OCI bundle (tmpfs rootfs holding just `/bin/sh` and its linker closure — independent of how the host filesystem is mounted), `runsc run`s a workload mixing a pure-arithmetic spinner (guest-text samples) with several `cd /` syscall loops (systrap trampoline/handler samples), records 12s with the full in-process pipeline, then validates from the trace output that:

- stack samples attributed to the runsc process family exist;
- `unknown ([gvisor:*])` classification engaged (runtime-region samples);
- guest text symbolized: resolved frames from non-runsc modules are present.

Environment gating: needs root/BPF and a runsc binary (`SYSTING_TEST_RUNSC` or `$PATH`), and skips cleanly with a reason otherwise, so unprivileged CI and contributors without runsc are unaffected. `--ignore-cgroups` because test kernels commonly lack the `pids` controller runsc otherwise insists on; readiness is polled against the suite's slow-machine budget rather than fixed sleeps.

Validated with runsc `release-20260622.0` under virtme-ng/QEMU-TCG (no KVM) on a BPF-enabled 6.12 kernel — a deliberately slow environment: 133K sandbox-attributed samples across 5253 unique stacks; **10830 resolved guest-module frames** (e.g. `chdir (libc.so.6) <0x7febb6...>` — guest libc resolving through `map_files`); **8 `unknown ([gvisor:runtime])` frames** (trampoline-region samples are rare when emulation starves the guest — the multi-process syscall workload exists precisely to keep that count reliably nonzero; real hardware sees far more). Skip paths verified separately (no runsc; runsc without privileges).

josefbacik:
## rust-bpf-code-reviewer — review of the full PR diff (a3c2613 + d8900b5, per repo policy that the in-repo review agent passes every change pre-merge)

*Mechanics note: executed per `.claude/agents/rust-bpf-code-reviewer.md` — five-pass process, all criteria applied to the full diff vs `b575bcb`.*

### Summary
Clean, well-scoped change: a self-contained maps-analysis module with the symbolization hook kept minimal, zero new `unsafe`, and an integration test that asserts the change's own value proposition against a live sandbox. The mechanism-grounded classification rules (pool-memfd names over address heuristics) are the right call and are documented as constraints, not narration.

### Critical Issues
None found. Specifically checked: no schema changes (frame-string *content* changes only — `SCHEMA_CHANGES.md` correctly untouched, `SCHEMA_VERSION` correctly untouched); no new `unsafe` blocks; no raw `tid`/`pid` columns (the test's attribution goes through the prescribed `stack_sample.utid → thread.utid → tid` join); no source-line references in comments; no obvious-comments.

### Improvements
1. **Versioning (fixed in follow-up commit):** non-schema changes require a patch bump per the schema & versioning criteria — the branch shipped at 1.10.3. Bumped to **1.10.4**. (Noting for the record: #134, a test-only PR, merged without a bump — if test-only changes are meant to be exempt, the criteria could say so; this review applied the rule as written.)

### Suggestions
1. `parse_maps_line`'s backing classification uses an `if let` chain (`strip_prefix`/`starts_with` dispatch). Considered against the match-over-if/else rule: a `match` needs guards that can't bind the `strip_prefix` result, so the chain is the cleaner form here. No change requested.
2. `symbolize_user_addr` constructs a `Source::Elf` per bridged address. blazesym caches the parsed resolver by path and this is the cold (failed-symbolization) path with per-address memoization above it, so the allocation is immaterial — flagged only so the reasoning is on record.

### Positive Notes
- `Backing::Special` deliberately falls through to no-label, so `[vdso]`/`[stack]` entries in a sandbox process can never be misclassified as `[gvisor:runtime]` — a subtle correctness point handled correctly.
- The bridge rule requiring *both* neighbors to agree (same file, adjacent, offset-congruent) makes false-positive bridging structurally impossible rather than merely unlikely, and sidesteps the pool-memfd fake-ELF hazard documented in the PR.
- The test's environment gating (skip-with-reason carrying runsc's own stderr) and the multi-process workload rationale (keeping trampoline samples reliably nonzero on starved machines) are the kind of operational honesty that keeps integration suites trustworthy.

PR #138, opened 2026-07-03
Query gVisor control sockets for guest-side maps during symbolization

josefbacik:
Second phase of the gVisor symbolization work from #137: consume the *guest's own* view of
its address space, so classification no longer depends on host-side inference where the
Sentry has the authoritative answer.

## Mechanism

runsc's per-sandbox control socket (`<state-root>/runsc-<sandbox-id>.sock`) serves
`containerManager.ProcfsDump`: per guest process, the full guest maps (path, offset,
permissions), comm/exe/args, and task creation time. The wire protocol is one JSON
request/response over a unix stream socket. Crucially this is Sentry-side only — nothing
executes inside the guest, unlike `runsc exec`-based approaches, which makes it appropriate
for a profiler to do routinely.

The new `gvisor_guest` module:

- **Discovery**: probes the containerd-shim and standalone state roots
  (`SYSTING_RUNSC_ROOTS` overrides) for control sockets, one `ProcfsDump` per sandbox, once
  per symbolization pass and only after a gVisor process is actually encountered.
  Defenses for production nodes: per-call timeouts, a response size cap, and a total
  enumeration budget so a host full of wedged sandboxes cannot stall the pass — any
  failure just degrades that sandbox to host-only symbolization.
- **Correlation**: a stub is matched to the guest process it mirrors by exec-file range
  containment. Stub VAs are guest VAs under systrap, and the stub's file view is a fragment
  set of the guest's (syscall-patch islands split it), so every stub exec-file range must be
  contained in some guest range. Fork-siblings tie with identical layouts and symbolize
  identically, so ties are benign.
- **Lookup**: for a correlated process, one address query answers file (guest path + file
  offset + module), `[usertrap]` (runtime-injected), guest-anonymous, or unmapped.

The symbolization chain gains one step, between island bridging and the host-side label
fallback. What changes in output:

| case | before | after |
|---|---|---|
| pool-backed guest file mapping (copied-up, donation-less configs) | `unknown ([gvisor:guest])` | `unknown (<real module>)` |
| usertrap trampoline region | `[gvisor:runtime]` via below-image heuristic | `[gvisor:runtime]` via the Sentry's own `[usertrap]` label |
| guest anon (JIT output, heap) | inferred `[gvisor:guest]` | confirmed `[gvisor:guest]` |

`--no-gvisor-guest-maps` disables the lookups entirely. With them disabled (or the socket
unreachable, or nothing correlated) behavior is exactly #137's.

## Deliberately not in this PR

- **Guest comm/pid attribution enrichment** (naming stub processes after their guest
  workload): high value, but process records are written before the stack pass runs, so it
  needs its own ordering design — next PR.
- **Symbolizing pool-backed file content via the container rootfs**: `GuestAddr::File`
  carries the guest path and file offset as the seam, but resolving guest paths to host
  files (bundle `root.path`, multi-container ambiguity, staleness of copied-up binaries)
  is left until a deployment demonstrates the need — on the cluster measured in #137's
  analysis, gofer FD donation makes file-backed text resolve host-side already.
- Guest perf-map translation (guest `/tmp` is usually sandbox-internal tmpfs — needs the
  rootfs work above where it's feasible at all) and pystacks-in-sandbox.

## Validation

- Wire format and response shape verified against a live sandbox first (runsc
  `release-20260622.0`, systrap): the module's test fixture is the captured `ProcfsDump`
  response, values verbatim.
- Unit tests: wire framing (brace-balance completion incl. string/escape cases), fixture
  parsing, address classification (file offset math, `[usertrap]`, anon, unmapped), and
  correlation (fragment containment, alien-range rejection, partial-agreement rejection).
- The gVisor integration test now also drives the new path against the sandbox it launches:
  control socket reached in the test's private state root, dump parsed, and a live stub
  correlated back to the guest shell — asserted as `comm == "sh"`. Green under
  virtme-ng/QEMU-TCG on a BPF 6.12 kernel alongside all of #137's assertions
  (38K resolved-user frames of which 12K guest-module, labels present).
- `cargo test` (377), fmt, `clippy --no-default-features -- -D warnings` clean.

Reviewed pre-commit per repo policy (rust-bpf-code-reviewer): the review caught a missing
total-enumeration budget (a host of wedged sandboxes would have stalled symbolization —
fixed) and a match-over-if/else cleanup in the guest lookup (applied); non-schema patch
bump to 1.10.5 included.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,7 +3430,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.4"
+version = "1.10.5"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "systing"
-version = "1.10.4"
+version = "1.10.5"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/gvisor_guest.rs
+++ b/src/gvisor_guest.rs
@@ -1,0 +1,461 @@
+//! Guest-side process truth for gVisor sandboxes.
+//!
+//! The host view of a sandboxed process (a systrap *stub*) is enough to
+//! symbolize file-backed guest text, but three things only the Sentry
+//! knows: which guest process a stub mirrors (host comm is always the
+//! runtime's own), the guest-path identity of mappings whose host backing
+//! is the memory pool (copied-up files, donation-less configurations), and
+//! the authoritative extent of runtime-injected regions (`[usertrap]`).
+//!
+//! runsc's control socket already serves all of it: the
+//! `containerManager.ProcfsDump` RPC returns, per guest process, the full
+//! guest maps (path, offset, perms), comm/exe/args, and the task creation
+//! time. The socket lives in the runsc state directory
+//! (`<root>/runsc-<sandbox-id>.sock`), the protocol is a single JSON
+//! request/response over a unix stream socket, and the call is
+//! Sentry-side only — nothing executes inside the guest.
+//!
+//! [`SandboxIndex::load`] snapshots every reachable sandbox once per
+//! symbolization pass; [`SandboxIndex::correlate`] matches a host stub's
+//! address space to the guest process it mirrors (stub VAs are guest VAs
+//! under systrap, so file-backed ranges must agree exactly); the returned
+//! [`GuestProcess`] then answers per-address queries.
+
+use std::io::{Read, Write};
+use std::os::unix::net::UnixStream;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+use serde::Deserialize;
+
+/// State directories probed for control sockets: the containerd shim's
+/// (Kubernetes) and the standalone default. `SYSTING_RUNSC_ROOTS`
+/// (colon-separated) replaces the list when set.
+const DEFAULT_RUNSC_ROOTS: &[&str] = &["/run/containerd/runsc/k8s.io", "/var/run/runsc"];
+
+/// Per-call I/O timeout. The RPC is served from Sentry memory and is
+/// fast; a wedged sandbox must not stall symbolization.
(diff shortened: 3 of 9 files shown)


josefbacik:
## rust-bpf-code-reviewer — pre-commit review of this PR's diff (9a0db77 vs main)

*Mechanics: executed per `.claude/agents/rust-bpf-code-reviewer.md`, five passes over the full diff, findings addressed before the commit was pushed (this is the post-fix record).*

### Summary
Well-bounded second phase: one new module with a narrow public surface, one added step in the symbolization chain, and defensive I/O throughout. The correlation logic is grounded in a verifiable invariant (stub exec-file ranges are fragments of guest ranges under systrap) rather than heuristics, and the test fixture is a captured wire response from a live sandbox, values verbatim.

### Critical Issues
None outstanding. One was found during review and fixed pre-commit: `SandboxIndex::load` had per-call timeouts but no total budget, so a host with many wedged sandboxes could have stalled the symbolization pass for minutes — a 10s enumeration budget now bounds the whole snapshot, degrading to whatever was gathered.

### Improvements
1. (Applied during review) Guest address classification rewritten from an `if/else-if` chain to a `match` on the pathname with a file guard — the three-way dispatch reads as patterns.
2. Versioning: non-schema change → patch bump to 1.10.5, included in the commit. No schema surface touched (no new tables/columns; `SCHEMA_CHANGES.md`/`SCHEMA_VERSION` correctly untouched).

### Suggestions
1. `GuestProcess::lookup` allocates the module and path `String`s per query; it sits behind the per-address memoization cache on the failed-symbolization path, so this is cold by construction. No change requested.
2. The urpc client deliberately implements only what the one RPC needs (no FD passing, no method dispatch). If a second control-socket call is ever wanted (e.g. `Processes`), promote the request/response plumbing to take the method name — the seam already exists.

### Positive Notes
- Zero `unsafe` in the diff; all I/O failure modes (unreachable socket, oversized response, decode failure, `success:false`) degrade to host-only symbolization with a note rather than erroring the trace.
- The correlation containment test includes the negative cases that matter: alien ranges, partial agreement, and the empty stub view all refuse to correlate — misattribution is structurally rejected rather than merely unlikely.
- The integration test asserting a live stub correlates to the guest shell closes the loop between the captured fixture and reality; the fixture cannot silently drift from the wire format without the gate noticing.

PR #139, opened 2026-07-09
Shard the sched/IRQ consumer per ring and split IRQ into a standalone recorder

josefbacik:
## Problem

On large many-core hosts, systing drops sched/IRQ events at a sustained rate
in both `--duration` and `--continuous` mode. Reported and measured on a
144-CPU GB200 host (~29k threads, real GPU workload): ~7,200 missed
events/sec at defaults, and — the tell — pid-scoped loss is *ring-size
invariant* (~1,900/s at default rings and at `--ringbuf-size-mib 200`).

The bottleneck is not the rings: the BPF side is already sharded
(`NR_RINGBUFS = 8`, events land in ring `cpu % 8`) and there is one poller
thread per ring — but all 8 pollers funnel into ONE bounded channel drained
by ONE consumer thread holding one `SchedEventRecorder`. Once that consumer
saturates, backpressure fills the kernel rings and BPF drops (honestly
counted, but the data is gone). IRQ events amplify the problem under `-p`:
they record the *interrupted* task, and on a busy host nearly every IRQ
interrupts some traced task, so unfilterable IRQ volume competes with the
sched events the trace actually needs.

## Changes

Provenance: the two patches were authored by the reporting user's assistant
(`Claude Coworker <coworker@anthropic.com>`, authorship preserved in the
commits) against the report above, with both fix shapes agreed with the
maintainer; this PR carries the series **unmodified**, adds an independent
line-by-line review (comment below), and folds in the required version bump:

1. **recorders: shard the sched/IRQ consumer per ring with SPSC channels** —
   each ring gets its own bounded channel and consumer thread feeding its own
   `SchedEventRecorder` shard. Correct by construction: `cpu → ring` is a
   static modulo and every piece of recorder state is keyed by CPU, so shards
   share nothing. Shards stream into the shared parquet writer through the
   existing `SharedCollector`, with all per-event record types now locally
   batched (~10k) so the writer lock is not a per-event serialization point.
   Bounded-backpressure semantics preserved. No BPF or schema changes.

2. **recorders: split IRQ/softirq tracing into a standalone "irq" recorder** —
   the four irq/softirq programs move out of the sched recorder into their own
   registry entry (default ON), with `--no-irq` to shed them; when disabled the
   programs are not loaded at all. Schema unchanged (IRQ events remain
   task_events on the shared rings). **Behavior change:** `--no-sched` no
   longer disables IRQ tracing, and `--only-recorder sched` no longer records
   IRQ — list `irq` explicitly where it's wanted.

3. Version bump to 1.10.6 (non-schema change; no `SCHEMA_VERSION` /
   `SCHEMA_CHANGES.md` impact — none needed).

## Verification

- Applies to current `main` and builds clean: full unit/doc test suite
  416 passed / 0 failed, `clippy -D warnings` clean, `fmt --check` clean.
  (Also verified on the series' original base: 404/404, clean.)
- Hermetic VM integration gate (root + BPF, x86_64 v6.12, pure TCG), all six
  integration-test targets: **running at PR-open time** — results will be
  posted on this PR before any real-host validation run.
- Targeted live checks for the new surfaces (per-ring consumer threads
  spawning, `--no-irq` shedding, `--only-recorder irq` ring wiring,
  `--continuous` + SIGINT drain): same VM, queued behind the suite; results
  posted here.
- Independent line-by-line review posted as a comment below (structured
  verdict per the repo's reviewer convention), run against an independently
  derived design analysis of the same bottleneck.

**Not yet validated:** event-loss *rates* on real many-CPU hosts — the VM
gate proves the BPF programs load and the logic runs, not arch or throughput
behavior. The reporting host (real workload, standing by) validates loss
elimination before merge; results will be posted on this PR.

## Notes

- Memory: per-ring channel preallocation is ~83 MB total (vs ~10 MB for the
  single shared channel), paid at startup, independent of load — called out
  in the first commit message.
- Follow-ups identified during review, deliberately not in this PR:
  suppressing the kernel-side `SCHED_WAKEUP` emission that userspace discards
  as redundant with `SCHED_WAKING` (free ring-bandwidth win on wakeup-heavy
  hosts); writing the missed-event counters into the trace itself instead of
  only the exit summary; scaling `NR_RINGBUFS` with CPU count if 8-way
  consumer parallelism still ceilings on very large hosts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,7 +3430,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.5"
+version = "1.10.6"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "systing"
-version = "1.10.5"
+version = "1.10.6"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/docs/USAGE.adoc
+++ b/docs/USAGE.adoc
@@ -50,6 +50,7 @@ Use *--list-recorders* to see available recorders and their default states.
 Available recorders:
 
 * `sched` - Scheduler event tracing (enabled by default)
+* `irq` - IRQ and softirq event tracing (enabled by default)
 * `syscalls` - Syscall tracing (disabled by default)
 * `sleep-stacks` - Sleep stack traces (enabled by default)
 * `cpu-stacks` - CPU perf stack traces (enabled by default)
@@ -131,6 +132,7 @@ Example:
 # systing --list-recorders
 Available recorders:
   sched          - Scheduler event tracing (on by default)
+  irq            - IRQ and softirq event tracing (on by default)
   syscalls       - Syscall tracing
   sleep-stacks   - Sleep stack traces (on by default)
   cpu-stacks     - CPU perf stack traces (on by default)
@@ -141,6 +143,15 @@ Available recorders:
 === *--no-cpu-stack-traces*
 
 Do not record the `perf` style stack traces for tasks while they're running.
+
+=== *--no-irq*
+
+Disable IRQ and softirq event tracing (the `irq_handler_entry`/`irq_handler_exit` and `softirq_entry`/`softirq_exit` tracepoints).
+IRQ events cannot meaningfully be narrowed by *-p* (nearly every IRQ interrupts _some_ traced task on a busy host), so on large machines they can dominate a pid-scoped trace; this flag sheds that load independently of the scheduler recorder.
+Equivalent to leaving `irq` out of an *--only-recorder* list.
+
+Note: before the `irq` recorder was split out, these tracepoints were part of the `sched` recorder and *--no-sched* disabled them too.
+*--no-sched* no longer affects IRQ tracing; use *--no-irq* (or recorder selection) for that.
 
 === *--no-sched*
 
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -76,6 +76,11 @@ const volatile struct {
 	u32 no_sched;          /* Sched recorder off: sched_switch still runs for
 				* cpu_running_pid and sleep stacks, but task_event
 				* emission is suppressed. */
+	u32 no_irq;            /* IRQ recorder off: the irq/softirq programs are
+				* not loaded at all in that case, so this rodata
+				* gate is belt-and-braces (and lets the verifier
+				* prune the bodies if the programs ever get loaded
+				* for another reason). */
 } tool_config = {};
 
 enum event_type {
@@ -1617,8 +1622,12 @@ static void emit_stack_event(void *ctx, struct task_struct *task,
 
 static int trace_irq_event(struct irqaction *action, int irq, int ret, bool enter)
 {
-	struct task_struct *tsk = (struct task_struct *)bpf_get_current_task_btf();
+	struct task_struct *tsk;
 
+	if (tool_config.no_irq)
+		return 0;
+
+	tsk = (struct task_struct *)bpf_get_current_task_btf();
 	if (!trace_task(tsk))
 		return 0;
 
@@ -1649,8 +1658,12 @@ static int trace_irq_event(struct irqaction *action, int irq, int ret, bool ente
 
 static int trace_softirq_event(unsigned int vec_nr, bool enter)
 {
-	struct task_struct *tsk = (struct task_struct *)bpf_get_current_task_btf();
+	struct task_struct *tsk;
 
+	if (tool_config.no_irq)
+		return 0;
+
+	tsk = (struct task_struct *)bpf_get_current_task_btf();
 	if (!trace_task(tsk))
 		return 0;
 
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -1702,21 +1702,6 @@ static int handle_wakeup(struct task_struct *waker, struct task_struct *wakee,
 	return 0;
 }
 
-static int handle_sched_wakeup(struct task_struct *task, int success)
-{
-	struct task_struct *cur = (struct task_struct *)bpf_get_current_task_btf();
-
-	if (!trace_task(cur) && !trace_task(task))
-		return 0;
-	return handle_wakeup(cur, task, SCHED_WAKEUP);
-}
-
-SEC("tp_btf/sched_wakeup")
-int BPF_PROG(systing_sched_wakeup, struct task_struct *task, int success)
-{
-	return handle_sched_wakeup(task, success);
-}
-
 static int handle_sched_wakeup_new(struct task_struct *task)
 {
 	struct task_struct *cur = (struct task_struct *)bpf_get_current_task_btf();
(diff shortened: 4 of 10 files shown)


josefbacik:
## rust-bpf-code-reviewer — independent review of this PR's series (vs main 8220668)

over the full series diff. This series arrived as external patches; this review
is independent of the authors' own review process. The one Critical finding is
fixed by a commit in this PR rather than a respin (version bump — the series
itself is unmodified, authorship preserved).*

### Summary
Architecturally sound and closely matches an independently derived design for
the same bottleneck: per-ring consumers are correct by construction because
`cpu → ring` is a static modulo and all `SchedEventRecorder` state
(`cpu_states`, `pending_irqs`, `pending_softirqs`) is CPU-keyed, so shards
share nothing; the shared pieces (`UtidGenerator`, `SharedCollector`) were
already concurrency-safe. The IRQ split follows the existing recorder-registry
pattern exactly. Batching every per-event record type before touching the
shared writer removes the would-be new serialization point.

### Critical Issues
One, fixed in this PR: the series carried no `Cargo.toml` version bump
(non-schema change → patch bump required). Added as a separate commit
(1.10.6). Schema surfaces correctly untouched — no `SCHEMA_VERSION` /
`SCHEMA_CHANGES.md` entry needed.

### Improvements
Follow-up material, deliberately not blocking this series:
1. `systing_sched_wakeup` is attached unconditionally while userspace discards
   every `SCHED_WAKEUP` as redundant with `SCHED_WAKING` — each wakeup pays
   double ring+channel transit for nothing. Suppressing the emission is a free
   bandwidth win on wakeup-heavy hosts (verify first that the program has no
   side-effect map updates other recorders rely on; it appears to be pure
   emission).
2. Missed-event counts exist only in the exit summary; the trace itself never
   flags a gap. Writing them into the trace (final stats, ideally periodic
   samples) makes loss visible where the data is consumed.
3. `NR_RINGBUFS = 8` now bounds consumer parallelism as well as ring count
   (18 CPUs/ring on a 144-CPU host). If the reporting host still shows
   residual loss with 8 consumers, scaling ring count with CPUs is the next
   lever.

### Suggestions
1. `flush_pending` repeats the drain-warn loop six times; a macro could
   collapse it, but the explicit form is arguably clearer. No change requested.
2. Per-ring `CHANNEL_CAPACITY` preallocation is now ~83 MB total regardless of
   host size. Fine on the target class; worth revisiting if systing is run on
   memory-tight small hosts.

### Positive Notes
- The sharding-correctness argument is stated in the commit message in terms
  of the actual invariants (static CPU→ring mapping, CPU-keyed state) rather
  than "tests pass" — and it checks out against the source line by line,
  including the two subtle cases: `SCHED_WAKING` lands on the waker's ring but
  its handling is stateless, and IRQ ENTER/EXIT pairs cannot split across
  shards because handlers don't migrate mid-flight.
- `SharedCollector` handle lifetimes are handled correctly: extra clones are
  dropped before finalization so `Arc::try_unwrap` actually finishes the
  writer — an easy leak to write and a hard one to notice.
- `maybe_record_task` duplicate sends from per-shard discovery dedup were
  checked rather than assumed idempotent (they are — keyed check-then-insert).
- The finish-path drain test covers every locally buffered record type, which
  is exactly the regression surface the batching change created.
- Behavior changes (`--no-sched`/`--only-recorder sched` no longer implying
  IRQ) are documented in the commit message, USAGE.adoc, and the cover letter
  rather than left for users to discover.

josefbacik:
Addressed the CI failure and each review item; three commits pushed:

- **Clippy failure** (`fc0096d`): both errors came from a newer clippy on the
  CI runner promoting new lints in code that predates this series
  (`iter_over_map_values` in `perf.rs`, redundant reference in a `println!`
  in `stack_recorder.rs`). Fixed both; `clippy -D warnings` clean locally and
  should be green here now.
- **Review Improvement 1 — discarded SCHED_WAKEUP emission** (`924514a`):
  done in this PR. The `sched_wakeup` program is no longer built or attached,
  so wakeups now emit one event (SCHED_WAKING) instead of two. Verified the
  removed program was pure emission (no side-effect map writes; `trace_task`
  only reads the filter maps). The enum variant and the recorder's ignore arm
  stay, so nothing changes on the wire or for stray events.
  On wakeup-heavy hosts this reduces sched-ring volume before the sharded
  consumers even start — it compounds with patch 1.
- **Review Improvement 2 — missed-event counts in the trace**: splitting
  into a follow-up PR rather than growing this one mid-review. Shape it will
  take: a small poller (mirroring the existing sysinfo/cpufreq pattern)
  sampling the per-class `missed_events` map every second into counter
  tracks, so loss is visible and time-localized in the trace instead of only
  in the exit summary. Existing tables only — no schema change.
- **Review Improvement 3 — `NR_RINGBUFS` scaling**: deferring until the
  reporting host's validation run produces data. If 8 consumers already
  eliminate the loss there, scaling ring count buys nothing and costs mlocked
  memory on every host; if residual loss remains, that run tells us the
  factor to target. Wired to revisit either way.

VM validation status: the six-target integration gate is being re-run against
this exact head (it had been running against the pre-push head when these
commits landed); results will be posted here when it completes, before any
real-host run.

josefbacik:
VM validation complete on this exact head (`924514a`) — all green, plus one check-harness bug found and cleared along the way.

**Six-target integration gate** (virtme-ng guest, 6.12 BPF-enabled kernel, TCG/no-KVM, 4 vCPU / 3 GB):

| target | result | time |
|---|---|---|
| trace_validation | 31 passed / 0 failed | 2194s |
| memory_recorder | 5 passed / 0 failed | 599s |
| analyze_commands | 1 passed / 0 failed | 889s |
| stream_unix | 1 passed / 0 failed | 116s |
| network_throughput | 1 passed / 0 failed | 58s |
| perf_counter | 1 passed / 0 failed | 0.8s |

Every target exited 0 with a non-empty run (40 tests total).

**Series-specific live checks** (surfaces with no existing test coverage; release build of this head, same guest):

- **Per-ring consumers**: a default `-s` run spawns all 8 `sched_rec_{0..7}` consumer threads (observed via `/proc/<pid>/task/*/comm`), draining alongside the 8 ring pollers; an 8s window under spinner load wrote a full parquet set (`sched_slice` ~204KB, `stack` ~313KB) with zero missed events in all four classes.
- **`--no-irq`**: exits 0; `irq_slice`/`softirq_slice` parquet absent (present at 44KB/126KB in the default run), sched path unaffected (~125KB of slices), zero missed.
- **`--only-recorder irq`**: exits 0; the IRQ family comes up standalone — `irq_slice` + `softirq_slice` emitted, no sched/stack tables, converts to Perfetto cleanly (187 packets).
- **Continuous flight-recorder + SIGINT drain**: clean — rc 0, full 13-file parquet set, zero missed in all classes, 23K-packet trace written in 4.4s. A main-built control binary behaves identically (rc 0, 13 files, zero missed) — no drain-path delta from the sharding.

For the record: the first run of the SIGINT check appeared to hang. That was the check harness, not systing — the script signaled 6s after launch, but under TCG the BPF setup takes ~15s and the Ctrl-C handler registers only after setup; worse, shells start background children with SIGINT ignored (POSIX no-job-control semantics), so the early signal was silently discarded and the process was simply still waiting for a stop that never arrived. Verified by direct observation: `SigIgn` shows SIGINT ignored at t=1s and caught after startup, and the thread dump at "hang" time shows every thread parked healthily on its wait primitive. The series touches no signal-handling code, and the readiness-gated re-run above drains cleanly on both this head and main. Worth knowing if you Ctrl-C systing in its first seconds under an emulator; not a property of this series.

Per maintainer direction this merges on VM validation; the real-host loss-rate run (does 8-way consumer sharding eliminate the reported sched/IRQ loss?) happens post-merge, and its numbers drive the `NR_RINGBUFS` decision deferred as Improvement 3.

PR #140, opened 2026-07-10
Symbolize stripped Go binaries from .gopclntab

josefbacik:
## Problem

blazesym resolves user-space symbols from ELF `.symtab`/`.dynsym` only. Stripped Go binaries carry neither — but always keep `.gopclntab`, the function table the Go runtime itself needs for tracebacks and GC (`strip` and `-ldflags="-s -w"` cannot remove it). In both trace and `--continuous` output this produces two failure modes:

- **Fully stripped Go binaries** — upstream release builds of kube-state-metrics, etcd, cilium-agent, and most vendor-shipped Go infrastructure — render every user frame as bare hex. Concretely: the official `kube-state-metrics:v2.14.0` image binary reports `stripped` from `file`, has no symbol tables at all per `readelf`, and carries a complete 70k-function `.gopclntab`.
- **Stripped cgo Go builds that retain a few dynamic symbols** (e.g. COS/boringcrypto kubelet builds) are worse than hex: blazesym's nearest-symbol match resolves whole Go text ranges to whichever surviving C/asm symbol precedes them, so profiles confidently claim frames like `OPENSSL_free` ten deep in what is actually Go code. (blazesym treats zero-`st_size` symbols as extending to the next symbol; a separate upstream fix for that behavior is being proposed independently — the two changes compose, and neither depends on the other.)

Stack *walking* is unaffected either way (Go keeps frame pointers): the stacks are deep and correct, only naming fails.

## Fix

New `gopclntab` module + a probe wired into the existing process-dispatcher slot in `StackRecorder::create_symbolizer` (chained after debuginfod, before default handling):

- **Strict activation:** a member is claimed only when its ELF has *no* `.symtab` and a parseable `.gopclntab`. Binaries with a real symbol table keep the richer symtab+DWARF path (source locations, inlined functions) — a test asserts an unstripped build of the same fixture is declined.
- **Parser:** Go 1.16/1.17 and 1.18+ pclntab layouts (little-endian ELF64; the Go 1.20 magic shares the 1.18 layout), name-only resolution — functab binary search by pc, name from funcnametab. Every offset is bounds-checked or capped; malformed tables error or miss, never panic — this runs against arbitrary binaries found on the system. Go ≤1.15 tables and big-endian are declined.
- **Bounded probe I/O:** ELF header + section headers + shstrtab (a few KiB) are read first; the `.gopclntab` byte range is read only for binaries that have one. No whole-file reads.
- **Honest misses:** addresses outside the table (the C/asm portions of cgo text, which are legitimately not Go functions) return unknown and render through the existing contextual labels. When the probe is active, the sparse-dynsym misattribution class is bypassed entirely.
- `--no-gopclntab` reverts to previous behavior (same toggle pattern as `--no-frame-labels` / `--no-gvisor-guest-maps`).

No BPF changes; no schema changes (patch version bump per convention).

## Validation

- **Unit:** synthetic byte-level pclntabs (builder in test code, no toolchain dependency) — boundary pcs, sentinel, both layouts, wrong/swapped magics, truncation sweep at every prefix length, corrupt name offsets, oversized `nfunc`; all miss or error, never panic.
- **Real toolchain:** a `-trimpath -ldflags="-s -w"` fixture built at test time (skips with a notice when `go` is absent): every functab entry resolves back to itself; `main.main`/`main.helper` found by name; the unstripped twin is declined by the probe.
- **Real release binary:** `GOPCLNTAB_TEST_BINARY` hook — kube-state-metrics v2.14.0 resolves 70846/70846 functions.
- **End-to-end:** new `tests/gopclntab_record.rs` integration test records a live stripped Go spinner and asserts its `main.spinHot` frames appear in the output — possible only if the resolver engages through the full process-symbolization flow (member dispatch, file-offset translation, pc lookup). Run alongside the existing integration suite in the VM gate.

---
*Version note: 1.10.8 is taken by another in-flight PR.*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,7 +3430,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.6"
+version = "1.10.7"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "systing"
-version = "1.10.6"
+version = "1.10.7"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/gopclntab.rs
+++ b/src/gopclntab.rs
@@ -1,0 +1,876 @@
+//! Fallback symbolization for stripped Go binaries via `.gopclntab`.
+//!
+//! Stripped Go binaries carry no ELF symbol table, but the Go runtime needs
+//! its function table (pclntab) at run time for tracebacks and GC, so
+//! `strip` and `-ldflags="-s -w"` always leave `.gopclntab` behind. blazesym
+//! only reads `.symtab`/`.dynsym`, which has two failure modes on such
+//! binaries:
+//!
+//! * fully stripped (e.g. upstream kube-state-metrics, etcd, cilium-agent
+//!   release builds): every user frame renders as bare hex;
+//! * stripped cgo builds that retain a handful of dynamic symbols (e.g.
+//!   COS/boringcrypto kubelet builds): blazesym's nearest-symbol match
+//!   attributes whole Go text ranges to whichever C/asm symbol happens to
+//!   precede them, producing confidently *wrong* names.
+//!
+//! [`try_gopclntab_resolver`] probes a process member for exactly that
+//! shape — no `.symtab`, `.gopclntab` present — and returns a resolver that
+//! answers name lookups from the pclntab instead. Binaries with a real
+//! symbol table never take this path, so the richer symtab+DWARF resolution
+//! (source locations, inlined functions) is unaffected.
+//!
+//! Only function names are resolved here (no file/line, no inline
+//! expansion): that is the difference between an unreadable profile and a
+//! readable one, and pclntab line tables can be layered on later if wanted.
+//!
+//! Format reference: `go/src/debug/gosym/pclntab.go` and
+//! `go/src/internal/abi/symtab.go`. Layouts for Go 1.16/1.17 (`ver116`) and
+//! Go 1.18+ (`ver118`; the Go 1.20 magic shares the layout) are supported.
+//! The Go 1.2–1.15 layout and big-endian tables are rejected, in which case
+//! symbolization behaves exactly as before this module existed.
+
+use std::ffi::OsString;
+use std::fmt;
+use std::fs::File;
+use std::os::unix::fs::FileExt as _;
+use std::path::Path;
+
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+
+use blazesym::helper::ElfResolver;
+use blazesym::symbolize::FindSymOpts;
+use blazesym::symbolize::Reason;
+use blazesym::symbolize::Resolve;
+use blazesym::symbolize::ResolvedSym;
+use blazesym::symbolize::SrcLang;
+use blazesym::symbolize::Symbolize;
+use blazesym::symbolize::TranslateFileOffset;
+use blazesym::Addr;
+
+/// Go 1.2 through 1.15. Predates every Go release still in support; not
+/// worth carrying the layout. Recognized only to be rejected explicitly.
+const GO12_MAGIC: u32 = 0xFFFF_FFFB;
+/// Go 1.16 and 1.17.
+const GO116_MAGIC: u32 = 0xFFFF_FFFA;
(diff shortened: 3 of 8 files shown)


josefbacik:
# rust-bpf-code-reviewer — pre-push review of this PR's diff

*Mechanics: executed per `.claude/agents/rust-bpf-code-reviewer.md` by the authoring session over the full branch diff (`git diff main...HEAD`, final rev of this PR), five passes (structure / Rust quality / BPF quality / schema & versioning / integration). Findings from the review were fixed in-branch before this PR was opened; they are disclosed below rather than left as comments.*

### Summary

A self-contained fallback resolver for stripped Go binaries: a new `gopclntab` module (parser + `Resolve` impl + probe) and a dispatcher chain in `StackRecorder::create_symbolizer`. Activation is strictly gated (no `.symtab` AND `.gopclntab` present), all failure paths degrade to today's behavior, and the parse/lookup layer is panic-free against arbitrary input. No BPF-side changes; no schema changes.

### Critical Issues

None outstanding.

- Schema & versioning gates checked: no table or `create_schema()` change → no `SCHEMA_VERSION` increment, no `SCHEMA_CHANGES.md` entry required; non-schema change carries the required **patch** bump (1.10.6 → 1.10.7).
- Recorder-consistency gates: no new recorder or per-event table; the new integration test attributes via `stack_sample.utid → thread.utid/tid`, consistent with the schema's join discipline.

### Improvements (found during review, fixed in-branch)

1. **Probe I/O was unbounded.** The first version of `try_gopclntab_resolver` read entire member files before checking sections — every symbol-table-less member of every profiled process (all stripped `.so`s, ~100 MiB Go binaries) per symbolization pass. Reworked to targeted reads: ELF header + section headers + shstrtab (a few KiB, explicitly capped) and then exactly the `.gopclntab` byte range, only for binaries that have one.
2. **Panic path in lookup accessors.** `functab` word reads used `.expect()` justified by a parse-time invariant enforced at a distance; a future edit could turn a corrupt table into a symbolization-phase panic. Accessors now return `Option` and `find_func` propagates — corrupt tables degrade to honest misses.
3. **Field/method name collision.** `functab_field` (field) vs `functab_field()` (method) — method renamed to `functab_word`.

### Suggestions (noted, not blocking)

- `nameoff` is `int32` in Go's layout and is read here as unsigned: a negative offset (not produced by any real toolchain) becomes a large value and fails the bounds check into a miss — the honest outcome, so no separate sign handling.
- The `ver116` layout is tested with `ptrsize == 8` only; 32-bit Go binaries are out of scope for this profiler's targets and fail safe (probe declines non-ELF64).
- Per-process member probing re-reads section headers for shared members (e.g. the same stripped library mapped by many processes). Bounded at a few KiB per probe; a (dev, inode)-keyed cache is a possible follow-up if it ever shows up in profiles.

### Positive Notes

- The activation condition guarantees the richer symtab+DWARF path (source locations, inlined functions) is never shadowed — including a regression test that an unstripped build of the same fixture is declined.
- Malformed-input handling is tested aggressively: truncation sweep at every prefix length, corrupt name offsets, oversized `nfunc`, wrong/byte-swapped magics — all error or miss, never panic.
- Validation is layered exactly right for a format parser: synthetic byte-level tables for edge cases (no toolchain dependency), a real-toolchain stripped fixture (every functab entry resolves; `main.main`/`main.helper` by name), an arbitrary-binary hook (`GOPCLNTAB_TEST_BINARY`), and a live end-to-end recording test that can only pass if the resolver engages through the process-symbolization flow.

josefbacik:
Design note on dispatcher precedence, for the record (two questions that came up in review):

**Does the probe interfere with debuginfod?** No. The process dispatcher tries debuginfod first when it is configured: a debuginfod hit returns that resolver immediately and the gopclntab probe never runs for that member, so debuginfod-served symbolization is unchanged. When debuginfod misses (no build-id note in the binary — common for pure-Go static builds — or the server has nothing for that build-id), the member falls to the gopclntab probe instead of straight to the default ELF path, which upgrades the miss case for stripped Go from hex to names. Debuginfod errors propagate exactly as before.

**Do plain (non-Go) ELF executables still symbolize?** Yes, unchanged. The probe claims a member only when all of the following hold: little-endian ELF64, no `.symtab` section, a `.gopclntab` section present, and the table parses with a valid Go magic. Plain ELF fails the `.gopclntab` check; unstripped Go fails the no-`.symtab` check and keeps the richer symtab+DWARF path (regression-tested: an unstripped build of the test fixture must be declined); anything malformed fails the parse. Every failure path declines to blazesym's default handling, so the only binaries whose behavior changes are stripped Go binaries — the ones that previously rendered as hex or misattributed names. `--no-gopclntab` restores prior behavior entirely.

Precedence summary: `debuginfod (when configured) → gopclntab probe (strict activation) → default ELF/DWARF handling`.

PR #141, opened 2026-07-10
Poll missed-event counters into per-class counter tracks

josefbacik:
Follow-up to #139, implementing the in-trace loss counters declared there (Improvement 2).

## Problem

Event loss is only reported in the exit summary today. When a run does lose events (the situation #139 addressed), the summary tells you *how many* but not *when* — and for long or continuous recordings, "when" is the difference between "one burst during the incident window" and "steady bleed the whole run". The summary is also easy to lose entirely (scrollback, wrapper scripts, remote runs).

## Change

A dedicated poller thread samples the per-class `missed_events` BPF map once a second and streams the cumulative totals into `Missed <class> events` counter tracks:

- **`MissedEventsRecorder`** follows the `SysinfoRecorder` streaming shape exactly: lazily-created track per class, shared counter writer (the `counter`/`counter_track` parquet files stay single-writer), track IDs from the shared generator so they never collide with the perf-counter or CPU-frequency tracks.
- **Class gating matches the exit summary**: only classes whose subsystems are active this run are polled (sched/IRQ, stack, probe, cache always; network/packet/poll with networking; marker with markers; memory with memory tracing). Track naming matches the summary lines, so the trace and the printout name the same things.
- **The final sample carries the settled totals**: the poller takes one last sample after observing the shutdown flag, which is set only after tracing is disabled and the ringbuf consumer threads are joined — so the last track value equals the exit-summary number.
- **Continuous mode**: samples buffer in the flight-recorder ring like any other event, so the window rotation applies to loss counters too.
- `dump_missed_events` now takes the map (via `MapCore`) instead of the whole skeleton, so the poller can use a duplicated `MapHandle` while the exit summary keeps the same summing code.

Always-on by design: it's a handful of map lookups per second, and a flat zero track is itself useful signal — "this run lost nothing" becomes verifiable from the trace alone instead of an absence in scrollback.

## Why 1s / cumulative values

The map values are cumulative totals, so one-second steps localize a burst to the second while adding negligible trace volume (at most 9 counter rows/sec). Rendering the cumulative value (not deltas) means the counter graph shows loss as steps, monotonicity is a checkable property, and the final value is directly comparable to the summary printout.

## Validation

- Unit tests: parquet readback of streamed tracks (per-class track creation, exit-summary naming, cumulative values in ts order) and the continuous-mode ring path (buffer, drain, land).
- New integration test (`tests/missed_events.rs`, root/BPF, via `scripts/run-integration-tests.sh missed_events`): records a real trace, then asserts per-class track presence with ≥2 samples and monotonic values, disabled classes leave no tracks, track IDs stay unique across all counter recorders, and zero counter rows orphan their track.
- Full integration suite green in the same VM rig used for #139 (all six standard targets plus the new one) — results in a comment below.

Version: 1.10.10.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.9"
+version = "1.10.10"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.10.9"
+version = "1.10.10"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/session_recorder.rs
+++ b/src/session_recorder.rs
@@ -145,6 +145,7 @@ pub struct SessionRecorder {
     pub stack_recorder: Mutex<StackRecorder>,
     pub perf_counter_recorder: Mutex<PerfCounterRecorder>,
     pub sysinfo_recorder: Mutex<SysinfoRecorder>,
+    pub missed_events_recorder: Mutex<MissedEventsRecorder>,
     pub probe_recorder: Mutex<SystingProbeRecorder>,
     pub network_recorder: Mutex<NetworkRecorder>,
     pub memory_recorder: Mutex<MemoryRecorder>,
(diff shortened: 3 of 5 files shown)


josefbacik:
## rust-bpf-code-reviewer review

Mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition, executed by the authoring session over the full diff vs `main` (802b641) before push, per this repo's review convention.

### Summary

Adds in-trace visibility for per-class event loss by polling the existing `missed_events` map into counter tracks. The implementation deliberately mirrors the established `SysinfoRecorder` streaming pattern and the sysinfo poller thread; there are no BPF-side changes, no schema changes, and no new files in the parquet output. Test coverage is proportionate: unit-level parquet readback plus an integration test with integrity assertions.

### Critical Issues

None found. The gates were checked explicitly:

- **Schema**: no new tables or columns — the new rows ride the existing `counter`/`counter_track` tables and `create_schema()` is untouched, so `SCHEMA_VERSION` correctly stays at 11 and no `SCHEMA_CHANGES.md` entry is required.
- **Versioning**: non-schema change → patch bump present (1.10.6 → 1.10.7, lock file updated).
- **Recorder consistency**: loss counters are system-wide (like the CPU-frequency tracks), so the per-thread `utid` attribution rule doesn't apply to the emitted rows; `ThreadAwareRecorder` is implemented because the `SystingRecordEvent` supertrait requires it, same as `SysinfoRecorder`. The integration test asserts the analogous integrity properties: zero counter rows orphaning their `counter_track`, and track-id uniqueness across every counter-emitting recorder.

### Improvements (applied during review)

- The poller originally slept the full poll interval between samples, which could delay shutdown by up to a second; it now waits in 100ms slices re-checking the shutdown flag (`MISSED_EVENTS_SHUTDOWN_CHECK_MS`), so stop isn't held up by the sampling cadence.

### Suggestions

- None blocking. If a use case ever wants to shed even the 1Hz sampling, a `--no-missed-counters` flag is a trivial follow-up; the always-on rationale (zero-loss as a verifiable signal rather than an absence) is documented at the spawn site.

### Gate finding (fixed before push)

The VM integration gate caught one correctness issue the review passes above missed: the new tracks initially wrote `unit: "events"`, which is outside the validated counter-unit vocabulary (`VALID_COUNTER_UNITS`) — every trace-validating integration test failed with `InvalidEnumValue` on `counter_track.unit`. Fixed to `unit: "count"`, the same unit the perf-counter tracks already use (and the value that maps to perfetto's first-class `UNIT_COUNT` instead of a custom unit name). Both the unit test and the integration test now assert the unit value, so a vocabulary regression is caught before the VM gate. The verdict above was re-checked against the amended diff (unit fix, the two test assertions, and the version bump); no other findings.

### Positive Notes

- The settled-totals invariant is real, not aspirational: `shutdown_signal.store(true)` happens in `run_tracing_loop` only after `tracing_enabled` is cleared and the ringbuf consumer threads are joined, so the poller's final sample provably equals the exit-summary numbers, and the comment at the join site documents exactly that ordering.
- `SharedCollector`'s documented sequential-finish contract is respected: the new handle is finished third in the same single-threaded finish chain, and the `Arc::try_unwrap` last-handle finalization holds for three handles the same as it did for two.
- `dump_missed_events` taking `&dyn MapCore` keeps a single summing implementation for both the exit summary and the poller instead of duplicating the per-CPU fold.

josefbacik:
VM validation on this exact head (`6cc5a9e`): full integration suite green, including the new `missed_events` target.

**Seven-target integration gate** (virtme-ng guest, 6.12 BPF-enabled kernel, TCG/no-KVM, 4 vCPU / 8 GB):

| target | result | in-test time |
|---|---|---|
| trace_validation | 31 passed / 0 failed | 2305s |
| memory_recorder | 5 passed / 0 failed | 558s |
| analyze_commands | 1 passed / 0 failed | 888s |
| stream_unix | 1 passed / 0 failed | 123s |
| network_throughput | 1 passed / 0 failed | 52s |
| perf_counter | 1 passed / 0 failed | 0.8s |
| missed_events (new) | 1 passed / 0 failed | 55s |

41 tests total, every target exited 0 with a non-empty run.

**What the new target verifies live** (recording a real trace as root in the VM, then reading it back through DuckDB): each always-on class produces exactly one `Missed <class> events` track with at least baseline+final samples and monotonically non-decreasing cumulative values; classes whose subsystems are disabled in the run leave no tracks behind; track IDs stay unique across all counter recorders; zero counter rows orphan their track; and units stay inside the validated counter-unit vocabulary.

**Validation history for the record:** an earlier run of this suite (on the pre-rebase content of this branch) caught the tracks initially writing `unit: "events"` — outside `VALID_COUNTER_UNITS` — which failed all seven trace-validating tests with `InvalidEnumValue`. Fixed to `unit: "count"` (the perf-counter precedent) with unit assertions added at both test levels; details in the review comment above. The suite then ran green on the fixed pre-rebase commit and again 7/7 at this exact rebased head, so the numbers above carry no caveats.

PR #143, opened 2026-07-10
Extract Go symbolization into a publishable gopclntab crate

josefbacik:
## What

The Go symbolization added in #140 moves into `crates/gopclntab`, a zero-dependency workspace member, so other projects can resolve function names in stripped Go binaries without pulling in a profiler. The crate is publishable as-is (`cargo publish --dry-run` passes; LICENSE/README/keywords included) — actual publication stays a maintainer action.

## API split

- **Crate (`gopclntab`)** — mechanics only, `std`-only, zero dependencies, no profiler types in the surface:
  - `GoPclntab::parse(bytes, text_vaddr)` + `find_func(pc) -> GoFunc{name, entry, size}` + `func_count`/`func_entry` over raw table bytes;
  - `ElfPclntab::from_path`/`from_bytes` — bounded ELF reads (headers + the table itself, never whole binaries) that also report `has_symtab`;
  - a real error taxonomy: `Error::Unsupported` (Go ≤1.15 layouts, big-endian, unknown magics) vs `Error::Malformed` (truncation, out-of-bounds offsets) — callers can distinguish "not for me" from "broken". `anyhow` stays out of the public surface.
  - **Policy is deliberately the caller's**: the ELF layer reports whether a symbol table exists rather than deciding what to prefer. A profiler wants the pclntab only for stripped binaries; other tools may always want it.
- **systing (`src/gopclntab_resolver.rs`, renamed to avoid shadowing the crate)** — the blazesym `Resolve` implementation and the stripped-Go activation policy (no `.symtab` AND parseable `.gopclntab`), unchanged in behavior.

## Workspace CI

`cargo test`/`cargo clippy` were single-crate; a workspace would have silently narrowed CI to the root crate. Both gain `--workspace` (`cargo fmt --all` already was workspace-wide). Verified locally with the workflow's commands:

- `cargo test --workspace`: both members demonstrably run — `gopclntab` lib suite **14 passed** (including the Go-toolchain fixture test: stripped build fully resolves, unstripped twin reports its symtab; and a doc-test), systing lib suite **378 passed** (the 12 moved tests now live with the crate; totals reconcile exactly).
- `cargo clippy --workspace --no-default-features -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- `cargo doc -p gopclntab --no-deps`: no warnings. `cargo publish --dry-run`: packages and compiles standalone.

Real-binary validation through the crate path: kube-state-metrics v2.14.0 resolves 70846/70846 functions (`GOPCLNTAB_TEST_BINARY` hook).

## Behavior

None changed: the extraction is code motion plus API re-surfacing. The end-to-end integration test (records a live stripped Go workload and requires its frames to resolve) passes on this branch, as does the full integration suite.

---
*Version note: 1.10.8 is taken by another in-flight PR.*

### DIFF
--- a/crates/gopclntab/src/elf.rs
+++ b/crates/gopclntab/src/elf.rs
@@ -1,0 +1,313 @@
+//! Locate and extract `.gopclntab` from ELF binaries.
+//!
+//! Reads are targeted and bounded — ELF header, section headers, and the
+//! section-name string table first (a few KiB, explicitly capped), then
+//! exactly the pclntab byte range — so probing large binaries (Go binaries
+//! run to hundreds of MiB) never reads whole files.
+//!
+//! Only little-endian ELF64 is supported; anything else reports "not a
+//! candidate" (`Ok(None)`) rather than an error, as do binaries without a
+//! `.gopclntab` section. Policy is deliberately left to the caller: this
+//! module reports whether a symbol table is present ([`ElfPclntab::
+//! has_symtab`]) but does not decide whether the pclntab *should* be used
+//! — a profiler may only want it for stripped binaries, while other tools
+//! may always want it.
+
+use std::fs::File;
+use std::path::Path;
+
+use crate::Error;
+use crate::GoPclntab;
+
+/// Caps on metadata reads. A binary whose metadata exceeds these is
+/// skipped (`Ok(None)`), not mis-parsed; a `.gopclntab` larger than the
+/// cap is reported as malformed (real tables in very large binaries run
+/// to tens of MiB).
+const MAX_SECTION_HEADERS: u64 = 4096;
+const MAX_SHSTRTAB_SIZE: u64 = 1 << 20;
+const MAX_PCLNTAB_SIZE: u64 = 1 << 30;
+
+const SHT_SYMTAB: u32 = 2;
+
+/// A `.gopclntab` extracted from an ELF binary, with the context a caller
+/// needs to decide how to use it.
+#[derive(Debug)]
+pub struct ElfPclntab {
+    /// The parsed function table.
+    pub table: GoPclntab,
+    /// Whether the binary also carries a `.symtab`. Symbolizers typically
+    /// prefer the symbol table (and its associated DWARF) when present —
+    /// the pclntab is most valuable when this is `false`, i.e. the binary
+    /// is stripped.
+    pub has_symtab: bool,
+}
+
+/// A bounded random-access byte source: implemented for files (pread) and
+/// in-memory slices.
+trait ReadAt {
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()>;
+}
+
+impl ReadAt for File {
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()> {
+        std::os::unix::fs::FileExt::read_exact_at(self, buf, offset)
+    }
+}
+
+impl ReadAt for &[u8] {
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> std::io::Result<()> {
+        let start = usize::try_from(offset).ok().filter(|start| {
+            start
+                .checked_add(buf.len())
+                .is_some_and(|end| end <= self.len())
+        });
+        match start {
+            Some(start) => {
+                buf.copy_from_slice(&self[start..start + buf.len()]);
+                Ok(())
+            }
+            None => Err(std::io::Error::from(std::io::ErrorKind::UnexpectedEof)),
+        }
+    }
+}
+
+impl ElfPclntab {
+    /// Extract and parse `.gopclntab` from the ELF file at `path`.
+    ///
+    /// Returns `Ok(None)` when the file is not a candidate: not a
+    /// little-endian ELF64, no section headers, or no `.gopclntab`
+    /// section. Returns an error for I/O failures and for tables that
+    /// exist but do not parse.
+    pub fn from_path(path: impl AsRef<Path>) -> Result<Option<Self>, Error> {
+        let file = File::open(path)?;
+        Self::from_read_at(&file)
+    }
+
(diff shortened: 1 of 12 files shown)


josefbacik:
# rust-bpf-code-reviewer — pre-push review of this PR's diff

*Mechanics: executed per `.claude/agents/rust-bpf-code-reviewer.md` by the authoring session over the full branch diff (this PR's extraction commit atop #140), five passes (structure / Rust quality / BPF quality / schema & versioning / integration). Findings were fixed in-branch before the PR opened and are disclosed below.*

### Summary

Code motion with a deliberate API boundary: parsing and ELF mechanics become a zero-dependency publishable crate; the blazesym integration and the stripped-Go activation policy stay in systing (module renamed `gopclntab_resolver` to avoid shadowing the crate). No behavior change intended, and the live end-to-end recording test plus the full integration suite back that up. No BPF changes; no schema changes.

### Critical Issues

None outstanding.

- Schema & versioning gates: no table or `create_schema()` change → no `SCHEMA_VERSION`/`SCHEMA_CHANGES.md` requirements; the systing patch bump is present (1.10.7 → 1.10.9; 1.10.8 belongs to another in-flight PR, noted in the PR body). The new crate starts at 0.1.0.
- Workspace-CI blindness check (the extraction's main hazard): bare `cargo test`/`cargo clippy` would have run the root crate only, silently dropping the new member from CI. Both workflow lines gain `--workspace`, and the local run of the exact commands shows both members' suites executing (14 + 378, totals reconciling with the pre-extraction count).

### Improvements (found during review, fixed in-branch)

1. **Needless double-shim in the ELF byte-source abstraction.** The first draft disambiguated `FileExt::read_exact_at` from the local `ReadAt` trait through a second private trait; a single fully-qualified call does it. Removed.
2. **Public error surface**: the moved code used `anyhow` internally; a published crate should not force it. Replaced with a two-variant taxonomy (`Unsupported` vs `Malformed`) that callers can act on, plus `Io`.

### Suggestions (noted, not blocking)

- `func_entry(i)` covers table enumeration; a proper iterator (`funcs()`) would be a natural semver-compatible addition when someone needs it.
- The crate's `rust-version = "1.89"` matches the workspace floor rather than the true minimum; conservative and honest to what is tested, can be lowered later if a consumer asks.

### Positive Notes

- The `has_symtab`-reported-not-decided split is the right reusability boundary: systing's preference for symtab+DWARF is policy, and the crate does not impose it on other consumers.
- Docs are load-bearing: the crate-level example is a doc-test (compiles in CI), `cargo doc` is warning-free, and `cargo publish --dry-run` builds the packaged copy standalone — publishability is verified, not assumed.
- Test ownership moved with the code: the synthetic-table suite, toolchain fixture (now also asserting the unstripped twin reports its symtab), and the real-release-binary hook all live with the crate; systing keeps the end-to-end recording test that proves the composed behavior.

crates/gopclntab/src/elf.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+        // \x7fELF, 64-bit (class 2), little-endian (data 1).
+        if ehdr[..4] != [0x7f, b'E', b'L', b'F'] || ehdr[4] != 2 || ehdr[5] != 1 {
+            return Ok(None);
+        }
+        let e_shoff = u64::from_le_bytes(ehdr[0x28..0x30].try_into().unwrap());
+        let e_shentsize = u64::from(u16::from_le_bytes(ehdr[0x3a..0x3c].try_into().unwrap()));
+        let e_shnum = u64::from(u16::from_le_bytes(ehdr[0x3c..0x3e].try_into().unwrap()));
+        let e_shstrndx = usize::from(u16::from_le_bytes(ehdr[0x3e..0x40].try_into().unwrap()));
+        // e_shnum == 0 covers both section-header-stripped binaries and
```
josefbacik: Do not use magic numbers, these should be constants. In fact there should be a crate for Elf parsing that we could likely use to get these constants or do the Elf parsing bit completely.

crates/gopclntab/src/elf.rs
```diff
@@ -0,0 +1,313 @@
...
+use std::path::Path;
+
+use crate::Error;
+use crate::GoPclntab;
+
+/// Caps on metadata reads. A binary whose metadata exceeds these is
+/// skipped (`Ok(None)`), not mis-parsed; a `.gopclntab` larger than the
+/// cap is reported as malformed (real tables in very large binaries run
+/// to tens of MiB).
+const MAX_SECTION_HEADERS: u64 = 4096;
```
josefbacik: I don't like arbitrary size caps, is there ever a case where `.gopclntab` stubs are larger? I don't want to reject valid things just because they're large.

crates/gopclntab/src/elf.rs
josefbacik: Done in 06e025e — switched the ELF layer to the `object` crate (read core + ELF format only, same crate the main systing crate already uses), which removes the hand-rolled parsing and every raw offset with it. Reads stay targeted via object's `ReadCache` (headers plus exactly the pclntab byte range, never the whole binary). The crate's dependency line goes from zero to one; README updated to say so.

crates/gopclntab/src/elf.rs
josefbacik: Agreed, and 06e025e removes all three caps. To your question: real pclntabs do get big — they scale with text (kube-state-metrics carries ~9 MiB for 70k functions; monolithic Go binaries run to tens of MiB) — and nothing in the format bounds them, so any threshold would eventually reject a valid binary. The replacement is a consistency check instead of a size check: the section's claimed file range is validated against the input's actual length, so the only "too large" rejected now is a header claiming bytes past end-of-input — inconsistent metadata, not bigness (that case is pinned by a test that corrupts a real fixture's `sh_size`). The other two caps guarded the hand-rolled header walking and went away with it when `object` took over parsing.

PR #144, opened 2026-07-10
Add --symbolize-names-only to bound symbolization memory

josefbacik:
Whole-system captures on hosts running many freshly-built debug binaries (CI nodes mid-build) can blow their memory limit during end-of-trace symbolization and get killed at the finish line: blazesym retains parsed debug info per binary with no eviction, and with code-info + inlined-fn resolution one `-g` binary holds hundreds of MiB. The RSS valve can't fully contain it — growth between checks is unbounded (a large parse happens inside one symbolize call), and the anti-thrash floor deliberately stops rebuilding when `malloc_trim` can't return fragmented arenas. Notably this term is invariant to sample rate: one sample into a fresh binary costs the same parse as ten thousand, so lowering the sampling frequency doesn't help.

Measured with blazesym 0.2.3 on systing's own debug build (839 MiB of `.debug_*`), 2000 addresses spread across `.text`:

| config | resident |
|---|---|
| code-info + inlined fns (current) | 464 MiB |
| DWARF names only | 220 MiB |
| symbol-table names only | **4.3 MiB** |

Across three such binaries: 928 MiB → 10 MiB, **byte-identical function name strings**, ~10× faster. Symbol tables carry the same names DWARF does; DWARF funds `[file:line]` suffixes and inline expansion, which aggregate flame-graph pipelines never render.

**Change:**
- `--symbolize-names-only`: user-space symbols resolve from ELF symbol tables only — `Process` and bridged-`Elf` sources get `debug_syms = false`, code-info/inlined-fns off, debuginfod skipped (with a note when both flags are set). `.gopclntab` handling for stripped Go binaries is unaffected (it provides names at symbol-table-like cost). Default behavior unchanged: interactive traces keep full fidelity. This is also what earns continuous profiling back on the host pools where it had to be disabled to stop the kills — with the symbolization term bounded, those hosts can be re-enabled rather than left dark.
- Anon-RSS notes at symbolization phase boundaries (start / after exited-process pass / done), so a memory blowup or an OOM kill's last stderr line attributes to the exited-process pass vs the live-process pass without host tooling.

**Testing:** `cargo test --lib` 381/381; new end-to-end test (`test_finish_inner_names_only_still_resolves_live`) self-symbolizes via `/proc/self` and asserts the live frame resolves *and* carries no `[file:line]` suffix (the test binary has full DWARF, so a suffix means the flag silently stopped working); dead-process path asserted unchanged. Clippy clean.

Version: 1.10.11.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.10"
+version = "1.10.11"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.10.10"
+version = "1.10.11"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,6 +77,16 @@ struct Command {
     /// to symbol-table-only resolution, which renders them as hex)
     #[arg(long)]
     no_gopclntab: bool,
+    /// Resolve user-space symbols from ELF symbol tables only: skips DWARF
+    /// debug info, source line info, inlined-function resolution, and
+    /// debuginfod. Function names are unchanged for binaries that carry a
+    /// symbol table; frames lose their "[file:line]" suffix and inlined
+    /// frames collapse into their caller. Bounds symbolization memory on
+    /// hosts running many freshly-built debug binaries (CI), where parsed
+    /// debug info — cached per binary with no eviction — otherwise reaches
+    /// GiBs.
+    #[arg(long)]
+    symbolize_names_only: bool,
     /// Disable scheduler event tracing (sched_* tracepoints and scheduler event recorder)
     #[arg(long)]
     no_sched: bool,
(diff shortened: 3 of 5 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** (repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed over the full diff before push; findings below were applied pre-push)

### Summary
Small, well-layered change mirroring the existing `no_gopclntab` plumbing (CLI → Config → `configure_recorder` → recorder setter). Userspace-only; no BPF changes. The flag's semantics are documented at every layer with the memory tradeoff spelled out.

### Critical Issues
None.
- Schema/versioning gate: no schema changes → no `SCHEMA_VERSION`/`SCHEMA_CHANGES.md` requirement; non-schema change carries the required **patch** bump (1.10.10 → 1.10.11, Cargo.toml + Cargo.lock consistent). Frame strings with code-info off render as the pre-existing location-less variant (`name (module) <addr>`), so no downstream format change.

### Improvements (applied)
- The new test originally asserted only that a live frame resolves — which also passes under full DWARF, so it couldn't catch the flag silently breaking. Added an assertion that the frame carries no `.rs:` location suffix (the test binary has full DWARF, making the suffix's absence a real discriminator).
- `--enable-debuginfod --symbolize-names-only` silently ignored debuginfod while its "Debuginfod enabled" banner still printed; added a startup note stating the precedence.

### Suggestions (applied)
- Trimmed a comment on the `debug_syms` assignment that restated the code; kept the pointer to the `names_only` field docs.

### Positive Notes
- `create_symbolizer` is also the valve's rebuild path, so names-only survives mid-run symbolizer rebuilds for free.
- The bridged-Elf site (`symbolize_user_addr`) getting `debug_syms=false` too means sandbox island bridging can't reintroduce DWARF parsing under the flag.

Mechanics: review performed by the PR author's tooling against the checked-in agent definition; all listed findings were fixed before the branch was pushed. `cargo test --lib` 381/381, clippy clean.

PR #145, opened 2026-07-10
Normalize stack storage to uniform root-to-leaf order (schema v12)

josefbacik:
## Summary

Recorded stacks are a concatenation of three segments — python, user,
kernel — and through schema v11 the segments disagreed on direction. The
BPF unwinder delivers frames leaf-first and
`filter_and_reverse_{user,kernel}_stack` reverse those segments to
root-to-leaf at capture, but the python segment kept pystacks' leaf-first
buffer order. So `frame_ids` was `[py leaf→root][user root→leaf][kernel
root→leaf]` — not one coherent sequence, and the readers disagreed about
which direction they were getting:

- `leaf_name` took the combined array's **first** entry: the python leaf
  when python frames are present, but the **root** frame for native-only
  stacks — despite the column name.
- `systing analyze flamegraph` (and the MCP `flamegraph` tool) reversed
  the array under a leaf-to-root assumption, emitting **inverted folded
  lines** (`leaf;…;root`) for native stacks, and scrambling blended
  python stacks either way (no single reverse can fix a mixed-direction
  array).
- The perfetto exporter passed the array through — correct for native
  segments (`Callstack.frame_ids` is "Bottom frame first" per the proto),
  but **inverted python segments** in .pb output.

## Fix

Normalize once, at the single construction point both recorders share:
`Stack::new` now reverses the python segment too, so every segment is
stored root-to-leaf and `frame_ids`, folded output, and exports all read
one coherent direction. `leaf_name` comes from the **last** frame (the
innermost, executing one), and `format_folded_stack` emits storage order
— which is already the folded convention — instead of reversing.

Schema version bumps to 12 to mark the data-semantics change (no table
shapes change); `SCHEMA_CHANGES.md` documents the details, including
what happens to pre-v12 databases.

## Compatibility

- **Native-only stacks are byte-identical** in storage either way — the
  user/kernel segments were already root-to-leaf. For those, this PR
  only fixes the readers: folded output from *old* databases comes out
  right-side-up too.
- **Pre-v12 python segments stay leaf-first** in old databases — they
  can't be re-ordered after the fact without classifying frames by
  name. `_traces.systing_version` is the per-trace discriminator, with a
  caveat documented in SCHEMA_CHANGES.md: it is stamped at
  parquet-to-DuckDB conversion time with the converting binary's
  version, so it names the recorder only when recording and conversion
  happen in the same run (the normal pipeline). Embedding the recorder's
  version in the parquet directory itself is planned follow-up.
- Perfetto .pb exports become fully spec-correct (python segments were
  the violation).

## Verification

- `cargo test --lib`: 379 passed / 0 failed. The old
  `test_format_folded_stack_reversal` fixture is itself a demonstration
  of the bug: storage has been root-first for native frames since
  capture-time reversal was introduced, so the reverse-on-output it
  asserted produced `leaf;…;root` on real data. Replaced by
  `test_format_folded_stack_storage_order`, plus new tests pinning the
  construction invariant (`test_stack_new_reverses_py_stack`) and the
  leaf_name semantics (`test_emit_stack_record_leaf_name_is_last_frame`).
- `cargo clippy` clean in both the CI configuration
  (`--no-default-features -- -D warnings`) and `--all-targets`;
  `cargo fmt --check` clean.
- Frame-order convention for the perfetto export checked against the
  vendored proto (`perfetto_protos-0.48.1`,
  `protos/perfetto/trace/profiling/profile_common.proto`: "Frames of
  this callstack. Bottom frame first.").
- Swept the other order-sensitive consumers: the spill
  serializer/deserializer round-trips the normalized order (no
  re-reverse), symbolization and the DuckDB import/merge SQL are
  order-preserving (`ORDER BY u.idx`), and the memory recorder shares
  `Stack::new` so it inherits the invariant.

## Known adjacent issue (deliberately not in this PR)

`systing-util`'s perfetto **.pb import** (`get_or_create_callsite`)
builds its callsite tree assuming leaf-first `Callstack.frame_ids`
input, so spec-compliant .pbs — including systing's own exports — import
with inverted parent chains, and its `depth` numbering is inverted under
either assumption. That path reads .pb files, not parquet/DuckDB, so
nothing here changes its behavior; it needs its own fix and will be
reported separately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.10.11"
+version = "1.11.0"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.10.11"
+version = "1.11.0"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -211,3 +211,47 @@ strings there) — the normalization happens at DuckDB import time.
   (`trace_id, id, frame_names, depth, leaf_name`). Ad-hoc queries can use it as
   a drop-in replacement; for hot paths join `frame` directly, since the view
   re-aggregates names per row.
+
+## Schema Version 12 (systing 1.11.0) — 2026-07-09
+
+Stack frame ordering is now uniform. Recorded stacks are a concatenation of
+three segments — python, user, kernel — and through v11 the segments disagreed
+on direction: the BPF unwinder's user and kernel segments were reversed to
+root-to-leaf at capture, while the python segment kept pystacks' leaf-first
+buffer order. `leaf_name` took the first entry of the combined array, which is
+the python leaf when python frames are present but the ROOT frame for
+native-only stacks. (The v11 entry's claim that ordering was "leaf-to-root"
+described the reader's assumption, not the data: native segments have been
+root-to-leaf since capture-time reversal was introduced.)
+
+No table shapes change. The version bump marks the data-semantics change so
+tools can distinguish uniformly-ordered databases.
+
+### Changed semantics
+- `stack.frame_ids` (and `frame_names` in parquet / the `stack_frames` view):
+  the python segment is now stored root-to-leaf like the user and kernel
+  segments, making the whole array one coherent root-to-leaf sequence
+  (outermost caller first, innermost executing frame last). Segment layout is
+  unchanged: python (outermost), then user, then kernel. v11 and older
+  databases keep the mixed order they were written with — the python segment
+  cannot be re-ordered after the fact without classifying frames by name
+  (note for anyone attempting that: root-side markers — `<module>`, threading
+  bootstraps, and on CPython 3.12+ an `<interpreter trampoline>` frame that
+  sits beyond `<module>` — fall in the run's root-side half, not necessarily
+  at its literal end); native-only stacks are identical either way. Importing an older trace into
+  a v12 database does not re-order it either. `_traces.systing_version` is
+  the per-trace discriminator, with one caveat: it is stamped at
+  parquet-to-DuckDB conversion time with the converting binary's version
+  (then preserved across database-to-database imports), so it names the
+  recorder only when recording and conversion happen in the same run — true
+  for the normal pipeline, not for a retained parquet directory converted
+  later by a newer binary. Embedding the recorder's version in the parquet
+  directory itself is planned follow-up.
+- `stack.leaf_name` now holds the name of the innermost executing frame (the
+  last array entry). Through v11 it held the first entry — the root frame for
+  native-only stacks, despite the column name.
+- `systing analyze flamegraph` (and the MCP flamegraph tool) previously
+  reversed the array on output under the leaf-to-root assumption, emitting
+  inverted folded stacks for native frames. It now emits storage order, which
+  is correct for all schema versions' native segments; python segments in
+  pre-v12 databases remain leaf-first within the blend.
--- a/src/analyze/flamegraph.rs
+++ b/src/analyze/flamegraph.rs
@@ -302,8 +302,9 @@ fn build_flamegraph_query(
 
 /// Format a chr(31)-separated string of frame ids into folded stack format.
 ///
-/// Frame ids are stored leaf-to-root in the database, so they are reversed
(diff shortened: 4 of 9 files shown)


josefbacik:
# rust-bpf-code-reviewer — pre-commit review of this PR's diff (stack-order-normalize vs main 8220668)

over the full diff (structure / Rust quality / BPF quality / schema & versioning /
integration), by the authoring seat with the empirical checks listed under
"What was verified" — disclosed per project review policy.*

### Summary

The change fixes a real, user-visible correctness bug with the minimal correct
mechanism: normalize segment direction once at `Stack::new` (the single
construction point both recorders share), and remove the two read-side
compensations that had encoded the wrong assumption (`leaf_name` from
`.first()`, `format_folded_stack`'s blanket `.rev()`). All downstream paths —
disk spill, symbolization, parquet, DuckDB import/merge, perfetto export —
inherit the invariant without modification, and the ones that could have been
double-affected were each checked (see Integration notes). Versioning gates are
satisfied.

### Critical Issues

None found. The three schema gates all pass:

1. `SCHEMA_CHANGES.md` — v12 entry present, documents the semantics change,
   the legacy mixed-order caveat, and import behavior.
2. `SCHEMA_VERSION` incremented 11 → 12 in `src/duckdb.rs`.
3. `Cargo.toml` **minor** bump 1.10.5 → 1.11.0 (schema change ⇒ minor, not
   patch) — `Cargo.lock` consistent.

### Improvements

1. **(applied during review)** `SCHEMA_CHANGES.md`: added the sentence covering
   older traces imported into a v12 database — they keep their written order;
   `_traces.systing_version` is the per-trace discriminator. Without it, the
   entry implied database schema version alone determines array order, which is
   false after cross-version import. A second pass found the discriminator
   itself needed a caveat: `parquet_to_duckdb` stamps the **converting**
   binary's `CARGO_PKG_VERSION` (nothing in the parquet directory records the
   recorder's version), so the entry now states the same-run condition and
   marks embedding the recorder's version in the parquet directory as planned
   follow-up.

### Suggestions

1. `reverse_py_stack` uses `.cloned()` — correct as written (`PyAddr` is
   `Clone` but not `Copy`, per `pystacks/stack_walker.rs`); noted only to
   record that `.copied()` was considered and does not apply.
2. Local clippy (1.95-nightly) is older than the CI runner's stable, which
   recently promoted new lints (see the `iter_over_map_values` fix on the
   sched-sharding PR). Local run is clean; the new-lint classes don't appear in
   this diff's added lines, but CI remains the authority.

### Positive Notes

- Fixing direction at construction rather than at read sites is the right
  architecture: one invariant, one comment stating it (on `Stack`), all
  consumers simplified. The removed `.rev()` + removed `.first()` make the
  read paths *shorter*, not more conditional.
- The v12 entry's honesty about what CANNOT be fixed (pre-v12 python segments
  are not re-orderable without frame classification) is exactly what a reader
  migrating old traces needs.
- Tests assert the invariant at both ends: construction
  (`test_stack_new_reverses_py_stack`) and emission
  (`test_emit_stack_record_leaf_name_is_last_frame`), plus the folded-output
  contract (`test_format_folded_stack_storage_order`).

### What was verified (not just read)

- `cargo test --lib`: **379 passed / 0 failed**, including the three
  new/changed tests and the pre-existing spill-roundtrip, schema-version, and
  legacy-migration tests.
- `cargo clippy --no-default-features -- -D warnings` (CI-exact invocation):
  clean. `cargo fmt --all -- --check`: clean.
- Perfetto pass-through claim checked against the vendored proto
  (`perfetto_protos-0.48.1`, `profile_common.proto`): `Callstack.frame_ids` is
  "Bottom frame first" — root-first — so emitting storage order is correct,
  and the python segment in .pb exports is *fixed* by this change (it was
  inverted before).
- Consumer sweep for order-sensitive reads of `frame_ids` / `frame_names` /
  `leaf_name` / `py_stack`: spill serializer+deserializer (order-preserving,
  no re-reverse), symbolization (order-preserving map), DuckDB import/merge
  SQL (`ORDER BY u.idx` preserves array order), memory recorder (shares
  `Stack::new`), MCP flamegraph tool (routes through the fixed
  `format_folded_stack`).

### Known adjacent issue — deliberately NOT in this PR

`src/bin/systing_util.rs::get_or_create_callsite` (the **perfetto .pb import**
path) builds its callsite tree assuming leaf-first `Callstack.frame_ids`
input; spec-compliant .pbs (including systing's own exports) are root-first,
so imported chains come out inverted, and its `depth` numbering is
root-highest under either assumption. That path reads .pb files, not parquet
or DuckDB, so nothing in this PR touches or changes its behavior. Kept out
deliberately — one behavioral change per PR; to be reported and fixed
separately.

---
*Currency note: the five-pass review above was performed against the diff at its original base (main 8220668). The branch has since been rebased twice (onto 76a073b, then 0646db6) — both auto-merges were conflict-free with no content changes to this diff, and the full validation suite (fmt, clippy, 380 lib tests, workspace doctests, and an 8-target VM integration gate including live gvisor and gopclntab capture) was re-run green on the pushed head cf8bb77.*

PR #146, opened 2026-07-10
Add a lightweight profile export format (.systing)

josefbacik:
## What

Adds a lightweight profile export format, `.systing`: a line-oriented JSON
serialization of the sampling-profile view of a recording — a header line
with sampling provenance (`sample_event`/`sample_period`), then interned
frames, interned leaf-first stacks, and per-thread sample tallies per stack
event type. Spec in `docs/PROFILE_EXPORT_FORMAT.md` (version 1, with
forward-compat rules).

Two producers share one writer:

- `systing --output profile.systing -- <cmd>` (or `.systing.gz` for gzip)
  exports straight from the recording's parquet files.
- `systing-util convert trace.duckdb --output profile.systing` exports from
  an existing trace database (`--trace-id` selects among multiple traces; a
  parquet directory also works as input).

## Why

The DuckDB database is the right home for the full trace, but reading it
requires the DuckDB engine — which walls the profile view off from tools
that can't take that dependency (pure-JS/Python consumers, Markdown
reporters, LLM pipelines). The export carries exactly the stacks/tallies/
provenance those tools need, streams line-by-line, and gzips well.

## Details worth review attention

- **Stack order:** `stack.frame_ids` is stored root-first (index 0 =
  outermost frame — verified empirically against recordings; note the
  in-tree comments disagree with each other on this), except that
  recordings from versions predating the stack-order normalization stored
  python segments leaf-first within the same array. The exporter enforces
  the format's `"stack_order":"leaf_first"` guarantee rather than passing
  storage through: legacy python runs are un-inverted before the uniform
  reversal, decided by structural evidence (root-side markers' position in
  the leading python run, per-trace majority) with version keying only as
  fallback — `_traces.systing_version` names the converting binary, not the
  recorder, and parquet directories carry no recording version at all.
  Validated against a real pystacks recording: every blended stack exports
  truly leaf-first through both producers, which agree tally-for-tally.
- **Referenced-rows filtering:** only stacks reachable from `stack_sample`
  export — the stack table is shared with the memory recorder
  (`memory_map`/`memory_fault`/`memory_alloc`), whose stacks aren't part of
  the sampling profile.
- **Tallies are pre-aggregated** per `(utid, stack_id, event_type)`; sleep
  events are occurrence counts, not durations (documented in the spec).
- No schema change (patch version bump per the versioning policy).

## Testing

- Unit tests: writer output shape, root-first→leaf-first reversal pinned
  against a hand-built database, referenced-only filtering, multi-trace
  selection errors, empty-recording error, gzip output.
- `trace_validation` integration suite gains a cross-producer check: the
  parquet-path and duckdb-path exports of the same recording must agree on
  header semantics and frame-name-resolved tallies (define-before-use and
  event-type validity asserted along the way).
- Full integration gate run green in a BPF-enabled VM before pushing;
  additionally validated a real recording's export against an independent
  SQL aggregation of the same database (tallies matched exactly).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.11.2"
+version = "1.11.4"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.11.2"
+version = "1.11.4"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/README.md
+++ b/README.md
@@ -25,7 +25,11 @@ cargo install --locked --git https://github.com/josefbacik/systing.git
 ```
 
 This will generate a `trace.pb` file which can be uploaded to a
-[Perfetto](https://perfetto.dev/) instance for further analysis.
+[Perfetto](https://perfetto.dev/) instance for further analysis. Other
+`--output` extensions select other formats: `.duckdb` writes a queryable
+[DuckDB](https://duckdb.org/) trace database, and `.systing` (or
+`.systing.gz`) writes a lightweight profile export any tool can parse without
+DuckDB — see [docs/PROFILE_EXPORT_FORMAT.md](docs/PROFILE_EXPORT_FORMAT.md).
 
 ## Development Setup
 
--- a/docs/PROFILE_EXPORT_FORMAT.md
+++ b/docs/PROFILE_EXPORT_FORMAT.md
@@ -1,0 +1,139 @@
+# systing profile export format (`.systing`)
+
+A lightweight, line-oriented profile interchange format for systing
+recordings. It carries the *sampling profile* view of a trace — interned
+stacks, per-stack sample tallies, and the process/thread metadata needed to
+attribute them — in a form any tool can parse without DuckDB: one JSON value
+per line, first line identifying the format.
+
+This is a **profile summary**, not the full trace. Timestamps, scheduler
+slices, counters, network/memory events, and everything else systing records
+stay in the DuckDB database (`--output trace.duckdb`); use the export when a
+downstream tool only needs "which stacks, how often, in which threads"
+(flamegraphs, Markdown reports, LLM pipelines).
+
+Produced by:
+
+```sh
+# directly from a recording
+systing --duration 30 --output profile.systing -- ./my-workload
+
+# from an existing DuckDB trace database
+systing-util convert trace.duckdb --output profile.systing
+```
+
+The file is plain UTF-8 text. Producers may gzip it (`profile.systing.gz`);
+consumers are encouraged to sniff the two-byte gzip magic rather than trust
+file names.
+
+## Structure
+
+Each line is one JSON value. Line 1 is the **header object**; every
+subsequent line is a **record array** whose first element is a one-letter
+type tag. Parsers must ignore record types they do not recognize (future
+minor additions), and must reject the file if the header is missing or its
+`systing_profile_export` version is greater than what they support.
+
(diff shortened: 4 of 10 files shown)


josefbacik:

*Mechanics: the repo's `rust-bpf-code-reviewer` agent definition (blob `b75d1052`) was executed by the authoring agent over the full branch diff before commit, per this repo's review flow. Issues found during the review were fixed in the branch before it was pushed; they're listed below with their resolutions so the review trail is visible.*

### Summary

Adds a self-contained profile-export module (`src/profile_export.rs`) with a spec doc, wires it into the record path and `systing-util convert`, and extends the `trace_validation` integration suite with a cross-producer consistency check. No BPF code is touched; no schema change. The module follows the codebase's established patterns (anyhow contexts, duckdb-rs usage mirroring `analyze/`, arrow batch reading mirroring `parquet_to_perfetto.rs`).

### Critical Issues

None found. The two versioning gates were checked explicitly:

- **Schema:** `create_schema()` and `src/trace/schema.rs` are untouched — no `SCHEMA_VERSION` bump or `SCHEMA_CHANGES.md` entry required.
- **Version:** patch bump `1.10.5 → 1.10.6` per the non-schema-change policy. Note for the maintainer: other in-flight branches also claim version numbers; whoever lands second has a one-line rebase.

### Improvements (found in review, fixed in the branch)

1. **Gzip trailer errors were swallowed.** The writer originally held a `Box<dyn Write>`; a dropped `GzEncoder` writes the gzip trailer but discards any I/O error, silently truncating `.systing.gz` output on e.g. a full disk. Fixed with an `ExportSink` enum whose `finish()` ends the gzip stream explicitly and propagates errors.
2. **Parquet tables were fully materialized.** The reader collected every `RecordBatch` of each table into a `Vec` before processing — the whole `stack_sample` table in memory for large recordings. Fixed by streaming batches through a visitor (`for_each_batch`), so no table is ever held whole; peak state is the tally/interning maps, which are proportional to distinct stacks, not samples.

- The DuckDB producer scans `stack_sample` in several `IN (SELECT DISTINCT …)` subqueries (threads, frames, stacks, tallies). A CTE or temp table would scan once; at observed sizes (a 5 MB trace exports in under 2 s) this isn't worth the complexity yet.
- `run_convert`'s new error message ("director(y/ies)") reads awkwardly; cosmetic.
- Pre-existing asymmetry surfaced by testing, not introduced here: the record path derives `trace_id` from the raw output-dir name while `systing-util` sanitizes hyphens to underscores (`generate_trace_id`). Harmless, but the two entry points can produce different `trace_id`s for the same directory name.

### Second pass (legacy python stack-order handling, added after the first review)

The addition that normalizes pre-normalization python segments was reviewed
under the same criteria:

- **Correctness of the decision procedure:** structural evidence (root-side
  markers' position within the leading python run, per-trace majority) is
  primary; version keying is fallback only, because `_traces.systing_version`
  names the converting binary — a reconverted parquet directory would
  otherwise be version-keyed into silently skipping (or wrongly applying)
  the fixup. Tests pin all four quadrants: structure overriding a lying
  newer version, structure confirming normalized data under an older
  version, and both version-fallback directions when structure abstains.
- **Validated on real data:** a python-blended recording (pystacks,
  CPython 3.12) exports with every blended stack truly leaf-first — 341/341
  stacks pass leaf-before-caller ordering checks including the
  `<module>`/trampoline root markers — and the two producers agree
  tally-for-tally on the blended recording.
- One Rust note: buffering the stack table (needed for the per-trace vote)
  replaces the previous fully-streaming duckdb path; stacks are the small
  interned table (samples stay aggregated in SQL), so the memory profile
  remains proportional to distinct stacks.

### Positive Notes

- The empirical stack-order note in both producers (root-first storage, reversed to the spec's leaf-first, with `stack_order` stated in the header) resolves a real ambiguity in the tree rather than propagating it.
- Referenced-rows filtering (only stacks reachable from `stack_sample`) shows awareness that the stack table is shared with the memory recorder — easy to get wrong silently.
- The integration check compares producers on frame-name-resolved tallies rather than raw ids, which is the right equality (ids are producer-internal).
- Unit tests pin the reversal against a hand-built database and cover the unhappy paths (multi-trace without `--trace-id`, unknown trace, empty recording, gzip round-trip).

PR #147, opened 2026-07-10
Fix perfetto import to read Callstack.frame_ids root-first

josefbacik:
## Summary

`systing-util`'s perfetto-trace importer read `Callstack.frame_ids` as
leaf-first and built its callsite tree from the reversed array. Perfetto's
convention is root-first — and systing's own exporter emits storage order,
uniformly root-to-leaf since schema v12 (#145) — so importing any
spec-compliant trace, including systing's own exports, produced an inverted
tree: the real leaf became a parentless root, the callsite id recorded on
perf samples pointed at the real root, and `depth` counted leaf=0.

## Fix

`get_or_create_callsite` now reads `frame_ids` in order (root first), chains
parents root→leaf, and writes `depth` root=0 increasing toward the leaf,
matching perfetto trace processor's `stack_profile_callsite` semantics. The
full-stack dedup fast path keys the input slice directly instead of
allocating a reversed copy.

## Compatibility

Table shapes are unchanged — no `SCHEMA_VERSION` bump; `SCHEMA_CHANGES.md`
documents the semantics change. `callsites` tables written by older binaries
from spec-compliant inputs carry the inverted parent/depth shape and cannot
be distinguished in-band; re-import with a fixed binary where tree direction
matters. Patch version bump to 1.11.2 (1.11.1 is taken by the in-flight export-tooling PR; versions need not be dense).

## Testing

- New regression test pinning the convention end-to-end
  (`test_callsite_frame_ids_are_root_first_per_perfetto_spec`): a
  main→foo→bar sample must root at main, return bar as the leaf, with
  depths 0/1/2.
- Existing callsite tests updated from the leaf-first expectations they
  encoded; shared-prefix dedup now tested on the root-side prefix (which is
  what actually shares under either convention's tree).
- `cargo test --bin systing-util`: 27/27. fmt and clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.11.0"
+version = "1.11.2"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.11.0"
+version = "1.11.2"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -255,3 +255,19 @@ tools can distinguish uniformly-ordered databases.
   inverted folded stacks for native frames. It now emits storage order, which
   is correct for all schema versions' native segments; python segments in
   pre-v12 databases remain leaf-first within the blend.
+
+## Perfetto import callsite order (no schema version change)
+
+`systing-util`'s perfetto-trace importer previously read `Callstack.frame_ids`
+as leaf-first and built the callsite tree from the reversed array. Perfetto's
+convention is root-first (and systing's own exporter emits storage order,
+uniformly root-first since v12), so importing any spec-compliant trace —
+including systing's own exports — produced an inverted tree: the real leaf
+became the parentless root, the returned "leaf" id pointed at the real root,
+and `depth` counted leaf=0. The importer now reads frame_ids root-first and
+writes `depth` root=0 increasing toward the leaf, matching perfetto trace
+processor's `stack_profile_callsite`. Table shapes are unchanged (no
+SCHEMA_VERSION bump); `callsites` tables written by older systing-util
+binaries from spec-compliant inputs carry the inverted parent/depth shape and
+cannot be distinguished in-band — re-import with a fixed binary if the tree
+direction matters.
--- a/src/bin/systing_util.rs
+++ b/src/bin/systing_util.rs
@@ -1597,17 +1597,19 @@ impl TraceExtractor {
         Ok(())
     }
 
-    /// Convert flat frame_ids to parent-child callsite tree. Returns leaf callsite ID.
+    /// Convert flat frame_ids to a parent-child callsite tree. frame_ids
+    /// follow the perfetto Callstack convention: ROOT first (frame_ids[0]),
+    /// leaf last — the order perfetto's own tooling produces and systing's
+    /// exporter emits (storage-order pass-through, uniform since schema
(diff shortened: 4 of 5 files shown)


josefbacik:
# rust-bpf-code-reviewer — pre-push review of this PR's diff (pb-import-order vs main 17d725a)

over the full diff (structure / Rust quality / BPF quality / schema & versioning /
integration), by the authoring session prior to push, per this repo's review
convention.*

### Summary

Single-function semantic fix with proportionate test coverage: the perfetto
importer's `get_or_create_callsite` now reads `Callstack.frame_ids` root-first
per the perfetto convention (and per what systing's own exporter emits), with
depth semantics aligned to trace processor's `stack_profile_callsite`. Tests
that previously pinned the wrong convention are rewritten, a spec-anchoring
regression test and a wire-level round-trip test are added, and two stale
segment-position comments in the recorder are corrected.

### Critical Issues

None found.

### Improvements

None required. The full-stack dedup fast path now keys the input slice
directly (`callsite_map.get(frame_ids)` via `Borrow<[u64]>`), removing a
per-call `Vec` allocation that existed only to reverse the key.

### Suggestions

None material. The `depth: i as i32` cast is bounded by real stack depths;
same shape as the pre-existing code.

### Positive Notes

- `test_perfetto_wire_round_trip_preserves_tree_direction` exercises the full
  packet path (interned data → sample → tree → symbols), locking the
  export/import contract at the .pb boundary rather than only inside the tree
  builder, and asserts the sample's `callsite_id` is the true leaf — the
  user-visible face of the original bug.
- Versioning gate satisfied: non-schema change carries a patch bump
  (1.11.2; 1.11.1 was taken by an in-flight PR — versions need not be dense),
  `SCHEMA_VERSION` correctly untouched (table shapes unchanged), and
  `SCHEMA_CHANGES.md` documents the semantics change including the
  in-band-indistinguishability caveat for tables written by older binaries.
- The rewritten shared-prefix test now tests root-side sharing, which is what
  actually dedups under the corrected convention.

josefbacik:
**Full integration suite on `1d0bba63`: 8/9 targets PASS; `trace_validation` 30/31 — the one failure is `test_pystacks_fork_exec`, analyzed below with rerun evidence.**

| target | result |
|---|---|
| trace_validation | **30/31 — see below** (2124s) |
| memory_recorder | PASS 5/5 (590s) |
| stream_unix | PASS (127s) |
| analyze_commands | PASS (869s) |
| perf_counter | PASS |
| network_throughput | PASS (64s) |
| gvisor_record | PASS — live gVisor capture (767s) |
| gopclntab_record | PASS — live Go-symbolization capture (243s) |
| missed_events | PASS (51s) |

**The failure, precisely:** `test_pystacks_fork_exec` panicked in the *recorder's* symbolization phase — `src/pystacks/linetable.rs:207: attempt to shift left with overflow` — not in anything this PR touches.

**Why this is a pre-existing latent bug, not a regression from this diff:**
1. **Reachability:** this PR changes `src/bin/systing_util.rs` (perfetto *import*), two comments in `src/stack_recorder.rs`, and the version. `src/pystacks/linetable.rs` is untouched by this PR (and unchanged since the pure-Rust pystacks rewrite). The panicking code cannot be reached from this diff.
2. **History:** the same suite ran 31/31 three times earlier today, including on this PR's parent tree at 19:46Z. The failing test is timing-dependent by construction (it forks and execs Python children mid-trace).
3. **Mechanism (from source):** `read_varint`'s continuation loop has no shift cap. A fork+exec'd child's code-object memory can be torn/replaced mid-read, presenting arbitrary continuation bytes; six or more drive the shift past 31 and a debug build panics. Release builds wrap silently — corrupt line numbers — which is arguably worse.
4. **Reruns on this exact tree (`1d0bba63`):** the identical test re-run twice after the failure — **PASS and PASS** (67s and 67s). Same tree, same VM image, same Python 3.13.

**Merge implication, stated plainly:** if the rerun evidence reads as it does above, the failure is a nondeterministic latent bug that any of today's runs could have drawn, and the merge decision on this PR is unaffected by it. Nothing is pre-decided — your call.

**The latent bug itself is real and worth fixing:** an unlucky fork/exec during any whole-system pystacks capture can panic the recorder mid-trace (tonight it killed the streaming parquet writer mid-run). A bounds-check fix with a pathological-input unit test follows as its own small PR.

PR #148, opened 2026-07-11
Bound varint decoding in pystacks location tables

josefbacik:
## Problem

`read_varint` in `src/pystacks/linetable.rs` decodes CPython 3.11+ location-table varints with an uncapped continuation loop. The input bytes come from traced process memory, which can be torn or replaced mid-read (fork/exec churn), so the parser must tolerate arbitrary bytes. A run of six or more fake continuation bytes drives the shift amount to 36, and `u32 << 36` panics in debug builds ("attempt to shift left with overflow") — observed killing an entire capture mid-trace when `test_pystacks_fork_exec` drew exactly that input from a fork+exec'd child's replaced code objects. Release builds mask the shift silently and corrupt line numbers instead.

The line-number accumulator one level up (`parse_location_table`) has the same exposure: two torn entries carrying saturated signed varints overflow the i32 sum — debug panic, release wrap.

## Fix

- `read_varint`: accumulate in u64 so all six well-formed groups (up to shift 30, reaching bit 35) decode losslessly; stop at a sixth continuation byte — the table is not well-formed past that point and no further decoding recovers it; saturate values past `u32::MAX` instead of wrapping. `read_signed_varint` inherits the bound.
- `parse_location_table`: `line_number` accumulates with `saturating_add`.
- Audited the rest of the parser for the same class: the `addr` accumulator and the whole 3.10 lnotab path are bounded by `MAX_LINE_TABLE_SIZE` (1 MiB) and need no cap; no other unbounded shift decoders exist under `src/pystacks/`.

## Validation

- New tests: the six-continuation-byte overflow reproducer (the exact field panic), six-group saturation, endless-continuation termination + position semantics, multibyte decode, and the torn-delta line-accumulator reproducer. 12/12 linetable unit tests pass.
- Negative controls: both reproducers were run against the unfixed code first and fail there as required — the varint reproducer panics at the shift site, the accumulator reproducer at the add site. (A first draft of the varint reproducer used five continuation bytes and passed against the old code — the shift only reaches 36 at six bytes; the negative-control step caught the weak test.)
- `cargo fmt` clean; `cargo clippy --lib` introduces no new warnings.
- Full integration suite on this exact commit (`1c48203`):

| target | result |
|---|---|
| trace_validation | PASS 31/31 (2179s) — includes `test_pystacks_fork_exec`, the test that drew the original panic |
| memory_recorder | PASS 5/5 (613s) |
| stream_unix | PASS (153s) |
| analyze_commands | PASS (901s) |
| perf_counter | PASS (36s) |
| network_throughput | PASS (100s) |
| gvisor_record | PASS — live gVisor capture (813s) |
| gopclntab_record | PASS — live Go-symbolization capture (264s) |
| missed_events | PASS (88s) |

9/9 targets green.

Version: ships 1.11.5 (non-schema patch bump; originally claimed 1.11.3, reconciled past 1.11.4 after #146 and #147 merged ahead of this PR).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.11.0"
+version = "1.11.3"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.11.0"
+version = "1.11.3"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/src/pystacks/linetable.rs
+++ b/src/pystacks/linetable.rs
@@ -176,7 +176,7 @@ impl PyLineTable {
                 0
             };
 
-            line_number += line_delta;
+            line_number = line_number.saturating_add(line_delta);
             let end_addr = addr + delta * 2;
 
             if !callback(addr as usize, end_addr as usize, line_number) {
@@ -188,6 +188,16 @@ impl PyLineTable {
 }
 
 /// Read a varint from the location table data.
+///
+/// The bytes come from traced process memory, which can be torn or
+/// replaced mid-read (fork/exec churn), so this must tolerate arbitrary
+/// input. A well-formed CPython location-table varint fits u32 in at
+/// most six 6-bit groups; the accumulator is u64 so the sixth group
+/// (shift 30, reaching bit 35) cannot lose bits, and a sixth
+/// continuation byte stops the read — the shift would reach 36, the
+/// table is not well-formed, and no amount of further decoding
+/// recovers it. (`x << 36` on u32 is the debug-build panic this
+/// replaces; release builds shifted into oblivion silently.)
 fn read_varint(data: &[u8], pos: &mut usize) -> u32 {
     let len = data.len();
     if *pos >= len {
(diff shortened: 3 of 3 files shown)


josefbacik:
**Code review — `rust-bpf-code-reviewer`** (review executed against the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition, full diff of this PR, blob `b75d1052`; mechanics: definition followed pass-by-pass in-session by the submitting maintainer's tooling, disclosed per repo practice)

### Summary

Small, focused hardening of the CPython location-table parser against torn traced-process memory. The fix is minimal and correct: the varint decoder gets a u64 accumulator, a six-group cap, and saturation; the one sibling accumulator with the same exposure is saturated too. Tests validate both failure modes against the unfixed code (negative controls) before validating the fix.

### Critical Issues

None outstanding. One found during review and fixed in the same commit: `parse_location_table`'s `line_number += line_delta` could overflow i32 on two torn entries carrying saturated signed varints (±i32::MAX each) — the same debug-panic/release-wrap class as the shift overflow, reachable through the same torn-memory input, and not covered by the varint cap alone. Fixed with `saturating_add`; reproducer test added and verified to panic against the unfixed accumulator.

### Versioning & Schema (gate checks)

- No schema change (parser-internal only) → `SCHEMA_VERSION` untouched: **correct**.
- Non-schema change → patch bump required: `Cargo.toml` 1.11.0 → 1.11.3: **present** (1.11.1/1.11.2 are claimed by open PRs).
- No recorder/table changes → utid/thread-attribution criteria: **N/A**.

### Improvements

None required. Bounds analysis for the untouched accumulators (`addr`, 3.10 lnotab path) is documented in the diff and checks out against `MAX_LINE_TABLE_SIZE` = 1 MiB: max `addr` accumulation is 16 bytes of address space per table byte ≈ 2^24, far under i32::MAX; 3.10 line deltas are i8-bounded at ≤127 per 2-byte entry ≈ 2^26 worst case.

### Suggestions

None. `val.min(u32::MAX as u64) as u32` reads clearly; the doc comment explains the why (input is untrusted process memory) without restating the code and references functions by name rather than line numbers.

### Positive Notes

- Negative-control discipline: both reproducers were first run against the unfixed code and demonstrated the exact panics ("attempt to shift left with overflow" at the shift site; "attempt to add with overflow" at the accumulator). The first varint reproducer draft (five continuation bytes) passed against the old code — the control caught the weak test and forced the correct six-byte reproducer.
- `test_varint_endless_continuation_run_terminates` pins the pos-advance contract (stop after six groups, don't consume the buffer), which protects the outer parse loop's termination argument.
- The fix keeps failure semantics defined in release builds (saturate) rather than only silencing the debug panic.

PR #149, opened 2026-07-13
Make --only-recorder network state-only

josefbacik:
`--only-recorder network` was documented as recording "only network traffic", and the code comment in `enable_recorder` promised state-only capture was available that way — but the coupling made it unreachable: naming `network` on any CLI path also enabled `network-packets`, so every invocation got the full per-packet instrumentation. The library mode works (`Config { network: true, network_packets: false }` — `test_e2e_network_suite` exercises exactly that and relies on only the state probes being loaded); the CLI just couldn't express it.

This splits the selection semantics instead of the recorders:

| invocation | before | after |
|---|---|---|
| `--add-recorder network` | state + packets | state + packets (unchanged) |
| `--only-recorder network` | state + packets | **state only** |
| `--only-recorder network-packets` | state + packets | state + packets (`network` is a required base, still pulled in) |
| disabling `network` | disables both | disables both (unchanged) |

Motivation: per-packet event volume is traffic-bound — measured ~90K events/s on a busy node pushing ~1.5GB/s — fine for a bounded investigation, prohibitive for continuous or fleet-wide profiling. Connection-state tracking alone is ~2 orders of magnitude cheaper on the same node (state transitions only) and is the shape an always-on profile wants.

**Semantics call for you — two reasonable shapes, both small:**
- **A (this PR): companions are `--add-recorder` behavior.** `--add-recorder network` keeps enabling packets (the usual investigation shape); `--only-recorder` means exactly what you name. Consequence: zero behavior change for existing add-mode invocations; state-only becomes reachable.
- **B: never couple.** `network` never implies `network-packets`; packet users always name it explicitly. Consequence: strictest one-name-one-recorder rule, but `--add-recorder network` alone stops collecting packet data — a behavior change for existing usage.

I lean **A** — it keeps the documented investigation ergonomics intact and only changes the path that was broken. Flipping to B is a three-line change to the same structure if you'd rather.

Note for API consumers driving systing programmatically: a request that passed only `network` and relied on implicitly getting packet tracing should name `network-packets` too (order doesn't matter).

Covered by CLI-layer tests through the real parse path: state-only, dependency pull-in, order independence, add-mode companion, defaults-off, and the analogous `memory`/`memory-alloc` base dependency. README and USAGE.adoc updated to describe the two tiers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.11.5"
+version = "1.11.6"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.11.5"
+version = "1.11.6"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/README.md
+++ b/README.md
@@ -119,7 +119,8 @@ Use `--only-recorder` to disable all recorders and enable only the ones you spec
 # Only record syscalls (disable everything else)
 sudo ./target/debug/systing --only-recorder syscalls --duration 60
 
-# Only record network traffic (disable everything else)
+# Only record network connection state (disable everything else,
+# including the packet-level probes — see Network Traffic Recording)
 sudo ./target/debug/systing --only-recorder network --duration 60
 
 # Only record syscalls and cpu-stacks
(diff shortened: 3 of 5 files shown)


josefbacik:
**rust-bpf-code-reviewer** — the repo agent definition (`.claude/agents/rust-bpf-code-reviewer.md`) executed over the full diff at 8d8bcec before push; structured verdict below, mechanics as stated.

### Summary
Focused CLI-layer decoupling with correct semantics on every selection path; no BPF or schema surface touched. Tests exercise the real clap parse path rather than hand-built structs, which is the right level for a selection-semantics fix.

### Critical Issues
None.
- Schema gates: no schema change → `SCHEMA_VERSION` and `SCHEMA_CHANGES.md` correctly untouched; non-schema change → patch bump 1.11.5 → 1.11.6 present in `Cargo.toml` + `Cargo.lock`.
- Recorder-consistency gates (utid/thread attribution): no recorder or table changes — N/A.

### Improvements
None blocking. One distinction worth preserving as this grows: `network-packets → network` is a *dependency* (selection-independent, still pulled in under `--only-recorder`), while `network → network-packets` is a *companion* (only under `--add-recorder`). The code now expresses both correctly; keep them separate if further recorder relationships are added.

### Suggestions
None.

### Positive Notes
- `Selection` as a named enum keeps call sites readable (`Selection::Exact` vs an anonymous bool).
- The test matrix includes order-independence and the sibling `memory`/`memory-alloc` dependency, so the next change to `enable_recorder` inherits guardrails.

Verification: `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` exit 0; `cargo test` full suite 436 passed, 0 failed, 11 ignored (the ignored ones are the privilege-gated e2e legs; these are the pre-commit hook's own three checks).

PR #150, opened 2026-07-13
Bump blazesym to upstream 862c2cf5d424

josefbacik:
Picks up libbpf/blazesym#1610 — sizeless `.dynsym` symbols no longer absorb every address up to the next symbol, so stripped cgo Go binaries stop resolving to confidently wrong names (stray dynamic-symbol anchors claiming whole Go text ranges) and fall through honestly to the gopclntab resolver instead.

Details in the commit message. Verification:
- `cargo test --workspace --no-default-features`: 414 passed / 0 failed
- `cargo clippy --workspace --no-default-features -- -D warnings`: clean (after boxing `ExportSink::Gz`, which upstream's zlib backend switch pushed over the `large_enum_variant` threshold via feature unification)
- `cargo fmt --all -- --check`: clean

Upstream MSRV is now 1.88. Once upstream tags 0.2.6 on crates.io this can move back to a registry version with no code change.

PR #151, opened 2026-07-13
build.rs: use the target multiarch include dir when cross-compiling

josefbacik:
Cross-compiling systing (e.g. `cargo build --target aarch64-unknown-linux-gnu` on an x86_64 host) fails while compiling the pystacks BPF object:

```
/usr/include/x86_64-linux-gnu/gnu/stubs.h:7:11: fatal error: 'gnu/stubs-32.h' file not found
```

`detect_multiarch_include()` derives the multiarch triplet from the build HOST (`dpkg-architecture` / `cc -dumpmachine`), so a cross build passes the host's `-I/usr/include/x86_64-linux-gnu` alongside the TARGET's `-D__aarch64__`. Under `-target bpf`, clang defines neither `__x86_64__` nor `__i386__`, so glibc's x86 `stubs.h` takes its 32-bit branch and demands a header that doesn't exist on 64-bit-only installs. Native builds never see this: there the arch define matches the host headers.

**Fix**: when `CARGO_CFG_TARGET_ARCH` differs from the host arch, resolve the include dir from the target triplet instead — preferring `/usr/include/<triplet>` (Debian multiarch, `libc6-dev:<arch>`), then `/usr/<triplet>/include` (cross-toolchain sysroot) — and emit a cargo warning naming the missing package when neither exists, rather than silently offering host headers. The native path is unchanged.

**Verification**
- x86_64 native: `cargo check`, `cargo clippy --all-targets`, `cargo fmt --check` all clean (native path unchanged).
- x86_64 → aarch64 cross, clean `ubuntu:24.04` with `gcc-aarch64-linux-gnu` + `libc6-dev:arm64`, no extra env: `cargo check --target aarch64-unknown-linux-gnu` **fails at the base commit** with the stubs error above and **passes with this patch** — both BPF objects compile against the aarch64 headers.

Version: patch bump to 1.11.8 (1.11.6 and 1.11.7 are claimed by #149 and #150, currently in flight).

**Note for cross builders**: libbpf-cargo's skeleton generation also requires the `rustfmt` component (rustup's `minimal` profile omits it) — independent of this change, but you'll hit it on the same path.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -12,8 +12,53 @@ const SRC: [&str; 1] = ["src/bpf/systing_system.bpf.c"];
 /// under `/usr/include/<triplet>/` (e.g. `/usr/include/x86_64-linux-gnu/`).
 /// On Fedora/RHEL they're directly in `/usr/include/asm/`.
 ///
-/// Returns a `-I/usr/include/<triplet>` string if needed, or None.
+/// When cross-compiling, the triplet must be the TARGET's, not the build
+/// host's: the BPF objects are compiled with the target's `-D__<arch>__`
+/// define, and host libc headers plus a foreign arch define fall over in
+/// arch-conditional glibc internals. Concretely, compiling for aarch64 on
+/// an x86_64 host, x86's `gnu/stubs.h` hits `#include <gnu/stubs-32.h>`
+/// (absent on a 64-bit-only install) because under `-target bpf` neither
+/// `__x86_64__` nor `__i386__` is defined.
+///
+/// Returns a `-I<dir>` string if needed, or None.
 fn detect_multiarch_include() -> Option<String> {
+    // Cross build: prefer the target's header locations. On Debian/Ubuntu
+    // these are /usr/include/<triplet> (multiarch, from libc6-dev:<arch>)
+    // or /usr/<triplet>/include (cross-toolchain sysroot, shipped alongside
+    // gcc-<triplet>). If neither exists, warn and return None rather than
+    // silently offering the HOST's headers — setups with a real sysroot
+    // pass their own include flags.
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").expect("CARGO_CFG_TARGET_ARCH not set");
+    if target_arch != env::consts::ARCH {
+        // get_arch_config() already restricts supported targets to these two.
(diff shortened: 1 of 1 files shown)


josefbacik:

**Mechanics**: this review executed the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this PR (build.rs, Cargo.toml, Cargo.lock), by the authoring session prior to push. Verification receipts are in the PR body.

### Summary
Targeted fix for a real cross-compilation failure, structured as a guard clause ahead of the existing host-detection logic — the native path is byte-identical, keeping the regression surface minimal. The failure mechanism is documented where future readers need it (the doc comment), and the fallback fails toward diagnosability (warn + `None`) instead of silently wrong headers.

### Critical Issues
None.
- Versioning gate: non-schema change carries the required patch bump (1.11.5 → 1.11.8; 1.11.6/1.11.7 are claimed by #149/#150 currently in flight).
- No schema, recorder, or BPF program changes; no `unsafe`.

### Improvements
None required.

### Suggestions
- `detect_multiarch_include()` now reads `CARGO_CFG_TARGET_ARCH`, which `get_arch_config()` also reads; passing the arch in as a parameter would deduplicate, at the cost of the function's self-contained signature. Not worth the churn for a build script — noted for a future refactor if this function grows.
- The `_ => return None` match arm is unreachable in practice (`get_arch_config()` panics first for unsupported targets); the inline comment already says so, and defensive depth in a build script is cheap. Fine as is.

### Positive Notes
- The doc comment explains the *why* (host headers + foreign arch define under `-target bpf` → glibc's stubs.h wall) rather than restating the code.
- The cargo warning names the concrete missing package per target arch (`libc6-dev:arm64` / the cross toolchain) instead of a generic complaint — it converts the next person's failure into a one-command fix.
- Checking both `/usr/include/<triplet>` and `/usr/<triplet>/include` covers the two ways Debian-family systems actually provide target headers (multiarch package vs cross-toolchain sysroot).

**Gates**: `cargo fmt --check` clean · `cargo clippy --all-targets` clean · native `cargo check` clean (unchanged path) · cross `cargo check --target aarch64-unknown-linux-gnu` fails at base / passes at HEAD in a clean ubuntu:24.04 cross environment.

PR #152, opened 2026-07-14
Add a network-syscalls recorder tier

josefbacik:
Between `network` (connection state only — no byte accounting) and `network-packets` (per-packet probes whose event volume is bounded by traffic) there was no middle: to get bytes, stalls or drops per connection you had to take the full per-packet firehose. This adds `network-syscalls`: per-syscall send/receive accounting plus the low-frequency diagnostics (retransmit timer, zero-window probes, sndbuf stalls, packet drops), without the per-packet and per-poll probes.

Measured on a busy node (~1.5GB/s, the same capture behind #149's motivation): full packet tier ≈ 88K events/s, the syscall+diagnostics subset ≈ 29K events/s, state alone ≈ 0.3K events/s — this tier delivers per-connection throughput and health at roughly a third of the packet tier's event rate, and the excluded probes (transmit/receive path, qdisc, epoll) are exactly the ones that fire per packet or per poll.

The implementation is declarative: a new `RecorderInfo` listing the 13-program subset (the existing autoload gating does the rest — no BPF changes), a `network_syscalls` config bool, and the CLI arm. One non-obvious inclusion, documented on the program list: `skb_recv_udp_exit` emits no events but fills the pending UDP recv entry with the peer address — without it `handle_recvmsg_exit` drops every UDP recv record, so leaving it out would silently blank UDP receive accounting in this tier.

Selection semantics follow #149: `network-syscalls` pulls in the base `network` recorder it requires (a dependency, under both `--add-recorder` and `--only-recorder`) and is never enabled implicitly — `--add-recorder network` still means state + packets, unchanged. Ringbuf families: syscall records land in `network_ringbufs`, the diagnostics are `packet_event`s, and `epoll_ringbufs` stays off in this tier.

Covered by CLI-layer tests through the real parse path (tier isolation, base pull-in, never-implied, tier union). README and USAGE.adoc updated; the enumerated recorder list in USAGE.adoc had drifted from the registry (several recorders missing) — I completed the network trio there and left the full refresh alone in case you'd rather batch it separately.

Version: 1.11.13 — renumbered above 1.11.10, which landed first via #153; .11 and .12 are claimed by #150 and #151 (open); whichever lands later reconciles, same as the recent ladder.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3438,7 +3438,7 @@ dependencies = [
 
 [[package]]
 name = "systing"
-version = "1.11.10"
+version = "1.11.13"
 dependencies = [
  "anyhow",
  "arrow 54.3.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ members = [".", "crates/gopclntab"]
 
 [package]
 name = "systing"
-version = "1.11.10"
+version = "1.11.13"
 edition = "2021"
 # Floor set by safe (no-unsafe) std::arch::x86_64::__cpuid in detect_hypervisor.
 rust-version = "1.89"
--- a/README.md
+++ b/README.md
@@ -137,14 +137,20 @@ The network recorder captures detailed network traffic information including:
 
 **Note:** Network recording is disabled by default to minimize overhead. Enable it explicitly when you need to analyze network performance.
 
-Network tracing is split across two recorders: `network` tracks TCP
-connection state (socket lifecycle and state transitions via
-`inet_sock_set_state`), and `network-packets` adds the per-packet and
-per-syscall probes listed below. `--add-recorder network` enables both,
-which is the usual shape for an investigation. `--only-recorder` enables
-exactly what you name, so `--only-recorder network` records connection
-state alone — useful when the packet-level event volume would be
-prohibitive, e.g. continuous or fleet-wide profiling.
+Network tracing is split across three recorders, ordered by event volume:
+`network` tracks TCP connection state (socket lifecycle and state
+transitions via `inet_sock_set_state`); `network-syscalls` adds per-syscall
+send/receive accounting plus the low-frequency diagnostics (retransmit
+timer, zero-window probes, sndbuf stalls, packet drops) — bytes, stalls and
+drops per connection at syscall-rate cost; and `network-packets` adds the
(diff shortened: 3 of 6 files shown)


josefbacik:
**rust-bpf-code-reviewer** — the repo agent definition (`.claude/agents/rust-bpf-code-reviewer.md`) executed over the full diff at 2afc515 before push; structured verdict below, mechanics as stated.

### Summary
A new recorder tier expressed entirely through the existing declarative registry — one program-list const, one `RecorderInfo`, one enabled-arm, CLI plumbing, tests, docs. No BPF changes; the autoload gating and family-derived ringbuf setup do all the work, which is the right architecture for this change.

### Critical Issues
None.
- Schema gates: no schema change → `SCHEMA_VERSION` and `SCHEMA_CHANGES.md` correctly untouched; non-schema change → patch bump 1.11.6 → 1.11.9 in `Cargo.toml` + `Cargo.lock` (ladder note in the PR body).
- Recorder-consistency gates (utid/thread attribution): no new tables or emitters — N/A.

### Improvements
None blocking. The dependency analysis behind the program list is the load-bearing part and it's documented where it matters: `skb_recv_udp_exit` is included as a map-fill helper (zero ringbuf reservations — verified) because `handle_recvmsg_exit` drops UDP recv records whose peer was never extracted; UDP send needs no analogous helper (the address is read from `msghdr` at entry); TCP is self-sufficient both directions.

### Suggestions
None.

### Positive Notes
- Consumer paths need no changes because they're event-driven and family-derived (`get_required_ringbuf_families` → `setup_ringbuffers`), so the new tier cannot drift from its consumers.
- The test matrix pins the semantics that matter: tier isolation (`--only-recorder network-syscalls` ⇒ no packets), base pull-in under both selection modes, never-implied (`--add-recorder network` does not enable the tier), and tier union.
- `epoll_ringbufs` correctly excluded — the tier's families were derived per-handler (send/recv exits → `reserve_network_event`; diagnostics → `reserve_packet_event`), not assumed.

Verification: `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` exit 0; `cargo test` full suite 440 passed, 0 failed, 8 ignored (the ignored ones are the privilege-gated e2e legs; the pre-commit hook re-ran all three checks green at commit time).

PR #153, opened 2026-07-14
pystacks: add real Python 3.14 offset support

josefbacik:
pystacks silently falls back to Python 3.13 struct offsets for any interpreter version above 3.13 (`for_version`: `m if m > 13 => py313()`). CPython 3.14 moved the load-bearing internals, so on 3.14 processes the walk dies at the first step — thread-state discovery reads `_PyRuntime` at the wrong offset — and no Python frames are ever produced:

```
PYRUNTIME_TLS_KEY_OFFSET          2164 -> 2340   (thread-state discovery)
PYRUNTIME_INTERPRETERS_HEAD       632  -> 808    (3.13+ GIL relocation path)
PY_INTERP_FRAME_LOCALSPLUS        72   -> 80
PY_INTERP_FRAME_OWNER             70   -> 74
PY_TUPLE_OBJECT_OB_ITEM           24   -> 32
PY_CODE_OBJECT_CO_CODE_ADAPTIVE   200  -> 208
PYINTERP_GIL_{LOCKED,LAST_HOLDER} +8 each
```

**Fix**: real 3.14 support via the repo's own generator — no guessing:

- `src/pystacks/bindings/v3_14_0.rs`: generated by `scripts/generate_python_bindings.py` against CPython `v3.14.0` (tag `ebf955d`). The generator compiled clean against the 3.14 headers, i.e. every field pystacks reads still exists under the same name — this is a constants-only change, no walker logic needed.
- `offsets.rs`: `py314()` (same shape as `py313()` — 3.14 keeps the no-`_PyCFrame` `current_frame`-on-`_ts` layout), `14 =>` arm, fallback now `>14 => py314()`.
- The `minor >= 13` GIL-relocation branch in discovery naturally covers 3.14; `py314()` populates the interpreter-state GIL offsets it consumes.
- Generator `VERSIONS` list gains `v3.14.0`; tests assert the generated ground-truth values (`tls_key 2340`, `interpreters_head 808`, frame/tuple offsets) and the future-fallback test now checks 3.15 → 3.14.

**Verification**: 401/401 lib tests (including the new `test_py314_offsets`), `cargo fmt --check` and `clippy --all-targets` clean; full 3.8–3.14 pystacks integration matrix green in a BPF-capable VM (virtme-ng on the repo's v6.12 vmtest kernel), including the new `test_pystacks_python314` walking a live 3.14.6 interpreter.

**Integration tests**: `scripts/setup-pystacks.sh` adds `3.14.6` to the interpreter set and `tests/trace_validation.rs` adds `PYTHON_314_VERSION` + `test_pystacks_python314`, so the per-version matrix exercises thread-state discovery and the frame walk against a real 3.14 binary — the leg that offset unit tests alone can't cover.

**Reviewer gate**: `.claude/agents/rust-bpf-code-reviewer.md` gains a "Python Version Support Review Criteria" section making version-adds self-enforcing: a diff that adds a new Python version without the matching `setup-pystacks.sh` entry and integration-matrix update is flagged as a Critical Issue — the exact gap that let the silent `>13 => py313()` fallback sit unnoticed.

Version: 1.11.10 (1.11.7–1.11.9 are claimed by #150–#152 currently in flight).

Note: like the existing `v3_8_0`–`v3_13_0` files, the constants were generated on x86_64; the fields involved are pointer/int members of LP64 structs, so the offsets hold on aarch64 as well — same pre-existing assumption as the sibling files, now stated.


josefbacik:

**Mechanics**: this review executed the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this PR (bindings/v3_14_0.rs, bindings/mod.rs, offsets.rs, generate_python_bindings.py, Cargo.toml, Cargo.lock), by the authoring session prior to push.

### Summary
Mechanical version-table extension following the file's established per-version pattern exactly: generated constants file, a `py314()` mirroring `py313()` (3.14 keeps the no-`_PyCFrame` thread-state shape), and the version match extended with the fallback moved up. No walker or BPF changes — the generator compiling clean against 3.14 headers is the evidence that none are needed.

### Critical Issues
None.
- Versioning gate: non-schema change carries the required patch bump (1.11.10; .7–.9 claimed by in-flight PRs, skip explained in the body).
- No schema, recorder, or BPF program changes; no `unsafe`.

### Improvements
None required.

### Suggestions
- `py314()` repeats `py313()` nearly verbatim — that is this file's idiom (every version function is spelled out against its own bindings module), and deduplicating would trade greppable per-version explicitness for indirection; correctly not attempted here.
- The previous `test_for_version_future_fallback` asserted the silent-fallback behavior this PR removes; it now asserts 3.15 → 3.14. Good that it was flipped rather than deleted — the fallback behavior itself remains covered.

### Positive Notes
- Constants come from the repo's own generator run against the real v3.14.0 tag rather than hand-derived values, and the tests pin the generated ground truth (`tls_key_offset 2340`, `interpreters_head 808`), so a future regeneration that drifts will fail loudly.
- The body documents which discovery-path branch (`minor >= 13` GIL relocation) covers 3.14 and why no new branch is needed — the review question a reader would ask is pre-answered.
- The x86_64-generation/LP64-portability assumption is stated instead of silently inherited from the sibling files.

**Gates**: 401/401 lib tests (incl. `test_py314_offsets`) · `cargo fmt --check` clean · `cargo clippy --all-targets` clean.

josefbacik:
**rust-bpf-code-reviewer follow-up pass** — covers the two commits added after the first verdict (integration-test matrix + reviewer-definition update), plus a cumulative re-check of the full diff. Mechanics as before: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over `origin/main..HEAD`, this time using the updated definition carried by this PR (so the new Python Version Support criteria were applied to the very diff that introduces them).

### Summary
The follow-up commits close the loop the first pass could only note: 3.14 is now exercised end-to-end (setup script + per-version integration test against a live 3.14.6 interpreter), and the review agent gains a completeness gate so the next version-add can't ship half-done. No walker, BPF, or schema changes; both commits are matrix/tooling only.

### Critical Issues
None.

### Improvements
None on this delta.

### Suggestions
- `setup-pystacks.sh`'s header comment says the array "must match PYTHON_VERSIONS in tests/trace_validation.rs"; the test file holds per-version consts rather than a single array. Pre-existing wording, and the new criteria's version-agreement check papers over it in practice — fine to leave.

### Positive Notes
- Version agreement holds exactly: `3.14.6` in `setup-pystacks.sh` == `PYTHON_314_VERSION` — the pair the new reviewer criterion pins.
- Self-test of the new gate: this PR's own cumulative diff satisfies all four Python Version Support checks (bindings+offsets+generator ⇒ setup entry ✓, matrix test ✓, identical release ✓, fallback arm advanced with its unit test ✓). Under the updated definition, the *original* first commit alone would have been flagged Critical for missing items 1–2 — which is precisely the gap this gate exists to catch.

### Verification
Full 3.8–3.14 pystacks matrix green in a BPF-capable VM (virtme-ng, the repo's v6.12 vmtest kernel): `7 passed; 0 failed`, including `test_pystacks_python314` resolving the test function symbol from a live 3.14.6 frame walk. Lib tests, `cargo fmt --check`, and clippy remain clean.

PR #154, opened 2026-07-14
Bump version and tag on merge instead of in PRs

josefbacik:
**Move version bumps out of PRs and into a bump-on-merge workflow.**

Every PR currently edits `version` in Cargo.toml, so any two open PRs conflict on that line and a merge burst needs serial rebases. This PR removes that coupling: PRs stop touching the version entirely, and a small workflow bumps + tags on every merge to main.

Mechanics:
- **Patch bump per merge** (`1.11.x -> 1.11.x+1`), annotated tag `vX.Y.Z`. Version keeps mapping 1:1 to merged content.
- **Minor bumps** (schema changes): the PR carries `[bump minor]` in its title; the squash/merge commit message delivers it to the workflow. SCHEMA_VERSION / SCHEMA_CHANGES.md requirements are unchanged.
- **No self-trigger**: bump commits are prefixed `release:` and the workflow skips them.
- **No burst races**: a concurrency group serializes bump jobs, each run rebases onto the live tip before computing, and the `[bump minor]` check reads the *triggering* commit by pinned sha (not HEAD, which may already be a newer release commit).

Tradeoff considered: batch-debounced bumps (one bump per burst) would produce fewer tags but break the version-to-content mapping that build identification relies on; per-merge keeps that property and tag noise is cheap at this merge cadence.

The in-repo reviewer checklist is updated in the same PR: the "must bump version" criteria become "must NOT touch version" (workflow owns it), with the `[bump minor]` title token replacing the manual minor-bump rule for schema changes.

Suggested merge order for the open ladder: #150 / #151 / #152 first (their version edits are valid under current rules and already sequence cleanly), then this PR — after which no open PR needs a de-version rebase.

### DIFF
--- a/.claude/agents/rust-bpf-code-reviewer.md
+++ b/.claude/agents/rust-bpf-code-reviewer.md
@@ -74,8 +74,8 @@ When changes touch the database schema (`src/duckdb.rs` `create_schema()`, `src/
 
 1. **SCHEMA_CHANGES.md updated**: Every schema change must have a corresponding entry in `SCHEMA_CHANGES.md` describing what changed
 2. **SCHEMA_VERSION incremented**: The `SCHEMA_VERSION` constant in `src/duckdb.rs` must be incremented for any schema change
-3. **Minor version bump**: Schema changes require a **minor** version bump in `Cargo.toml` (e.g., 1.0.0 → 1.1.0)
-4. **Patch version bump**: Non-schema changes require a **patch** version bump in `Cargo.toml` (e.g., 1.0.0 → 1.0.1)
+3. **No version edits in PRs**: `version` in `Cargo.toml` (and the matching `Cargo.lock` entry) must NOT be modified by a PR — the version-bump workflow bumps and tags on every merge to main. Flag any PR that touches it as a **Critical Issue**.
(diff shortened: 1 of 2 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict: PASS** (executed over the full 2-commit diff pre-push; mechanics: the in-repo agent definition run against the workflow YAML + checklist edit — no Rust/BPF surface in this diff, so the applicable passes were correctness, versioning semantics, and integration).

Two issues found during review, both fixed before push:
1. **Lightweight tags wouldn't publish** — `--follow-tags` only pushes *annotated* tags; `git tag` created lightweight ones. Fixed: `git tag -a "v$new" -m "v$new"`.
2. **Burst ordering read the wrong commit for `[bump minor]`** — after the rebase-onto-live-tip step, `git log -1` can be a newer run's `release:` commit rather than the triggering merge. Fixed: the token check reads the triggering commit by pinned sha (`github.event.head_commit.id`).

Residual notes (accepted, documented): a human commit authored with a `release:` prefix skips the bump (consistent with the release-commit convention this introduces); `cargo update --workspace` relies on the runner's stock Rust toolchain, and a missing toolchain fails the job loudly rather than silently.

PR #157, opened 2026-07-14
frame labels: decompose JIT output by runtime and name residual classes

josefbacik:
Frame labels today decompose gVisor mechanics well, but the biggest unresolved class on JIT-heavy workloads — the runtime's own compiled output — is only partially attributed, for four reasons this PR fixes together:

1. **The guest-maps chain never JIT-classifies.** For a correlated sandbox process the guest view answers first, and it renders every anonymous guest page `[gvisor:guest]` — so the JIT output of a sandboxed node/JVM/python service is lumped into generic guest memory even though the guest's own maps name `libnode.so`/`libjvm.so` with real paths and carry per-mapping exec permissions. On a gVisor-based cluster we operate, sandbox-family stacks still render >50% raw hex after the label deployment while `[jit:*]` accounts for only a few percent; this gap is the main term.
2. **Reclaimed JIT pages fall through every view to bare hex.** Maps (host) and the sandbox snapshot (guest) are read once per symbolization pass, after the sampling window; JIT runtimes allocate and free code pages faster than that. A sampled page that was recycled before the read is in *no* view, and today renders as raw hex — on V8-heavy processes this is the dominant residual.
3. **The runtime table is narrow and its matching is loose.** Only node/libnode, libjvm, libpython3 and libv8 are recognized; `starts_with` matching also lets `node_exporter` mark a process as a node runtime (dormant today only because Go emits no anonymous executable pages).
4. **What remains has no name.** An anonymous executable page in a process with no recognized runtime, a non-executable anonymous page, and an unmapped address all render identically as hex, though they are observably different things.

**Changes**

- `gvisor_guest`: detect the JIT runtime from the guest's file mappings at snapshot parse (`GuestProcess.jit_runtime`), and classify anonymous *executable* guest pages of such a process as `GuestAddr::Jit(<rt>)` → `[jit:<rt>]`. Non-executable and unrecognized-runtime anonymous pages stay `[gvisor:guest]`.
- `sandbox_maps`: unmapped addresses now label — `[jit:<rt>]` when the process runs a recognized JIT runtime (see the note below), `[unmapped]` otherwise. Anonymous pages in non-sandbox processes name their observable class when no runtime is recognized: `[anon:exec]` (executable) / `[anon]` (not). Bracketed pseudo-entries (`[vdso]`, `[stack]`) are untouched. Net: raw hex from a live process now means "maps unreadable" and nothing else.
- Runtime table: add ruby/libruby, libcoreclr (dotnet), beam.smp, libluajit (lua), libjulia, deno, bun, pypy/libpypy, nodejs, and `python3` for statically linked interpreters whose runtime lives in the exe. Matching now requires a version/extension boundary after the prefix — `node` matches `node`, `node22`, `libnode.so.115`, but no longer `node_exporter`.

**Labeling-semantics note for review — `[jit:<rt>]` now includes an inference.** Previously the tag was only applied to *observed* anonymous executable pages. This PR also applies it to *unmapped* addresses in a process running a recognized JIT runtime, on the argument that page recycling between sample and maps-read is overwhelmingly the mechanism that produces them there (and for a correlated sandbox stub it is the only mechanism — everything actually mapped gets some label). The cost: a garbage frame from an unwind glitch in such a process is now labeled `[jit:<rt>]` rather than hex, so the tag is no longer a page-table-verified fact. If you want the two distinguishable downstream, the alternative is a distinct tag for the inferred case — `[jit:<rt>:reclaimed]` — which is a one-line change; the PR ships the merged form because flamegraph consumers aggregate by runtime and splitting the tag halves both bars, but it's a semantics call worth making consciously.

**Vocabulary after this PR** (all behind the existing `--no-frame-labels` flag, rendered in the established `unknown (<label>) <0xADDR>` miss format): module basename, `[gvisor:runtime]`, `[gvisor:guest]`, `[jit:<rt>]` (rt ∈ node, jvm, python, v8, ruby, dotnet, beam, lua, julia, deno, bun, pypy), `[anon:exec]`, `[anon]`, `[unmapped]`, `[exited]`.

**Verification**: full lib tests including new units for the boundary matcher (positive forms and the `node_exporter`/`bundler`/`denoise` rejections), guest JIT classification (JIT guest / plain guest / vdso exclusion / verbatim-fixture regression), and every new `label_for` arm; `cargo fmt --check` and `clippy --all-targets` clean; `gvisor_record` and `trace_validation` integration targets green in a BPF-capable VM (virtme-ng, the repo's v6.12 vmtest kernel). No version change per the new merge-workflow convention; non-schema, so no `[bump minor]` marker.

### DIFF
--- a/src/gvisor_guest.rs
+++ b/src/gvisor_guest.rs
@@ -97,6 +97,12 @@ pub struct GuestProcess {
     status: GuestStatus,
     #[serde(default)]
     maps: Vec<GuestMapping>,
+    /// JIT runtime detected among the guest's file mappings. The guest
+    /// view carries real paths, making it the authoritative signal for
+    /// classifying the guest's anonymous executable pages. Computed after
+    /// deserialization in [`parse_dump_response`].
+    #[serde(skip)]
+    jit_runtime: Option<&'static str>,
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
@@ -119,7 +125,11 @@ pub enum GuestAddr {
     },
     /// Runtime-injected region (`[usertrap]`): sandbox overhead.
     Runtime,
-    /// Guest-anonymous memory (JIT output, heap made executable, ...).
+    /// Guest-anonymous executable memory in a process running the tagged
+    /// JIT runtime: compiled output.
+    Jit(&'static str),
+    /// Guest-anonymous memory (heap, copied-up pages, executable pages of
+    /// an unrecognized runtime).
     Anon,
     /// Not mapped in the guest at all.
     Unmapped,
@@ -153,6 +163,12 @@ impl GuestProcess {
                 }
             }
             "[usertrap]" => GuestAddr::Runtime,
+            // Executable anonymous guest memory; bracketed pseudo-regions
+            // ([vdso], [vvar], ...) are runtime-provided, never JIT output.
+            path if m.perms.execute && !path.starts_with('[') => match self.jit_runtime {
+                Some(rt) => GuestAddr::Jit(rt),
+                None => GuestAddr::Anon,
+            },
             _ => GuestAddr::Anon,
         }
     }
@@ -337,7 +353,16 @@ fn parse_dump_response(buf: &[u8]) -> std::io::Result<Vec<GuestProcess>> {
     if !resp.success {
         return Err(std::io::Error::other(format!("urpc error: {}", resp.err)));
     }
-    Ok(resp.result)
+    let mut procs = resp.result;
+    for p in &mut procs {
+        p.jit_runtime = p
+            .maps
+            .iter()
+            .filter(|m| m.is_file())
+            .filter_map(|m| Path::new(&m.pathname).file_name().and_then(|f| f.to_str()))
+            .find_map(crate::sandbox_maps::detect_jit_runtime);
+    }
+    Ok(procs)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over the full diff (`origin/main..HEAD`, 3 files +218/−25) before push, per this repo's review convention.

### Summary
Focused classification change: the guest-maps chain gains JIT awareness, the runtime table grows behind a boundary-aware matcher, and the remaining unlabeled classes get honest names. No BPF, schema, or recorder changes; the new arms make `label_for`'s match exhaustive over the backing/exec space, so a future `Backing` variant becomes a compile error instead of a silent fallthrough — a structural improvement.

### Critical Issues
One found during review, fixed before commit: the first version of the guest-side classification labeled any executable guest mapping of a JIT-runtime process as `[jit:<rt>]` — including bracketed pseudo-regions, and a guest `[vdso]` *is* executable, so vdso samples in a node/JVM guest would have been misattributed as JIT output. The committed version excludes bracket-prefixed pathnames from the JIT arm (host side was already safe via `Backing::Special`), with a fixture assertion pinning it (`[vdso]` in a node guest → `GuestAddr::Anon`).

### Improvements
None outstanding.

### Suggestions
- `parse_dump_response` hydrates `jit_runtime` with first-match over guest maps while the host-side parse keeps its historical last-match; the table has no prefix pair mapping one basename to different tags, so the results cannot differ today — noting it in case the table ever grows such a pair.

### Positive Notes
- The boundary matcher closes a latent false positive that shipped with the original table (`node_exporter` marked processes as node runtimes; dormant only because Go emits no anonymous executable pages) and the rejection cases are unit-tested by name.
- Schema & Versioning gates: no schema change; no version edit, per the new bump-on-merge workflow (non-schema, so no `[bump minor]` title marker either). SCHEMA_CHANGES.md correctly untouched.

### Verification
403/403 lib tests (new: matcher boundary matrix, guest JIT classification incl. vdso exclusion and verbatim-fixture regression, every new `label_for` arm); fmt/clippy clean; `gvisor_record` + `trace_validation` green in the BPF VM gate at the exact pushed tree.

PR #158, opened 2026-07-14
gopclntab: find the table in sectionless PIE binaries

josefbacik:
The gopclntab resolver located the table by `section_by_name(".gopclntab")` only. That name doesn't always exist: distro external-linking PIE builds can merge the pclntab anonymously into `.data.rel.ro`, leaving no dedicated section — the binary is correctly identified as stripped Go, the resolver reports not-a-candidate, and every frame renders as `<binary>:unknown`. In production profiling data this presents as a stripped Go binary that stays ~100% unresolved while its neighbors (e.g. an upstream-built kubelet, which keeps the named section) symbolize fully.

**The artifact evidence** (public packages, reproducible): Amazon Linux 2023's containerd RPMs, pulled from the AL2023 CDN — `containerd-1.7.27` (go1.23.8) and `containerd-1.7.11` (go1.20.12). Both are PIE, external-linked, RPM-stripped (no `.symtab`), and have **no `.gopclntab` section at all**; the GO120-magic header sits inside `.data.rel.ro` (at `+0x4f49a0` in the 1.7.27 build) with the table fully intact. The shape spans the distro's whole build history, not one toolchain version. `ElfPclntab::from_path` on these binaries returned `Ok(None)` before this change.

**The fix — three location tiers:**

1. `.gopclntab` by name (unchanged fast path, including the existing malformed-metadata errors).
2. `.data.rel.ro.gopclntab` by name (the dedicated-name variant some PIE links emit).
3. A budgeted scan of non-executable data sections (`.data.rel.ro` and `.rodata` first, then remaining data sections) for a validating table: the 8-byte header signature (known magic, mandatory zero pad, plausible instruction quantum and pointer size) is matched at 8-byte-aligned offsets, and each candidate must survive the full header parse — its offset words must all land in-bounds — which rejects signature lookalikes. Empty tables are skipped rather than allowed to shadow a real one. Total scanned bytes are capped (1 GiB) so adversarial section layouts can't make the probe super-linear; real binaries resolve within the first few dozen MiB of `.data.rel.ro`.

**Negative controls:** binaries with a `.symtab` still defer to the richer symtab+DWARF path (`has_symtab` semantics untouched) — COS-style kubelet builds are unaffected; binaries with the named section never reach the scan, so the existing fleet behavior is byte-identical for them. Scan failures are "not a candidate", never errors: a binary with no findable table degrades exactly as today.

**Verification:** the real AL2023 binaries go from not-a-candidate to **59,228** (1.7.27) and **56,283** (1.7.11) resolved functions, and the existing full-table sweep test (`GOPCLNTAB_TEST_BINARY`) passes against them — every functab entry resolves back to itself. New unit tests cover the embedded-table scan (with a signature-decoy that must be rejected), unaligned/garbage rejection, and a real-toolchain fixture whose `.gopclntab` is renamed in `.shstrtab` — the bytes stay put, the name is gone, the scan must find it (the closest simulation of the sectionless link). 16/16 crate tests, full workspace lib suite, fmt and clippy clean.

No version change per the merge-workflow convention; non-schema. The `gopclntab` crate is published separately — whether this warrants a crates.io re-publish is left to the maintainer.

### DIFF
--- a/crates/gopclntab/src/elf.rs
+++ b/crates/gopclntab/src/elf.rs
@@ -1,16 +1,26 @@
-//! Locate and extract `.gopclntab` from ELF binaries.
+//! Locate and extract the pclntab from ELF binaries.
 //!
 //! ELF parsing is delegated to the [`object`] crate; reads are targeted —
 //! headers and the sections consulted, then exactly the pclntab byte
 //! range — so probing large binaries (Go binaries run to hundreds of MiB)
-//! never reads whole files.
+//! never reads whole files. The exception is the sectionless fallback
+//! below, which must read the data sections it scans.
 //!
-//! Binaries that are not ELF, or have no `.gopclntab` section, report
-//! "not a candidate" (`Ok(None)`) rather than an error. Section metadata
-//! is validated against the input's actual size, so the only inputs
-//! rejected as malformed are ones whose headers are inconsistent with the
-//! bytes on disk — there is no size threshold a large-but-valid binary
-//! could trip over.
+//! The table is located by section name first — `.gopclntab`, then the
+//! `.data.rel.ro.gopclntab` name some PIE links use. When neither
+//! exists, non-executable data sections are scanned for a validating
+//! pclntab header: distro external-linking PIE builds (observed on
+//! Amazon Linux 2023's containerd packages across go1.20–go1.23) merge
+//! the table anonymously into `.data.rel.ro`, and the header is its only
+//! remaining identification. See [`GoPclntab::scan`].
+//!
+//! Binaries that are not ELF, or in which no table can be located,
+//! report "not a candidate" (`Ok(None)`) rather than an error. Section
+//! metadata is validated against the input's actual size, so the only
+//! inputs rejected as malformed are ones whose headers are inconsistent
+//! with the bytes on disk — there is no size threshold a large-but-valid
+//! binary could trip over (the scan's own budget caps work done, not
+//! input size).
 //!
 //! Policy is deliberately left to the caller: this module reports whether
 //! a symbol table is present ([`ElfPclntab::has_symtab`]) but does not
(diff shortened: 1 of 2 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over the full diff (`origin/main..HEAD`, 2 files +237/−40) before push, per this repo's review convention.

### Summary
Contained change in the `gopclntab` crate: table location grows two fallback tiers (a second dedicated name, then a budgeted validating scan of data sections) while the named fast path — including its error behavior — is preserved verbatim via extraction into `read_named_section`. No systing-side changes needed: the resolver's `from_path` entry point gains the capability transparently. No BPF, schema, or recorder surface.

### Critical Issues
None.

### Improvements
None required. One consideration weighed and accepted: `GoPclntab::scan` copies `region[off..]` into a `Vec` per validating candidate (the parse API owns its bytes). The signature pre-filter makes false-positive candidates vanishingly rare, so in practice this is a single copy of the table tail — the real containerd probe confirms sub-second end-to-end.

### Suggestions
- If a future binary shape puts the table in an executable or unusual section kind, the scan's `Data | ReadOnlyData` filter would miss it; the module doc states the covered shapes, so this is a documented boundary rather than a silent one.

### Positive Notes
- The empty-table guard in `scan` (a parseable zero-function header cannot shadow a later real table) closes a subtle shadowing hazard before it could ever fire.
- The rename-fixture test simulates the sectionless shape against a real toolchain build by patching `.shstrtab` only — bytes stay put, name disappears — which is precisely the failure geometry observed in the wild, and it asserts agreement with the named path on the same binary.
- Versioning gates: no version edits (merge-workflow convention), no schema surface, SCHEMA_CHANGES.md correctly untouched.

### Verification
16/16 crate tests (new: scan positive with decoy rejection, unaligned/garbage rejection, renamed-section fixture); full workspace lib suite green; fmt/clippy clean; real-artifact receipts — AL2023 containerd 1.7.27 and 1.7.11 parse to 59,228 / 56,283 functions with the full-table sweep passing against both.

PR #160, opened 2026-07-15
analyze: let the MCP launcher bound DuckDB's on-disk spill

josefbacik:
## Problem

`systing-analyze mcp` opens trace databases with DuckDB's default `max_temp_directory_size`, which DuckDB derives from the filesystem's free space. On quota-limited scratch that number is a lie: a Kubernetes emptyDir `sizeLimit` is enforced by a periodic kubelet scan, not by the filesystem, so `statvfs` reports the whole node's free disk. A single spilling aggregation against a ~5M-row trace was observed writing >10 GiB into a 10 GiB scratch volume in under two minutes — the kubelet then evicts the entire pod mid-run, killing all of the run's bookkeeping, which is a strictly worse failure than the one query erroring out.

## Change

- New `--max-temp-directory-size <SIZE>` on the `mcp` subcommand (any DuckDB size string, e.g. `8GiB`), threaded through to **every** database the server opens — the initial `-d` database and lazily-opened ones alike.
- Applied on the `duckdb::Config` (like `temp_directory`), so the bound is in force from the first query and a malformed size string fails the open loudly instead of being ignored.
- `AnalyzeDb::open` is unchanged for all existing callers; it now delegates to the new `open_with_spill_cap`.

## What the cap does and does not promise

With the cap set, an oversized aggregation fails **that query** with DuckDB's temp-directory-limit error; the caller (an AI assistant driving the MCP server, in the motivating case) can rephrase or narrow it, and the process survives.

It is a structural default, not a jail: `enable_external_access(false)` locks `temp_directory` but — verified empirically, not assumed — does **not** lock `max_temp_directory_size`, so in-session SQL can still raise it deliberately. That is the right trade for the failure mode this addresses (queries that never think about spill at all), and `test_spill_cap_is_session_overridable` pins the boundary so a future DuckDB that starts locking the setting gets noticed and the docs can promise more.

## Tests

- `test_open_with_spill_cap_applies_bound` — capped open reports the bound; uncapped open differs (non-vacuous).
- `test_open_with_invalid_spill_cap_fails` — malformed size fails the open.
- `test_spill_cap_is_session_overridable` — documents the trust boundary above.

`cargo clippy --lib --bins -- -D warnings` clean; full lib suite 406/406.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -246,6 +246,30 @@ pub struct AnalyzeDb {
 impl AnalyzeDb {
     /// Open a trace database.
     pub fn open(path: &Path, read_only: bool) -> Result<Self> {
+        Self::open_with_spill_cap(path, read_only, None)
+    }
+
+    /// Open a trace database, optionally bounding DuckDB's on-disk spill.
+    ///
+    /// DuckDB derives its default `max_temp_directory_size` from the
+    /// filesystem's free space, which quota-limited environments misreport:
+    /// a Kubernetes emptyDir `sizeLimit` is enforced by a periodic kubelet
+    /// scan, not by the filesystem, so `statvfs` reports the whole node's
+    /// free disk and a single spilling query can blow past the quota and get
+    /// the pod evicted mid-run. A caller that knows its real scratch budget
+    /// passes it here (any DuckDB size string, e.g. "8GiB") so an oversized
+    /// aggregation fails that one query instead of killing the process.
+    ///
+    /// This is a structural default, not a jail: unlike `temp_directory`,
+    /// DuckDB does not lock this setting when external access is disabled,
+    /// so in-session SQL can still raise it (covered by a test). That is the
+    /// right trade for the intended failure mode — queries that never think
+    /// about spill — and it keeps deliberate in-session tuning possible.
+    pub fn open_with_spill_cap(
+        path: &Path,
+        read_only: bool,
+        spill_cap: Option<&str>,
+    ) -> Result<Self> {
         if !path.exists() {
             bail!("Database not found: {}", path.display());
         }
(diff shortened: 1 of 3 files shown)


josefbacik:

*Mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition (blob `f09e9b60`) was executed in-session by the authoring seat over the full 3-file diff, per the team's pre-push review protocol.*

### Summary
Small, well-scoped plumbing change: a launcher-provided bound on DuckDB's on-disk spill, threaded from a new `--max-temp-directory-size` flag on the `mcp` subcommand through to every database open the MCP server performs. The delegation pattern (`open` → `open_with_spill_cap`) keeps all ten pre-existing `AnalyzeDb::open` call sites untouched, and the trust boundary of the new setting is measured and pinned by a test rather than assumed.

### Critical Issues
None.
- Version gate: `Cargo.toml`/`Cargo.lock` untouched (bump-on-merge owns the version).
- Schema gate: no schema changes; `SCHEMA_VERSION` untouched; no `[bump minor]` needed.
- Recorder gate: N/A.

### Improvements
None required. The one design fork worth recording: a generic `--set key=value` pass-through was considered and rejected — `AnalyzeDb::open` applies settings in a deliberate order (`temp_directory` before `enable_external_access(false)`, which locks it), and a generic mechanism would let callers silently violate that ordering or fight `create_spill_dir`. The targeted flag preserves the hardening invariants.

### Suggestions
- The standalone `systing-analyze` subcommands (query/flamegraph/etc.) could accept the same flag; left out deliberately — they run on interactive dev boxes where DuckDB's free-space default is accurate. Add if a real need appears.
- `spill_cap` (internal name) vs `max_temp_directory_size` (CLI/API name): deliberate layering — the public surface uses DuckDB's own setting name for discoverability; internals name the intent.

### Positive Notes
- `test_spill_cap_is_session_overridable` pins what the cap does NOT promise (in-session SQL can raise it — DuckDB locks `temp_directory` under external-access-off but not this setting; verified empirically, not assumed). If a future DuckDB starts locking it, the test fails and the docs get to promise more.
- Malformed size strings fail the open loudly (`test_open_with_invalid_spill_cap_fails`) rather than being silently ignored.
- Doc-comments carry the operational why: kubelet's emptyDir sizeLimit is a periodic scan invisible to `statvfs`, so DuckDB's free-space-derived default overshoots the real budget and the pod gets evicted instead of the query failing.

Validation: `cargo clippy --lib --bins -- -D warnings` clean; full lib suite 406/406 green (3 new tests included); fmt applied.

PR #161, opened 2026-07-15
gopclntab: validate scan candidates before copying, gate on Go metadata

josefbacik:
Follow-up hardening for the sectionless-pclntab scan from #158, from a post-merge review of the scan tier's adversarial bounds. Two fixes:

**1. Candidate validation no longer allocates — kills a quadratic on adversarial content.** The scan copied `region[off..]` into a `Vec` for every signature-matching candidate *before* validating it. A region tiled with fake 8-byte signatures therefore cost a full tail copy per candidate: measured ~1.6 s per MiB and ×4 per size doubling on the merged code — extrapolating to days at the 1 GiB budget. That contradicts #158's claim that the budget prevents super-linear behavior: the budget caps bytes *read*, not per-candidate copy cost — that sentence in #158's description was wrong, and this PR is the correction. Reachable as an availability problem: the probe runs on every ELF member of every profiled process, so one crafted stripped binary in a workload could wedge a symbolization pass. The fix splits the owned parse into a borrowed header validation (`parse_header`, all the same checks including the functab-extent bound) plus construction; candidates are rejected in constant time on the slice, and bytes are copied exactly once — for the adopted table. Benign binaries see identical behavior (they have at most one candidate).

**2. The scan tier now requires Go build metadata.** Previously every stripped ELF with no named table paid a full read of its data sections (transient RSS up to section sizes, plus a ~0.23 s/GiB sweep) where a name-miss used to be near-free — a real cost on memory-sensitive deployments, charged mostly to non-Go binaries that could never benefit. The scan now runs only when `.go.buildinfo` or `.note.go.buildid` is present; both survive external linking and stripping, and both are present on the sectionless distro builds the scan exists for (verified on the AL2023 containerd artifacts from #158). This also structurally narrows the foreign-table case noted in review — a non-Go binary embedding a complete Go ELF in its data is no longer scanned at all; for a Go binary embedding another Go ELF the first validating table still wins, which the module doc now says explicitly.

One review note accepted as-is: an adopted table retains bytes from its start to the end of its host section rather than its exact extent (the true end isn't cheaply derivable from the header). Bounded by section size, one allocation, benign shapes are small — documented rather than engineered around.

**Verification:** new `scan_rejects_tiled_signatures_in_linear_time` test (a 2 MiB fully-tiled region must complete in milliseconds — it would take seconds-to-minutes with the pre-fix per-candidate copy) and a gate test (real-toolchain fixture with `.gopclntab` *and* all Go marker sections renamed away in `.shstrtab` must report not-a-candidate, while the marker-intact rename fixture from #158 continues to prove the scan path works). The real AL2023 containerd binaries (1.7.27/go1.23.8, 1.7.11/go1.20.12) still resolve their full tables through the gate — full-table sweep green against both. 17/17 crate tests, full workspace suite, fmt and clippy clean.

No version change per the merge-workflow convention; non-schema.

### DIFF
--- a/crates/gopclntab/src/elf.rs
+++ b/crates/gopclntab/src/elf.rs
@@ -93,13 +93,28 @@ impl ElfPclntab {
             return Ok(Some(Self { table, has_symtab }));
         }
 
-        // No dedicated section: scan non-executable data sections for a
-        // validating header, likeliest homes first. Failures here are
-        // "not a candidate", not errors — this path is discovery, and
-        // sections that cannot be read or hold no table are simply
-        // skipped. The budget bounds total work on adversarial inputs;
-        // real binaries put the table within the first few dozen MiB of
-        // `.data.rel.ro` or `.rodata`.
+        // No dedicated section. The scan below reads whole data sections,
+        // which every stripped ELF that reaches this point would pay —
+        // so first require evidence the binary is Go at all. Both marker
+        // sections survive external linking and stripping (they are
+        // present on the sectionless distro builds this fallback exists
+        // for), unlike `.gopclntab` itself. This also narrows foreign-
+        // table adoption: a non-Go binary embedding a complete Go ELF in
+        // its data is no longer scanned at all.
+        if elf.section_by_name(".go.buildinfo").is_none()
+            && elf.section_by_name(".note.go.buildid").is_none()
+        {
+            return Ok(None);
+        }
+
+        // Scan non-executable data sections for a validating header,
+        // likeliest homes first. Failures here are "not a candidate",
+        // not errors — this path is discovery, and sections that cannot
+        // be read or hold no table are simply skipped. The budget bounds
+        // total bytes read on adversarial inputs (candidate rejection
+        // within a region is constant-time per candidate — see
+        // [`GoPclntab::scan`]); real binaries put the table within the
+        // first few dozen MiB of `.data.rel.ro` or `.rodata`.
         const SCAN_BUDGET: u64 = 1 << 30;
         let mut budget = SCAN_BUDGET;
         let mut sections: Vec<_> = elf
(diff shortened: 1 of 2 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over the full diff (`origin/main..HEAD`, 2 files +135/−25) before push, per this repo's review convention.

### Summary
Tight hardening pass on the #158 scan tier: the owned parse splits into borrowed validation plus construction (single source of truth for the checks — `parse` delegates, no duplicated logic), and a two-section Go-metadata gate fronts the scan. Public API unchanged; named-path behavior unchanged; the fix path for the sectionless distro binaries re-verified against the real artifacts.

### Critical Issues
None.

### Improvements
None outstanding.

### Suggestions
- `scan`'s adopt arm re-runs the O(1) header parse on the copied bytes rather than threading the validated `Header` through; deliberate simplicity (the comment marks the parse as infallible there) and free at one adoption per scan.

### Positive Notes
- The tiled-signature test pins the performance contract behaviorally: a fully-tiled 2 MiB region (262k candidates, each rejected at the bounds checks) must finish within a generous wall-clock bound that the pre-fix quadratic missed by an order of magnitude — regression-proof without flakiness.
- The gate test asserts its own premise (`markers > 0`) before asserting the gate, so a future toolchain that stopped emitting marker sections would fail the fixture loudly instead of vacuously passing.
- Versioning gates: no version edits (bump-on-merge workflow), no schema surface.

### Verification
17/17 crate tests (new: linear-time tiling, gate-negative), full workspace lib suite green, fmt/clippy clean, real-artifact receipts re-run through the gate (containerd 1.7.27 and 1.7.11 full-table sweeps pass).

PR #162, opened 2026-07-17
Add support for riscv64

kxxt:
This PR adds support for riscv64.

- The first commit updates the vmlinux header generation script and includes my generated header on Linux riscv64 6.19.14.
- The second commit adds support for riscv64.
  - There is no equivalent field like `thread.fsbase`(x86) or `thread.uw.tp_value`(arm64) on riscv64, so the `struct pthread` is located by a different method.
- The third commit adds support for cross-compiling to riscv64 with cross-rs.
  -  I added `-Wl,--no-as-needed -latomic -Wl,--as-needed` as a workaround for missing libatomic linking of `libduckdb-sys`.
     Ideally this should be fixed in `libduckdb-sys` instead so I will submit a fix to `libduckdb-sys` in the future.


kxxt:
Tested on a riscv64 machine with command `sudo ./systing --collect-pystacks --pystacks-debug --duration 10`.

The resulting perfetto trace is [trace.pb.zip](https://github.com/user-attachments/files/30129982/trace.pb.zip)

josefbacik:
**rust-bpf-code-reviewer** — the repo agent definition (`.claude/agents/rust-bpf-code-reviewer.md`, blob `f09e9b60`) executed over the full diff at `5cdd513`, at the maintainer's request; structured verdict below, independent build verification at the end.

### Summary
Clean three-part riscv64 port: a generated `vmlinux_riscv64.h` plus an arch-aware generation script, a genuinely clever BPF-side answer to riscv64's missing kernel-side TLS-pointer field, and a cross-rs toolchain that faithfully mirrors the existing aarch64 pattern. The pthread-descriptor derivation via `clear_child_tid`/`robust_list` is sound — I re-derived the glibc layout math independently and it holds — and the shared-file refactor is arithmetically identical for x86-64/aarch64. No schema, event-format, or version surface touched.

**At a glance:** blocks merge — nothing. Worth folding into this PR — Improvements 1–2 (two-line message fixes) and Suggestions 1–3 (small polish). Fine as follow-ups — Improvement 3 (CI job) and Suggestion 4. Needs hardware — already covered by the author's attached riscv64 end-to-end trace; the one confirmation requested is that the `cross build` completes (see the verification section).

### Critical Issues
None.
- Version gate: `Cargo.toml`/`Cargo.lock` correctly untouched (bump-on-merge assigns the version).
- Schema gates: no schema change → `SCHEMA_VERSION`/`SCHEMA_CHANGES.md` correctly untouched; no `[bump minor]` needed in the title.

### Improvements
1. **`build.rs:182` — stale panic message.** The match arm now accepts riscv64 but the message still says "Only x86_64 and aarch64 are supported." Same at `build.rs:33` ("restricts supported targets to these two").
2. **`build.rs:49` — wrong hint for riscv64.** The cross-compile warning's `deb_arch` maps everything non-aarch64 to `amd64`, so a riscv64 cross build with missing headers would suggest `apt install libc6-dev:amd64`. Add a `"riscv64" => "riscv64"` arm.
3. **No CI coverage for the new target** — nothing in `.github/workflows` exercises riscv64, so the port will silently rot as BPF code evolves. A cheap guard that needs no container or qemu: compile the two BPF objects with `clang -target bpf -D__TARGET_ARCH_riscv -D__riscv64__` against the vendored header (that catches header/type/macro drift, which is the realistic rot vector — see verification below for the exact shape). The full `cross build` in CI would be even better if the container cost is acceptable.

### Suggestions
1. **`pthread_helpers.bpf.c:13` vs `:24-26` — validation asymmetry.** `get_riscv_pthread_descriptor` accepts any non-NULL `clear_child_tid`, while `get_riscv_specific1stblock` validates with `IS_VALID_USER_SPACE_ADDRESS`. A bogus non-NULL pointer fails gracefully downstream either way, but using the same validation in both makes the intent symmetric.
2. **`cross/riscv64.Dockerfile:13`** — comment says "aarch64 cross-compilation toolchain"; should say riscv64 (copy-paste from the aarch64 file).
3. **`pthread_helpers.bpf.c:70-71`** — trailing whitespace on the two new comment lines.
4. **`Cross.toml`** — `passthrough` entries in `NAME=value` form need a reasonably recent `cross`; older releases treat passthrough entries as names only, which would silently drop the `-latomic` workaround. Worth a one-line minimum-version note in `scripts/cross-build.sh`'s prerequisites.
5. Informational, no action: the 6.19-generated header emits five harmless "declaration does not declare anything" warnings (anonymous-scope forward declarations from newer BTF dumps, e.g. `vmlinux_riscv64.h:54355`). They'll appear in every riscv64 build log; the x86-64/aarch64 headers predate this bpftool behavior.

### Positive Notes
- The `clear_child_tid`/`robust_list` derivation is the highlight of the PR. Instead of chasing a nonexistent `thread_struct` field, it leans on two pointers the kernel itself maintains into `struct pthread`: glibc registers `&pd->tid` (via `set_tid_address` at startup and `CLONE_CHILD_CLEARTID` in `pthread_create`) and `&pd->robust_head` (via `set_robust_list`), for the main thread and spawned threads alike. I re-derived the offsets independently against glibc's `nptl/descr.h`: `tid` at 0xd0 (24-pointer union + 16-byte `list`), and `specific_1stblock` at +0x30 from `robust_head` in **both** robust-mutex layouts, exactly as the comment claims. The identity match also gets cheaper than the x86 path (one fewer probe read), and `pthread_self()` on riscv64 glibc returns precisely the descriptor address being compared.
- The `probe_read_pthread_tls_slot` refactor preserves x86-64/aarch64 addressing bit-for-bit (old: `tls_base+0x310+…` and `tls_base+0x310+0x200+…`; new: the same via `specific1stblock`). Good discipline in code shared across arches.
- `generate-vmlinux-header.sh`: the `BTF_ARGS` array cleanup deduplicates the module list against the invocation, and the `vfio_iommu_type1` guard leaves x86-64/arm64 output identical to before.
- `cross/riscv64.Dockerfile` mirrors the established aarch64 Dockerfile faithfully, including the include-path symlink strategy.
- Error paths are closed: NULL descriptor → `*value = 0` + error return in the TLS-slot reader; NULL `tls_base` → the existing zero-check returns `PYSTACKS_PTHREAD_ID_ERROR`.
- End-to-end validation on real hardware with the trace attached is exactly the right evidence for a port like this.

### Side observation (pre-existing, not this PR)
While verifying the refactor's aarch64 equivalence: the aarch64 constants look x86-shaped. On aarch64 glibc, `THREAD_SELF` is `(struct pthread *)__builtin_thread_pointer() - 1` — the descriptor sits *below* the thread pointer — so `tp_value + 0x310` and `tp_value + 0x10` would read static-TLS bytes rather than `struct pthread` fields (on x86-64, `fsbase` *is* the descriptor, so those constants are correct there). If pystacks has never been validated on aarch64 hardware, that path may be silently degraded — and the arch-generic `clear_child_tid`/`robust_list` approach introduced here would fix it. Question for the maintainer; out of scope for this PR.

### What was verified vs. not
**Verified independently:**
- Full diff read (all 9 non-generated files, plus targeted checks on the generated header); the glibc offset math re-derived rather than trusted.
- **riscv64 BPF build chain reproduced**: both translation units compiled with `clang -g -O2 -target bpf -D__TARGET_ARCH_riscv -D__riscv64__` — the exact defines the real build produces (libbpf-cargo 0.25.0 maps `riscv64` → `__TARGET_ARCH_riscv`; checked in its source) — against this PR's `vmlinux_riscv64.h` and the locked libbpf 1.5.0 headers, then linked with `bpftool gen object`: **clean**. This proves the 146k-line generated header is complete for every kernel type systing's BPF references, in both the pystacks and main objects.
- x86-64 baseline: identical procedure compiles clean (and this PR's CI is green: clippy, rustfmt, test suite).
- `vmlinux_riscv64.h` sanity: standard bpftool structure, `clear_child_tid`/`robust_list` present, vfio core types present, size plausible next to the x86-64/aarch64 headers.
- Version/schema gates as listed above.

**Not verified here (relying on the author's hardware test):** runtime load/attach and verifier acceptance on a riscv64 kernel; the 0xd0/+0x30 offsets against a live riscv64 glibc process; the cross-rs container build end-to-end (the `-latomic` workaround reaching link stage implies it was exercised — a one-line confirmation that `cross build --target riscv64gc-unknown-linux-gnu` completes would be good to have on the record). My riscv64 compile shimmed `errno.h` to `asm-generic/errno.h` on a non-riscv host — the same resolution the Dockerfile's fallback symlink produces — so the container's exact glibc header interplay wasn't reproduced.

kxxt:
Thanks for your review!

> ### Improvements
> 
>     1. **`build.rs:182` — stale panic message.** The match arm now accepts riscv64 but the message still says "Only x86_64 and aarch64 are supported." Same at `build.rs:33` ("restricts supported targets to these two").

Fixed.

>     2. **`build.rs:49` — wrong hint for riscv64.** The cross-compile warning's `deb_arch` maps everything non-aarch64 to `amd64`, so a riscv64 cross build with missing headers would suggest `apt install libc6-dev:amd64`. Add a `"riscv64" => "riscv64"` arm.

Done. I also made the fallback path a panic to avoid fallback to amd64 on unsupported architectures.

>     3. **No CI coverage for the new target** — nothing in `.github/workflows` exercises riscv64, so the port will silently rot as BPF code evolves. A cheap guard that needs no container or qemu: compile the two BPF objects with `clang -target bpf -D__TARGET_ARCH_riscv -D__riscv64__` against the vendored header (that catches header/type/macro drift, which is the realistic rot vector — see verification below for the exact shape). The full `cross build` in CI would be even better if the container cost is acceptable.

I can include a CI pipeline if you see fit. I think a cross build with `cross-rs` should provide some coverage to avoid silently breaking compilation on riscv64 when updating the BPF code.

For my personal projects. I am using a nix-based CI, which tests the code on many (CPU Architecture, Kernel version, Optionally LLVM version) combos with QEMU (https://github.com/kxxt/tracexec/actions/runs/29251312524#summary-86820293616). Though I am not sure if you would like such extensive coverage.

> ### Suggestions
> 
>     1. **`pthread_helpers.bpf.c:13` vs `:24-26` — validation asymmetry.** `get_riscv_pthread_descriptor` accepts any non-NULL `clear_child_tid`, while `get_riscv_specific1stblock` validates with `IS_VALID_USER_SPACE_ADDRESS`. A bogus non-NULL pointer fails gracefully downstream either way, but using the same validation in both makes the intent symmetric.

I updated the code to also add a user-space address check for `get_riscv_pthread_descriptor`.

>     2. **`cross/riscv64.Dockerfile:13`** — comment says "aarch64 cross-compilation toolchain"; should say riscv64 (copy-paste from the aarch64 file).

Fixed.

>     3. **`pthread_helpers.bpf.c:70-71`** — trailing whitespace on the two new comment lines.

Fixed.

josefbacik:
**Delta-read of the 5cdd5135 → 92192e0f push** (tree-level interdiff, both ends pinned; the rewrite folds the fixes into the original two commits in place — titles stable, the vmlinux-header commit untouched, no base drift, and the version/schema surfaces still untouched).

Interdiff = 3 files, +11/−8, against the review items:

- **Improvement 2 (deb_arch hint)** — addressed beyond the ask: the aarch64/else-amd64 branch is now an explicit match with `riscv64` and `x86_64` arms and a panic on anything else, which also removes the silent-amd64 default entirely.
- **Improvement 1 (stale messages)** — the `get_arch_config()` panic now says "x86_64, aarch64 and riscv64". **One residual:** the comment above the cross-include triplet match still reads "restricts supported targets to these two" over what is now a three-entry list — comment-only, non-blocking; one word if the branch gets touched again.
- **Suggestion 1 (validation symmetry)** — `get_riscv_pthread_descriptor` now applies `IS_VALID_USER_SPACE_ADDRESS` to `clear_child_tid`, matching the specific1stblock path. (The trailing NULL ternary is redundant behind the new check, but harmless.)
- **Suggestions 2 + 3** — the Dockerfile comment now says riscv64, and the trailing whitespace is gone.

Nothing new entered the pystacks/BPF surfaces beyond the two asked changes. The prior verdict extends: **mergeable** — the residual is a one-word comment nit, and Improvement 3 (CI coverage for the new target) plus Suggestion 4 (cross minimum-version note) remain fine as follow-ups.

kxxt:
>     * **Improvement 1 (stale messages)** — the `get_arch_config()` panic now says "x86_64, aarch64 and riscv64". **One residual:** the comment above the cross-include triplet match still reads "restricts supported targets to these two" over what is now a three-entry list — comment-only, non-blocking; one word if the branch gets touched again.

Fixed.

>     * **Suggestion 1 (validation symmetry)** — `get_riscv_pthread_descriptor` now applies `IS_VALID_USER_SPACE_ADDRESS` to `clear_child_tid`, matching the specific1stblock path. (The trailing NULL ternary is redundant behind the new check, but harmless.)

Good catch. I forgot that `IS_VALID_USER_SPACE_ADDRESS` does NULL check. Fixed now.

josefbacik:
**92192e0f → c6ee0d1b delta-read** (both ends pinned, same in-place two-commit rewrite): exactly the two remaining notes folded — the build.rs comment no longer counts arches, and the redundant NULL ternary behind the new validity check is gone. Nothing else changed. No residuals remain; the verdict extends clean: **mergeable**.

josefbacik:
Fix the rustfmt and then I'll merge. Thanks!

kxxt:
> Fix the rustfmt and then I'll merge. Thanks!

Fixed. Sorry I forgot that.

kxxt:
BTW I think some bpf code is reading user memory with `bpf_probe_read_kernel`. Perhaps those should be replaced with `bpf_probe_read_user`, though read_kernel appears to work in practice

e.g. https://github.com/josefbacik/systing/blob/main/src/pystacks/bpf/pystacks.bpf.c#L976

josefbacik:
> BTW I think some bpf code is reading user memory with `bpf_probe_read_kernel`. Perhaps those should be replaced with `bpf_probe_read_user`, though read_kernel appears to work in practice

Ah yeah that's fine in practice but should be cleaned up.

PR #163, opened 2026-07-17
Fix comm attribution for loader-launched processes and renamed threads

josefbacik:
Two related attribution fixes for process/thread naming.

## Problem

1. Processes launched by invoking the ELF dynamic loader directly
   (`ld-linux-x86-64.so.2 [options] /path/to/prog ...` — the standard
   pattern in hermetic build/CI environments) keep the loader as
   `/proc/PID/exe` for their whole lifetime, so the exe-preferred
   process naming labels every such process `ld-linux-x86-64.so.2`. On
   fork/exec-heavy CI hosts this collapses the top process bucket into
   the loader's name — python/compiler/worker samples all render under
   one meaningless label.

2. Task records were created on first sighting and never touched again:
   every event consumer deduplicated by bare tgidpid, so the fresh comm
   that already arrives with every sched event was discarded, and a
   task that renamed itself after first sighting (exec,
   `pthread_setname_np`) stayed frozen at its first-seen name forever.

## Fix

**Commit 1 — loader-aware process naming:** detect loader basenames
(glibc, musl, and versioned forms) in `record_new_process` and recover
the real program's basename from the already-captured cmdline. The
parser knows the glibc rtld option table (value-taking vs standalone),
honors `--argv0` overrides, and falls back to the exe basename whenever
the invocation isn't parseable (unknown option, missing value, no
program argument) rather than guessing.

**Commit 2 — comm freshness:** replace the per-consumer dedup HashSet
with a `TaskSightings` tracker that forwards a task_info on first
sighting and again whenever its comm changes to a new non-empty value
(zeroed comms are ignored — smaller BPF events leave `task_info` fields
uninitialized). `maybe_record_task` renames the thread record on such a
re-forward, with a truncation guard so a 15-byte kernel comm never
downgrades a longer stored name it is merely a prefix of. Process names
are deliberately NOT comm-renamed: they are exe-derived by preference,
and renaming them from thread comms would reintroduce the
child-name-leak problem the exe preference exists to avoid.

## Cost

No BPF changes — the comm already rides in every event. Per-event
userspace cost is unchanged in the common case (one hash lookup as
before, plus a 16-byte compare on hits); memory grows by the stored
16-byte comm per seen task per consumer. Rename traffic to the
discovery thread only occurs on actual comm changes.

## Known limitation

A thread record holds one name, so a task that execs mid-trace gets its
whole timeline attributed to the post-exec name (previously: the
pre-exec name). For the exec-chain populations this fix targets
(launcher → real program), post-exec is where the samples are, so this
is strictly better; fully correct exec-epoch attribution would need
time-sliced thread records, which is out of scope here.

## Validation

- 420/420 lib tests; new unit coverage for the loader parser (option
  table, `--argv0`, unknown-option fallback), rename semantics
  (truncation guard, zero-comm, process-name fence), and TaskSightings
  (first-sight/change/zero-comm sequences).
- Full vng integration gate green on this branch (per-target results in
  the results comment below).
- `clippy -D warnings` and `fmt` clean.
- Post-deploy the fix is directly observable in profiling views: on
  fork/exec-heavy hosts the loader-name process bucket should dissolve
  into real program names, and thread names should track
  `pthread_setname_np` renames instead of first-sighting snapshots.

### DIFF
--- a/src/session_recorder.rs
+++ b/src/session_recorder.rs
@@ -1056,28 +1056,17 @@ impl SessionRecorder {
             .insert(info.tgidpid, thread_descriptor);
     }
 
-    /// Update a known thread's name from a freshly observed BPF comm.
+    /// Apply a rename decided under the read lock in maybe_record_task.
     ///
-    /// Process names are deliberately not updated here: they are exe-derived
-    /// by preference (see record_new_process), and overwriting them from a
-    /// thread's comm would reintroduce the child-name-leak problem that
-    /// exe-preference exists to avoid.
-    fn maybe_rename_thread(&self, info: &task_info) {
-        let fresh = Self::extract_comm(info);
-        if fresh.is_empty() {
-            return;
-        }
-        {
-            let threads = self.threads.read().unwrap();
-            match threads.get(&info.tgidpid) {
-                Some(existing) if Self::rename_improves(existing.thread_name(), &fresh) => {}
-                _ => return,
-            }
-        }
+    /// Process names are deliberately not updated on comm changes: they are
+    /// exe-derived by preference (see record_new_process), and overwriting
+    /// them from a thread's comm would reintroduce the child-name-leak
+    /// problem that exe-preference exists to avoid.
+    fn rename_thread(&self, tgidpid: u64, fresh: String) {
         let mut threads = self.threads.write().unwrap();
-        if let Some(descriptor) = threads.get_mut(&info.tgidpid) {
+        if let Some(descriptor) = threads.get_mut(&tgidpid) {
             // Re-check under the write lock: another consumer may have
-            // renamed concurrently between our read and write.
+            // renamed concurrently since the read-lock decision.
             if Self::rename_improves(descriptor.thread_name(), &fresh) {
                 descriptor.set_thread_name(fresh);
             }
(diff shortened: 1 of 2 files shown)


josefbacik:
## rust-bpf-code-reviewer review

Mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over the full branch diff (main...HEAD, 2 commits) by the authoring session; all five passes (structure, Rust quality, BPF quality, schema/versioning, integration).

### Summary

Two tightly-scoped userspace attribution fixes (loader-aware process naming; comm freshness for thread records). No BPF, schema, or version-line changes. Test coverage is meaningful for the edges that matter.

### Critical Issues

None found. Checked explicitly: `Cargo.toml`/`Cargo.lock` untouched (bump-on-merge regime); no table/schema changes, so `SCHEMA_VERSION` correctly untouched and no `[bump minor]` needed; no new recorder; thread identity (utid) semantics unaffected — renames change descriptor values, not join keys.

### Improvements

- Exec-boundary semantics: a single thread record spans an exec, so a mid-trace exec now attributes the whole timeline to the post-exec name (the inverse of the old failure, and the better inverse for launcher→program chains). Accepted tradeoff, disclosed in the PR body; time-sliced thread records would be the complete fix and are out of scope here.

### Suggestions

- `Path::new(prog.as_str())` could be `Path::new(prog)`; kept for explicitness.
- `HashMap<u64, [u8; 16]>` hardcodes the kernel comm size; drift is compile-checked via the `info.comm` assignment, so a named constant would be cosmetic.

### Positive Notes

- The truncation guard in `rename_improves` closes the subtle downgrade path (a 15-byte kernel comm replacing the longer exe-derived name it is a prefix of) and is unit-tested in both directions.
- `loader_target_name`'s None-means-fallback contract keeps every unparseable invocation on today's behavior — the change can only improve names, never invent them.

josefbacik:
## Validation results — vng integration gate

Full gate on the exact PR head (6813ef7), one clean VM boot per target
(virtme-ng + QEMU TCG, kernel 6.12, 4 cpus / 8G):

| target | result | tests | wall |
|---|---|---|---|
| trace_validation | PASS | 32 | 2006s |
| memory_recorder | PASS | 5 | 591s |
| stream_unix | PASS | 1 | 67s |
| analyze_commands | PASS | 1 | 831s |
| perf_counter | PASS | 1 | 1s |
| network_throughput | PASS | 1 | 54s |
| gvisor_record | PASS | 1 | 656s |
| gopclntab_record | PASS | 1 | 188s |
| missed_events | PASS | 1 | 57s |

44 tests, 0 failures, no target vacuous (every log carries a non-zero
libtest summary and the in-guest exit marker). Host-side: 420/420 lib
tests, `clippy -D warnings` clean, `fmt` clean.

PR #164, opened 2026-07-17
memory: threshold-batch rss_stat events + tp_btf attach path

josefbacik:
# memory: threshold-batch rss_stat events + tp_btf attach path

## Summary

Two layered fixes for the memory recorder's `rss_stat` hook overhead on
fault-heavy hosts:

**Layer 1 — attach path.** Add `SEC("tp_btf/rss_stat")` as the preferred
attach, keeping the existing `SEC("tracepoint/kmem/rss_stat")` as a
load-time fallback (selected by probing vmlinux BTF for
`btf_trace_rss_stat`; forceable via the hidden
`--memory-rss-force-classic` for test coverage). The classic path routes
every tracepoint firing through the kernel's `perf_trace_rss_stat` glue,
whose `TP_fast_assign` on ≥6.2 kernels calls
`percpu_counter_sum_positive()` — an O(nr_cpus) cache-line walk per event.
`tp_btf` bypasses the glue and reads `mm->rss_stat[member]` directly via
CO-RE (dual-layout: `percpu_counter[4]` on ≥6.2, `struct mm_rss_stat` on
<6.2). The tp_btf read is a cheap `percpu_counter_read()` approximation;
see "Accuracy" below.

**Layer 2 — threshold batching.** The kernel fires `rss_stat` on every
per-page counter update, so a fault storm produces one event per 4 KiB page.
Both attach paths now feed a shared `handle_rss_stat()` helper that emits
only when the absolute per-(tgid, member) counter drifts
≥ `memory_rss_threshold_bytes` (default max(16 MiB, 64·nr_cpus·page_size),
overridable via `--memory-rss-threshold-bytes`; `0` restores the
pre-change emit-every-event behavior). Sub-threshold residuals are flushed at
last-thread-exit by a dedicated `tp_btf/sched_process_exit` program
(`systing_rss_stat_exit_flush` — separate from the sched recorder's own
exit hook so each recorder's map references stay independent). On map-full
the gate falls through to direct-submit, so no sample is ever dropped — only
the batching optimization is lost for overflow tgids. `memory_rss.size`
stays an absolute byte-valued counter sample; only emission cadence changes,
so any downstream per-window `max(size)` aggregate is unaffected within
±threshold.

## Motivation

On fault-heavy hosts (e.g. many-core CI nodes running allocator-intensive
jobs), `perf_trace_rss_stat` has been observed at ~1% of on-CPU samples
during a memory-recorder capture window — effectively all profiler
self-observation. tp_btf skips the per-event O(nr_cpus) sum; threshold
batching cuts the remaining per-event fixed cost from once-per-page to
once-per-N-MiB.

## Accuracy

On the tp_btf path (≥6.2 kernels), `memory_rss.size` is the approximate
`percpu_counter_read()` of `mm->rss_stat[member]` — it misses per-CPU
batch slots. Worst-case error is ±(percpu_counter_batch · nr_online_cpus)
pages, where the kernel sets batch = max(32, 2·nr_online_cpus), i.e.
quadratic in core count. In practice only CPUs the mm is actively faulting
on hold drift, and typical error is a small multiple of batch. The classic
fallback still emits the kernel's own exact sum. An exact-sum tp_btf emit
via `bpf_per_cpu_ptr` over `rss_stat[].counters` is not portably
verifiable (the helper needs `ARG_PTR_TO_PERCPU_BTF_ID`, which a
`BPF_CORE_READ`-obtained pointer is not; direct BTF-walk access preserves
percpu-ness only on clang-built kernels with `btf_type_tag("percpu")` in
vmlinux BTF). Documented in `MemoryRssRecord` and the BPF comments.

## Notes

- The aggregation map is keyed by **tgid**, not tgidpid: `rss_stat`
  counters are per-mm (process-wide), so a per-thread key would redundantly
  track the same value from every thread. Concurrent per-tgid updates are
  unsynchronized by design (worst case: one duplicate emit).
- No schema change; `memory_rss.size` semantics unchanged.

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -4914,26 +4914,32 @@ int BPF_PROG(systing_rss_stat_exit_flush, struct task_struct *task)
 {
 	if (tool_config.memory_rss_threshold_bytes == 0)
 		return 0;
-	/* Delete even for untraced tasks (no map-entry leak if trace_task's
-	 * answer changed between insert and exit); emit only for traced. */
-	u32 tgid = BPF_CORE_READ(task, tgid);
+	/* Filter first: this hook fires for every task exit on the host, so
+	 * untraced exits pay only the tool check and this. Trade: an entry
+	 * inserted while its task was traced is no longer deleted if the
+	 * task's filter answer changed before exit (e.g. cgroup migration in
+	 * cgroup-targeted mode) — bounded by the map's fall-through-on-full
+	 * semantics (batching degrades to direct-submit, correctness
+	 * unaffected) and impossible in whole-system mode where the filter
+	 * answer is constant. */
+	if (!trace_task(task))
+		return 0;
 	if (BPF_CORE_READ(task, signal, live.counter) != 0)
 		return 0;
+	u32 tgid = BPF_CORE_READ(task, tgid);
 	struct rss_stat_state *st = bpf_map_lookup_elem(&rss_stat_last, &tgid);
 	if (!st)
 		return 0;
 
-	if (trace_task(task)) {
-		/* task->mm is NULL here (exit_mm() precedes this tracepoint),
-		 * so flush latest_seen[] captured on the rss_stat path rather
-		 * than re-reading mm. */
-		for (u32 m = 0; m < NR_MM_COUNTERS; m++) {
-			s64 latest = st->latest_seen[m];
-			if (latest < 0)
-				continue; /* never seen */
-			if (latest != st->last_emitted[m])
-				emit_rss_stat_event(task, m, latest);
-		}
+	/* task->mm is NULL here (exit_mm() precedes this tracepoint), so
+	 * flush latest_seen[] captured on the rss_stat path rather than
+	 * re-reading mm. */
+	for (u32 m = 0; m < NR_MM_COUNTERS; m++) {
+		s64 latest = st->latest_seen[m];
+		if (latest < 0)
+			continue; /* never seen */
+		if (latest != st->last_emitted[m])
+			emit_rss_stat_event(task, m, latest);
 	}
 	bpf_map_delete_elem(&rss_stat_last, &tgid);
 	return 0;
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -4825,12 +4825,13 @@ static __always_inline int emit_rss_stat_event(struct task_struct *task,
 	return 0;
 }
 
-/* Threshold gate shared by both attach paths. */
+/* Threshold gate shared by both attach paths. Callers have already
+ * gated with trace_task() — the filter runs before any per-event work
+ * (counter reads, size math) so untraced tasks pay as close to zero as
+ * possible on this very hot tracepoint. */
 static __always_inline int handle_rss_stat(struct task_struct *task,
 					   u32 member, s64 size_bytes)
 {
-	if (!trace_task(task))
-		return 0;
 	if (member >= NR_MM_COUNTERS)
 		return 0;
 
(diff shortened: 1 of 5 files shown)


josefbacik:

Review executed per this repo's `.claude/agents/rust-bpf-code-reviewer.md` definition, over the full diff at `5638255b`.

**Summary:** Two-layer change (tp_btf attach preference + threshold-batched emission) is well-structured: the hot path stays minimal (one map lookup per event), attach selection is probed-and-fallback rather than assumed, and the kernel-version considerations are documented exactly where they're load-bearing. Verdict: no critical issues.

**Critical issues:** none. Gate checks: no schema change (no `SCHEMA_VERSION` bump required, no `SCHEMA_CHANGES.md` entry needed); `Cargo.toml`/`Cargo.lock` version untouched (bump-on-merge respected); no new tables or raw `tid`/`pid` attribution introduced.

- `systing_rss_stat_btf` performs the CO-RE counter read before `handle_rss_stat`'s `trace_task` filter, so targeted-trace mode spends one read per untraced event. Hoisting the filter would duplicate it across both attach shims; the current shape is a reasonable trade — noted, not requested.

**Positive notes:**
- `BPF_NOEXIST` on the state-map insert keeps a race-losing thread from clobbering the winner's populated state — subtle and correct.
- The exit-flush living in its own `tp_btf/sched_process_exit` program keeps `memory_ringbufs` references confined to memory-recorder programs, so default config (memory off) still loads — and the integration suite pins exactly that regression.
- Honest error-bound documentation on the approximate percpu read (quadratic worst case, spurious-emit consequence) in both the BPF comment and `MemoryRssRecord`.
- The integration tests bind hard: threshold-gate breakage shows as a three-orders-of-magnitude event-count change, and the byte-vs-page guard fails at 4096×.

Gates run alongside the review: `cargo test --lib` 406/406; clippy clean; `memory_recorder` integration target 7/7 in a v6.12 BPF VM (both attach paths, including the forced-classic fallback); default-config e2e regression pass.

josefbacik:
Implemented per maintainer review: `trace_task` now runs first in both rss_stat attach shims — before the CO-RE counter read on the tp_btf path and before the size handling on the classic path — so untraced tasks exit after the filter alone. `handle_rss_stat` documents the caller-gates contract instead of re-filtering.

Also audited the rest of the memory handler family for the same ordering: the mmap/munmap/brk, fault, and allocator handlers already filter first; the exit-flush keeps its documented delete-even-untraced map cleanup (preventing entry leaks when a task's filter answer changed between insert and exit), with `trace_task` still gating only its emit loop.

Re-verified: `cargo test --lib` 406/406; `memory_recorder` integration target 7/7 in the v6.12 BPF VM (both attach paths, including forced-classic).

src/bpf/systing_system.bpf.c
```diff
@@ -4719,14 +4818,127 @@ int systing_rss_stat(struct trace_event_raw_rss_stat *ctx)
...
+	/* Delete even for untraced tasks (no map-entry leak if trace_task's
+	 * answer changed between insert and exit); emit only for traced. */
+	u32 tgid = BPF_CORE_READ(task, tgid);
+	if (BPF_CORE_READ(task, signal, live.counter) != 0)
+		return 0;
+	struct rss_stat_state *st = bpf_map_lookup_elem(&rss_stat_last, &tgid);
+	if (!st)
+		return 0;
+
+	if (trace_task(task)) {
```
josefbacik: This needs to move higher up, it should be the first check after the tool check.

josefbacik: Moved — `trace_task` is now the first check after the tool check in the exit-flush; the last-thread test, tgid read, and map lookup all sit behind it (and the last-thread test now precedes the tgid read, so non-last exiting threads return one read sooner too).

The trade this makes is documented at the site: an entry inserted while its task was traced is no longer deleted if the task's filter answer changed before exit (cgroup migration in cgroup-targeted mode). It's bounded by the map's fall-through-on-full semantics — batching degrades to direct-submit, correctness unaffected — and can't occur in whole-system mode where the filter answer is constant.

Re-verified: `cargo test --lib` 406/406; `memory_recorder` integration target 7/7 in the v6.12 BPF VM (both attach paths).

PR #165, opened 2026-07-18
Symbolize processes that exit before end-of-trace symbolization

josefbacik:
Stacks are symbolized at end of trace, and processes that die before
then render every user frame as `unknown ([exited])`: blazesym needs
`/proc/<pid>/maps`, which is gone by then. On fork/exec-heavy hosts
(build and CI machines) the `[exited]` class can be the single largest
leaf in the profile — short-lived compilers, shells, and test children
dominate exactly the workloads being profiled.

## Fix: snapshot at first sample, on a dedicated fast lane

The first stack sample for a tgid proves the process is on-CPU and
alive — that is the earliest (and often only) chance to capture its
mappings. The new `exit_snapshot` module captures the process's
executable file-backed mappings and pins each backing file with a
read-only descriptor, deduplicated by (device, inode). At end-of-trace
symbolization, a dead tgid with a snapshot resolves user addresses to
(pinned file, file offset) and symbolizes through `/proc/self/fd/<n>`
— the same `Source::Elf` + `Input::FileOffset` path the sandbox island
bridging uses — with the frame's module reported as the original
binary basename. Addresses that miss the snapshot keep the `[exited]`
rendering.

The capture cannot ride the stack-event consumer: that queue runs
seconds behind under load, and a short-lived process is gone before
its first sample is consumed from it — the busier the host, the deeper
the queue, and the motivating workloads are the busy ones. BPF
therefore emits a tiny sidecar hint (the tgid) per user-stack sample
into its own ringbuf, drained by a dedicated thread that does nothing
but observe tgids and capture on first sight, so capture latency is
independent of the main queue's depth. Hints are advisory: a dropped
hint costs capture latency (the stack consumer makes the same
observation when it gets there), never correctness.

### Exec invalidates snapshots

A process sampled before exec and again after it would otherwise
resolve post-exec addresses against the pre-exec image — confidently
wrong names, worse than an honest `[exited]` — and the
one-attempt-per-tgid rule would block the correct re-capture. The
existing `sched_process_exec` program already notifies userspace for
pystacks re-discovery; it now runs (and its ringbuf is drained)
whenever either pystacks or exited recovery is enabled — the pystacks
config clearing stays pystacks-gated — and the handler drops the
tgid's snapshot before anything that can bail on the event, so even an
exec-then-exit leaves no stale snapshot behind. Either interleaving of
hint and exec event degrades safely: at worst the process re-captures
on its next sample, or its frames keep the `[exited]` rendering.

### Pinning

`/proc/<pid>/map_files/<range>` is the primary path: namespace-
independent, unlink-proof (ephemeral CI binaries are often deleted
before symbolization), and authoritative by construction. Where it is
unavailable (reduced-capability environments), pinning falls back to
the path through `/proc/<pid>/root`, then the plain path — each
verified against the mapping's (device, inode), so a recycled or
namespace-mismatched path can never pin the wrong file.

### Bounded on every axis

Every limit degrades to the existing `[exited]` rendering, never to an
error: one capture attempt per tgid; a per-second capture budget (fork
storms beyond it keep today's behavior for the overflow); a per-process
segment cap; a global pinned-descriptor cap (512, below common
RLIMIT_NOFILE soft limits) with oldest-first eviction; a 64KB hint
ringbuf whose overflow only defers capture to the stack consumer; an
8192-entry seen-tgid LRU whose eviction of a live tgid merely re-hints
once (the consumer's one-attempt rule absorbs the duplicate). All
snapshots and descriptors are released at the end of each
symbolization pass, so continuous mode starts every tick fresh and
never accumulates descriptors across ticks.

### Hot-path cost, measured

Enabling recovery costs where the feature delivers — at end-of-trace
symbolization — not on the hot path. Instrumented runs on an
allocation-heavy integration test (per-program `bpf_stats`
run_time/run_cnt, per-thread CPU, and `/proc/stat` phase timelines
sampled inside the test VM; receipts in the results comment) show:

- The recording phase and total BPF program time are at parity with
  recovery disabled. Four hot-path gating variants (per-sample
  emission, LRU-gated, LRU + per-cpu cache, plain-hash gate) landed in
  one scattered wall-clock band with no ordering by mechanism — the
  per-sample hot path is not a measurable term on this rig.
- The entire enable-delta sits in the post-recording, single-threaded
  end-of-trace phase: resolving the dead process's stacks through its
  snapshot does the ELF parsing and symbol lookup that the `[exited]`
  rendering never paid. The run-to-run scatter tracks how much each
  run's captures actually resolve.

That makes the cost a trade, not a tax: the delta scales with the
number of processes actually recovered — it is the price of the names
delivered, paid only where the profile would otherwise read
`[exited]` — and it is bounded by the existing caps (one attempt per
tgid, the per-second capture budget, the segment cap, the
512-descriptor cap). In continuous mode it rides the per-tick
symbolization pass that already exists and releases everything at
tick end. The emulated rig amplifies constants 10-20x — the
transferable result is the phase attribution, not the absolute
seconds.

The hint lane itself stays out of the per-sample picture by design:
emission is LRU-gated to once per tgid lifetime (wasted ringbuf
traffic and consumer wakeups are never generated), the mark is
written only after a successful submit so a failed reserve retries on
the next sample, racing duplicates are absorbed by the consumer's
one-attempt rule, and exec deletes the entry so a replaced address
space re-hints even if its exec event was dropped. Capture is one
maps read plus a handful of opens, once per new process,
budget-capped. When disabled, the hint ringbuf and the LRU shrink to
minimum size and the BPF hint path is compiled out by the rodata
gate.

`--no-exited-recovery` reverts to the previous behavior.

### Known limitation: tgid recycling

Snapshots are keyed by tgid. A tgid recycled within one trace (or one
continuous-mode tick) after its original owner died — and whose new
owner also dies before symbolization — can resolve against the old
owner's snapshot; under ASLR that yields wrong names rather than an
honest `[exited]`. The window needs pid-space wrap-around while a
trace is open, which is realistic only on fork-storm hosts with the
default `pid_max` of 32768 (~30s at 1000 forks/s). A follow-up can
close it by keying snapshots by (tgid, start_time) and sealing on the
exit event that userspace already receives; kept out of this PR to
hold the review surface down.

## Validation

- New integration test with its negative control run first: the burner
  (the test binary re-invoking itself, full symbol table) runs as
  systing's traced command, so it execs only after tracing is attached
  and the recording stops because the burner exited and was reaped —
  sampled-while-alive and gone-by-symbolization hold by construction at
  any machine speed, not by timing margins. With recovery disabled the
  dead burner must produce `[exited]` frames and zero named frames;
  with recovery enabled the same burner must resolve by name. The test
  fails against the consumer-side capture alone and passes with the
  hint lane, so it guards the latency property, not just the happy
  path.
- 8 unit tests on the snapshot module: maps-line parsing (executable
  file-backed only, deleted-marker handling), budget and seen-tgid
  behavior, resolution arithmetic against the test's own live process,
  reset semantics, eviction degradation, exec invalidation and
  re-capture.
- 428/428 lib tests; `clippy -D warnings` and `fmt` clean.
- Full vng integration gate green on this branch (per-target results in
  the comment below).

## Expected effect

On a busy CI host we measured the `[exited]` leaf as the single largest
attribution class; after this change it should shrink to the residue of
processes whose entire life fits inside the hint drain's wakeup latency
plus over-budget fork-storm overflow. The before/after is directly
observable in any profile view of a fork/exec-heavy host.

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -81,6 +81,10 @@ const volatile struct {
 				* gate is belt-and-braces (and lets the verifier
 				* prune the bodies if the programs ever get loaded
 				* for another reason). */
+	u32 collect_exited_recovery; /* First-sample mapping snapshots for
+				      * exited-process symbolization: gates the
+				      * snapshot-hint sidecar ringbuf and the
+				      * exec-invalidation path. */
 } tool_config = {};
 
 enum event_type {
@@ -396,12 +400,13 @@ struct memory_event {
  * Dummy instance to get skeleton to generate definition for
  * `struct task_event`
  */
-#ifdef SYSTING_PYSTACKS
 /*
- * Notify userspace that a traced process exec'd or forked so it can update
- * the pystacks discovery state. The handler in handle_exec_events() reads
- * /proc/<pid>/{exe,maps}, so this only carries the pid and a flag describing
- * what kernel event produced it (exec vs. fork) for logging/diagnostics.
+ * Notify userspace that a traced process exec'd or forked. Two consumers:
+ * pystacks discovery state updates, and exited-recovery mapping-snapshot
+ * invalidation (exec replaces the address space, so a snapshot captured
+ * for the old image must be dropped). The handler in handle_exec_events()
+ * reads /proc/<pid>/{exe,maps}, so this only carries the pid and a flag
+ * describing what kernel event produced it (exec vs. fork).
  */
 struct exec_event {
 	u32 pid;
@@ -412,7 +417,22 @@ struct exec_event {
 	 * memory to userspace. */
 	u32 is_fork;
 };
-#endif
+
+/*
+ * Sidecar notification that a stack sample with a user stack was taken for
+ * a tgid. A dedicated userspace thread drains these and captures the
+ * process's executable mappings on first sight: the main stack ringbuf can
+ * queue seconds behind under load, and a short-lived process is gone
+ * before its first sample is consumed from it — a snapshot captured then
+ * comes too late. Hints are advisory; if one is dropped the stack-event
+ * consumer makes the same observation later, just slower.
+ */
+struct snapshot_hint {
+	u32 tgid;
+	/* Explicit, zeroed padding: bpf_ringbuf_reserve() does not zero its
+	 * allocation, and implicit padding would leak kernel memory. */
+	u32 reserved;
+};
 
 struct task_event _event = {0};
 struct stack_event _stack_event = {0};
@@ -428,9 +448,8 @@ struct memory_event_header _memory_event_header = {0};
 enum memory_event_type _memory_event_type = MEMORY_RSS_STAT;
 enum memory_rss_member _memory_rss_member = MEMORY_MM_FILEPAGES;
 enum memory_alloc_op _memory_alloc_op = MEMORY_OP_MALLOC;
-#ifdef SYSTING_PYSTACKS
 struct exec_event _exec_event = {0};
-#endif
+struct snapshot_hint _snapshot_hint = {0};
 struct arg_desc _arg_desc = {0};
 enum event_type _type = SCHED_SWITCH;
 enum arg_type _arg_type = ARG_NONE;
@@ -739,7 +758,6 @@ struct memory_ringbuf_map {
 	ringbuf_memory_events_node6 SEC(".maps"),
 	ringbuf_memory_events_node7 SEC(".maps");
 
-#ifdef SYSTING_PYSTACKS
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
 	/*
@@ -752,7 +770,36 @@ struct {
 	 */
 	__uint(max_entries, 16384);
 } ringbuf_exec_events SEC(".maps");
-#endif
+
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	/*
+	 * 64KB of ~16-byte snapshot hints. Overflow is harmless by design:
+	 * the stack-event consumer performs the same first-sample
+	 * observation when it gets there — a dropped hint costs capture
+	 * latency, never correctness.
+	 */
+	__uint(max_entries, 65536);
+} ringbuf_snapshot_hints SEC(".maps");
+
+/*
+ * Once-per-tgid gate for snapshot hints. The consumer only acts on the
+ * first hint it sees for a tgid (one capture attempt per tgid), so
+ * emitting a hint per user-stack sample is pure waste past the first:
+ * gating emission here keeps the hint ringbuf and its consumer thread
+ * out of the per-sample picture entirely. Sized for 8192
+ * concurrently-live tgids before benign eviction: an
+ * evicted-then-resampled tgid re-hints, and the consumer's one-attempt
+ * rule absorbs the duplicate — the same advisory semantics as a
+ * dropped hint, in the other direction. Exec deletes the entry (see
+ * systing_sched_process_exec) so a replaced address space re-hints.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__uint(max_entries, 8192);
+	__type(key, u32);
+	__type(value, u8);
+} recovery_hinted_tgids SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_RINGBUF);
@@ -1611,7 +1658,41 @@ static void emit_stack_event_with_ts(void *ctx, struct task_struct *task,
 		event->kernel_stack_length = len / sizeof(u64);
 	else
 		event->kernel_stack_length = 0;
+
+	u32 user_stack_length = event->user_stack_length;
 	bpf_ringbuf_submit(event, flags);
+
+	/*
+	 * Exited-recovery snapshot hint, after the main submit so the stack
+	 * event's latency is unaffected. Only samples with a user stack need
+	 * a mapping snapshot; kernel-only samples symbolize without one, and
+	 * only the FIRST hint per tgid does any work downstream — the LRU
+	 * lookup keeps the hot path to one hash probe per sample.
+	 */
+	if (tool_config.collect_exited_recovery && user_stack_length > 0) {
+		u32 tgid = task->tgid;
+
+		if (!bpf_map_lookup_elem(&recovery_hinted_tgids, &tgid)) {
+			struct snapshot_hint *hint = bpf_ringbuf_reserve(
+				&ringbuf_snapshot_hints, sizeof(*hint), 0);
+			if (hint) {
+				u8 seen = 1;
+
+				hint->tgid = tgid;
+				hint->reserved = 0;
+				bpf_ringbuf_submit(hint, 0);
+				/*
+				 * Mark only after a successful emit: a failed
+				 * reserve retries on the next sample, keeping
+				 * hints advisory. Two CPUs racing the lookup
+				 * emit a duplicate, which the consumer's
+				 * one-attempt rule absorbs.
+				 */
+				bpf_map_update_elem(&recovery_hinted_tgids,
+						    &tgid, &seen, BPF_ANY);
+			}
+		}
+	}
 }
 
 static void emit_stack_event(void *ctx, struct task_struct *task,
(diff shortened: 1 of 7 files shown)


josefbacik:
Full vng integration gate at this head (emulated TCG VM, 4 cpus / 8G, per-target timeout 3600s), all targets:

```
trace_validation       PASS      2258s  test result: ok. 33 passed; 0 failed; 0 ignored; 0 measured; 8 filtered out; finished in 2219.90s
memory_recorder        PASS      1482s  test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 1443.30s
stream_unix            PASS       122s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 83.04s
analyze_commands       PASS       907s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 868.32s
perf_counter           PASS        37s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.54s
network_throughput     PASS        90s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 52.64s
gvisor_record          PASS       736s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 697.71s
gopclntab_record       PASS       254s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 215.23s
missed_events          PASS        91s  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 53.29s
```

428/428 lib tests, `clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean at this commit.

## Cost attribution for the recovery feature

All runs: `test_memory_alloc_e2e` isolated (one long-lived traced child, `memory_alloc_sample_rate: 1`), same rig, same day, libtest wall seconds.

**Recovery off — two variants, statistically identical:**
- fully off: 429.98s
- BPF hint side compiled out (rodata gate) with ALL userspace recovery machinery live — consumer-side per-event observe, capture path, threads: 426.10s

The second variant prices the userspace side of the feature at zero.

**Recovery on — four hot-path gating variants, one scattered band, no ordering by mechanism:**
- per-sample emission (no gate): 702.87s / 697.69s
- LRU-gated (shipping config): 596.55s
- LRU + per-cpu global-data cache: 661.05s
- plain-hash gate: 657.35s

If the per-sample hot path were a measurable term, these four would order by mechanism cost. They don't — the band is run-to-run scatter.

**Phase attribution (instrumented leg: `kernel.bpf_stats_enabled` per-program run_time/run_cnt + per-thread CPU + `/proc/stat`, sampled every 60s inside the VM):**
- Recording phase ends at ~+150s in ON and OFF configs alike; cumulative run_time across ALL of systing's BPF programs measures in single-digit seconds — three orders below the enable-delta.
- From recording end to test end, the VM sits ~75-80% idle with exactly one thread of the test process busy: the end-of-trace phase. The entire ON-delta lives there — snapshot-based symbolization of the dead child's stacks (ELF parse + symbol lookup through the pinned descriptors) that the `[exited]` rendering never pays. Within-phase split between symbol resolution and output writing was not separately instrumented; the phase-level placement is the claim.
- The ON-band scatter tracks how much each run's captures resolve, which is race-sensitive (capture timing vs. child startup) — consistent with OFF-class reproducing to ±0.5% while ON-class scatters ±9%.

TCG amplifies constants 10-20x. The transferable result is the phase attribution — the feature's cost is end-of-trace symbolization work proportional to what it recovers, bounded by the caps described in the PR body — not the absolute seconds.

**Direct phase-split pair** (both legs instrumented, same regime, back-to-back): recording ends ~+150s in both; the post-recording phase runs ~115s with recovery off vs ~340s with recovery on — the full 222s pair-delta (496.39s vs 274.67s totals) sits in the end-of-trace phase, recording phases equal. Both instrumented legs also ran ~130-160s faster in absolute terms than the earlier uninstrumented band (host-regime shift across the session) while the ON-vs-OFF delta reproduced — further evidence that the deltas, not the absolute walls, are the meaningful quantity on this rig.

josefbacik:
## rust-bpf-code-reviewer verdict — full diff `70cf334..0722aef` (7 files, +978/−28)

Mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session over the full branch diff (v1.11.24...HEAD, 1 commit, 7 files, +978/−28) by the authoring session; all five passes (structure, Rust quality, BPF quality, schema/versioning, integration).

### Summary
Well-bounded feature with a clear degradation story: every limit (attempt budget, segment cap, descriptor cap, ringbuf overflow, LRU eviction) falls back to the pre-existing `[exited]` rendering rather than erroring or mis-resolving. The BPF hot path is rodata-gated when off; when on, hint emission is LRU-gated to once per tgid so the per-sample lane never generates redundant ringbuf traffic — and instrumented runs (per-program bpf_stats, phase timelines) place the feature's wall cost in end-of-trace symbolization of recovered stacks, not in any per-sample path, with total BPF program time at parity with the feature disabled. No schema, version, or recorder-table changes.

### Critical Issues
None.
- Schema/versioning gates: no `src/duckdb.rs`/`src/trace/schema.rs` changes, `SCHEMA_VERSION` untouched (correctly — no schema change), `Cargo.toml`/`Cargo.lock` version untouched (bump-on-merge workflow owns it), no `[bump minor]` needed.
- Recorder consistency: no new tables or per-event attribution columns; frame rendering only.
- Unsafe: no new `unsafe` blocks (the `Plain` impls follow the existing pattern for BPF-mirrored `repr(C)` structs; `libc::major`/`minor` are safe wrappers).

### Improvements
1. `ExitSnapshots::seen` grows one entry per distinct tgid between `reset()` calls regardless of capture budget (the budget caps *captures*, not *marks*). Bounded in practice by pid-space reuse and the per-tick `reset()` in continuous mode, but a pathological fork storm on a long tick can hold a few tens of MB. Worth a ceiling (e.g. clear-when-larger-than-N) in a follow-up; not blocking — the per-tick reset is the operative bound for the deployed mode.
2. `observe_tgid` is a no-op when disabled, but the stack-consumer fallback still takes the mutex once per stack event to discover that. An `AtomicBool` fast-path (or caching enablement in `StackRecorder`) would make the disabled path lock-free. Minor — and empirically negligible: a control run with the BPF hint side compiled out and this path fully live measured identical to fully-off (426.10s vs 429.98s).
3. Snapshots are keyed by tgid alone, so a tgid recycled within one trace/tick whose new owner also dies can resolve against the old owner's snapshot (wrong names under ASLR rather than `[exited]`). Realistic only on fork-storm hosts with default `pid_max`; documented in the PR body with a concrete follow-up design (key by (tgid, start_time), seal on the exit event userspace already receives). Follow-up material, not blocking.

### Suggestions
- `snapshot_hint.reserved` is explicitly zeroed BPF-side and documented — good; if the struct ever grows, keep the explicit-padding discipline (bpf_ringbuf_reserve does not zero).
- `recovery_hinted_tgids` shrinks to `max_entries(1)` when recovery is off, following the `pystacks_symbols` precedent for always-referenced maps; the integration test's negative-control leg exercises exactly this load path, which is the right empirical guard for LRU-map minimum-size behavior across kernels.

### Positive Notes
- The emit-then-mark ordering in the BPF hint path (mark only after successful `bpf_ringbuf_submit`) preserves the advisory-hint property under reserve failure, and the racing-duplicate case is explicitly reasoned to the consumer's one-attempt rule.
- The exec-path LRU delete is placed before the exec-event reserve with the two-doors rationale in a comment — a dropped exec event cannot permanently wedge a tgid out of re-capture.
- `open_mapping_file`'s fallback chain verifies `(device, inode)` on both non-authoritative paths, so a recycled path or namespace mismatch cannot pin the wrong file; `map_files` needs no verification and the comment says why.
- The integration test pins its lifecycle to events (child blocks until tracing attaches; recording stops on reap), making it machine-speed independent — and its negative control runs first, so the test cannot vacuously pass on a broken positive path.
- Lock discipline in `symbolize_stack_frames`: one acquisition per dead-process stack, not per frame, with the contention window documented.
- Cost claims in comments and the commit message are stated as attribution (end-of-trace work proportional to recovery, bounded by the caps), not as unverified hot-path numbers — the map comment carries the mechanical rationale for the LRU gate without wall-clock claims.

Verdict: **PASS** — no Critical issues; the three Improvements are follow-up material, not blockers.

PR #166, opened 2026-07-20
Add --symbolize-elide-generics to bound pathological symbol names

josefbacik:
## Problem

Demangled Rust and C++ names carry the full monomorphized argument list. Deeply nested iterator/combinator stacks and expression templates legitimately demangle to multi-kilobyte strings — observed single instantiations up to ~28 KiB (Eigen contraction kernels) and hot 1-2 KiB Rust iterator-adapter names. Stacks are materialized as arrays of frame strings, so a handful of hot kilobyte-scale names repeated across millions of recorded frames comes to dominate recorder memory and the size of the resulting trace, while carrying no more grouping identity than their first and last path segments.

## Change

New opt-in `--symbolize-elide-generics` flag, default **off** — unflagged runs keep full symbolization, byte-identical output.

When set, names longer than 256 bytes (`ELIDE_GATE`) have their balanced `<...>` and `(...)` argument groups collapsed, preserving the path head and the function segment:

| input shape | rendered |
|---|---|
| `rayon_core::registry::in_worker::<deeply nested adapter types>` | `rayon_core::registry::in_worker::<...>` |
| `rayon_core::join::join_context::<...>::{closure#0}` | closure tail preserved |
| `<core::iter::adapters::map::Map<...> as core::iter::traits::iterator::Iterator>::next` | qualified-self type + trait heads preserved |
| `void Eigen::TensorEvaluator<...>::evalProductSequential(...)` | C++ templates and parameter lists |

Details:

- Names of 256 bytes or less pass through byte-verbatim (`Cow::Borrowed`, zero allocation) — kernel symbols and ordinary frames are untouched.
- `operator<`, `operator<<`, `operator<=>` and `operator()` never open a group; `->` inside function-pointer types never closes one; `{closure#N}` spellings are preserved.
- Rust qualified-self (`<Type as Trait>::method`) keeps the type and trait heads and collapses one level deeper.
- Names whose brackets do not balance, or that remain over-long after elision, fall back to a bounded middle cut with an 8-hex FNV-1a hash of the full original name spliced in, so distinct originals stay distinct and no input produces an unbounded result. FNV rather than `DefaultHasher` keeps the result identical across processes and hosts — the shortened spelling is a grouping key downstream.

Grouping properties (documented on the module): deterministic (same full name always shortens to the same string), prefix-preserving up to the first collapsed group, and instantiation-merging — monomorphizations of one source function aggregate under that function; two different source functions can never merge.

Composes independently with `--symbolize-names-only`; the plumbing is the same shape.

## Testing

- `cargo test --workspace --no-default-features`: all suites pass; 18 new unit tests in `symbol_shorten` cover the gate boundary, turbofish generics, closure tails, qualified-self, C++ operators, arrow returns, unbalanced/mismatched fallbacks, multibyte middle cuts, and the merge/no-merge grouping invariants.
- `cargo fmt --all -- --check` and `cargo clippy --workspace --no-default-features -- -D warnings` both clean.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,7 @@ pub mod sched;
 pub mod session_recorder;
 pub mod stack_recorder;
 pub mod stream;
+pub mod symbol_shorten;
 pub mod systing_core;
 pub mod tpu;
 pub mod trace;
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,6 +92,15 @@ struct Command {
     /// GiBs.
     #[arg(long)]
     symbolize_names_only: bool,
+    /// Collapse generic/template argument groups in over-long symbol names:
+    /// names longer than 256 bytes render as `path::func<...>` instead of
+    /// carrying the full monomorphized argument list, keeping their path
+    /// head and function segment. Deeply generic Rust and C++ names
+    /// legitimately demangle to multi-kilobyte strings; one hot name
+    /// repeated across millions of recorded frames dominates recorder
+    /// memory and trace size. Names of 256 bytes or less are unchanged.
+    #[arg(long)]
+    symbolize_elide_generics: bool,
     /// Disable scheduler event tracing (sched_* tracepoints and scheduler event recorder)
     #[arg(long)]
     no_sched: bool,
@@ -227,6 +236,7 @@ impl From<Command> for Config {
             no_gopclntab: cmd.no_gopclntab,
             no_exited_recovery: cmd.no_exited_recovery,
             symbolize_names_only: cmd.symbolize_names_only,
+            symbolize_elide_generics: cmd.symbolize_elide_generics,
             no_sched: cmd.no_sched,
             no_irq: cmd.no_irq,
             syscalls: cmd.syscalls,
--- a/src/stack_recorder.rs
+++ b/src/stack_recorder.rs
@@ -574,6 +574,16 @@ pub struct StackRecorder {
     /// mid-build) accumulates past any fixed budget — while the
     /// symbol-table path costs a few MiB per binary for identical names.
     names_only: bool,
+    /// When set, symbol names longer than
+    /// [`crate::symbol_shorten::ELIDE_GATE`] bytes have their generic /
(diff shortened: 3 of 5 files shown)


josefbacik:

Review executed per this repo's `.claude/agents/rust-bpf-code-reviewer.md` definition, over the full diff, before push.

### Summary

Opt-in, default-off rendering bound for pathological demangled names, applied at the single frame-formatting choke point in `stack_recorder.rs`. The transform is a bounded single-pass scanner with a deterministic fallback; the flag plumbing mirrors `--symbolize-names-only` exactly. Well scoped: no schema, no recorder, no BPF changes.

### Critical Issues

None found.

- Schema/versioning gates: no table definitions or `create_schema()` touched; `SCHEMA_VERSION` correctly unchanged (frame-string content under an opt-in flag, not schema); `Cargo.toml` version untouched (the bump-on-merge workflow owns it).
- Unbounded or nondeterministic output is ruled out by construction: every scan failure routes to the bounded middle cut with a stable FNV-1a hash of the full original name; unit tests cover unbalanced, mismatched, bracket-free, and multibyte inputs.
- The flag is applied at recorder configuration time, before recording starts, so the per-address frame-string caches never mix transformed and untransformed spellings.

### Improvements

None required.

### Suggestions

- An end-to-end integration test of the flag would need the root/VM trace harness; unit coverage of the transform plus compile-checked plumbing matches how `--symbolize-names-only` shipped.
- `format_symbolized_frame` is now at four parameters including a bool; if a third rendering option ever appears, fold them into a small options struct.

### Positive Notes

- `Cow::Borrowed` pass-through keeps the ≤256-byte common case allocation-free.
- The qualified-self and `operator<<` / `operator()` guards are each one small, tested special case rather than an attempted grammar; misparses degrade to the bounded fallback instead of wrong-balance output.
- The module docs state the downstream contract explicitly (determinism, prefix preservation, instantiation-merging) — the load-bearing properties for a grouping key.

Verified before push at 5739d9b: `cargo test --workspace --no-default-features` (all suites pass, 18 new unit tests), `cargo fmt --all -- --check`, and `cargo clippy --workspace --no-default-features -- -D warnings` all clean.

PR #167, opened 2026-07-20
Enumerate all network namespaces when network events were recorded

josefbacik:
The netns/interface enumeration behind `network_interface` records only walked recorded processes' namespaces. State-only network recording (`--only-recorder network`) records no tasks, so those traces typically carried interface rows for the host namespace alone — every pod/container-sourced socket in `network_socket` had no netns/IP row to attribute against downstream.

This extends the walk to every pid in /proc whenever the network recorder captured at least one event this session:

- Deduplicated by netns inode: one `/proc/<pid>/ns/net` stat per pid, one cgroup+comm read per NEW namespace, and the per-namespace interface enumeration (the setns round-trip) only for namespaces not already seen. On a node with ~10K processes and ~100 container namespaces this adds tens of milliseconds, once, at session end.
- Traces without network events pay nothing: the walk is gated on the network recorder having seen at least one event (`min_timestamp().is_some()`).
- Host and recorded-process entries keep precedence over walk-discovered representatives (`extend_netns_map` is `or_insert_with`-only).
- Dead pids between readdir and the /proc reads are skipped, same as the existing recorded-process path.

Downstream-visible effects, intentional:

- `network_interface` gains rows for every namespace on the node (previously ~1 row set on state-only traces). Same record shape — no schema change, `SCHEMA_VERSION` untouched.
- Consumers that classify sockets by "src IP appears on a local interface" will now classify pod-sourced flows as namespace-local instead of unattributed, and the namespace string carries the container id for non-host namespaces (existing `display_name()` behavior — this change only makes those rows exist).
- The walk's representative pid for a container namespace is typically its oldest member (the sandbox/pause process on k8s nodes), so the parenthesized comm in the namespace string may name the sandbox rather than the workload. The container id is the load-bearing field.

Testing: 450/450 lib tests including 4 new unit tests (dedup, seeded-entry precedence, dead-pid tolerance, /proc self-visibility); `cargo fmt --check` and clippy clean. Userspace-only change — no BPF programs touched.

### DIFF
--- a/src/session_recorder.rs
+++ b/src/session_recorder.rs
@@ -392,6 +392,46 @@ fn get_host_netns_inode() -> Option<u64> {
     get_netns_inode(1)
 }
 
+/// Extends `netns_map` with the network namespaces of `pids`. Existing
+/// entries win: a namespace already present keeps its representative (the
+/// host entry and recorded processes are seeded before any wider walk).
+/// Pids that disappear between enumeration and the /proc reads are skipped
+/// — `get_netns_inode` returns None for them.
+fn extend_netns_map(
+    netns_map: &mut HashMap<u64, NetnsInfo>,
+    host_netns_inode: Option<u64>,
+    pids: impl Iterator<Item = u32>,
+) {
+    for pid in pids {
+        if let Some(inode) = get_netns_inode(pid) {
+            netns_map.entry(inode).or_insert_with(|| {
+                let is_host = host_netns_inode == Some(inode);
+                NetnsInfo {
+                    inode,
+                    representative_pid: pid,
+                    container_id: if is_host { None } else { get_container_id(pid) },
+                    comm: get_comm(pid),
+                    is_host,
+                }
+            });
+        }
+    }
+}
+
+/// Every pid currently visible in /proc (numeric entries only).
+fn all_proc_pids() -> impl Iterator<Item = u32> {
+    fs::read_dir("/proc")
+        .into_iter()
+        .flatten()
+        .flatten()
+        .filter_map(|entry| {
+            entry
+                .file_name()
+                .to_str()
+                .and_then(|name| name.parse().ok())
+        })
+}
+
 /// Attempts to extract a container ID from a process's cgroup information.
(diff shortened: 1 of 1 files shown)


josefbacik:

*Mechanics: this review was produced by executing the repository's rust-bpf-code-reviewer agent definition (under the repo's agents directory) over the full diff before the PR was opened, per this repo's review workflow. Author-run; findings are structured per the definition's output format.*

### Summary

Surgical extension of the netns enumeration in `write_network_interface_records`: the existing recorded-process walk is factored into `extend_netns_map` and reused for a full `/proc` walk, gated on the network recorder having captured events this session. Purpose, cost model, and precedence semantics are documented at the call site; no BPF, schema, or unsafe surface is touched.

### Critical Issues

None.

- No schema change (`network_interface` row volume grows; record shape unchanged) — `SCHEMA_VERSION` correctly untouched, no `[bump minor]` title token needed.
- `Cargo.toml`/`Cargo.lock` untouched — the version bump belongs to the merge workflow.

### Improvements

None required. One considered and rejected: logging a walked-namespace count at session end — the surrounding stage already prints progress markers, and the count is recoverable from the emitted rows themselves.

### Suggestions

- `all_proc_pids`' double `.flatten()` (Result-of-ReadDir, then per-entry Results) is idiomatic but dense; acceptable as-is with its one-line doc comment.
- Silent degradation when `/proc` is unreadable matches the file's existing conventions (`get_netns_inode` / `get_container_id` return `Option`); if observability on partial walks is ever wanted, that is a separate change.

### Positive Notes

- The factoring removes a previously duplicated closure and tightens lock scope: the `process_descriptors` read guard now drops before the wider walk begins, and the `network_recorder` mutex guard releases at the end of the plain-`if` condition (bool result — no guard extends into the block).
- Precedence (`or_insert_with`-only) is pinned by a dedicated test (`extend_netns_map_existing_entry_wins`); dead-pid tolerance and /proc self-visibility each have deterministic tests (over-`pid_max` pids; `/proc/self`).
- The gating comment explains why (state-only recording's attribution gap and the per-namespace `setns` cost), not what.

PR #168, opened 2026-07-21
Enumerate namespace interfaces via /proc instead of setns

josefbacik:
Entering a network namespace with `setns(CLONE_NEWNET)` requires `CAP_SYS_ADMIN`. Recorders running under a restricted capability set (ptrace/BPF, no SYS_ADMIN) had every non-host interface enumeration fail — and fail invisibly, because the `setns` error path returned an empty list indistinguishable from a namespace that has no addresses. The visible symptom: `network_interface` tables carry host rows only, so sockets from other namespaces stay unattributable downstream. This is also why interface tables were host-only on historical traces from such deployments — the namespace walk found the namespaces, but never their addresses.

This replaces the namespace-entry approach with reads from `/proc/<pid>/net/`, which reflects the pid's namespace and needs only ordinary file access:

- **IPv6** from `if_inet6`, which names the owning interface.
- **IPv4 local addresses** from the `fib_trie` local-table dump (`/32 host LOCAL` entries, deduplicated across the `Local:`/`Main:` sections; broadcast and routed prefixes excluded).

Trade-offs, intentional:

- `fib_trie` does not name interfaces, so IPv4 addresses inside non-host namespaces are grouped under an empty interface name rather than guessed. The `namespace` and `ip_address` fields consumers key on are unaffected; interface names remain populated for IPv6 entries and for the host namespace (unchanged `getifaddrs` path).
- A non-host namespace whose enumeration comes back empty now logs one warning line (bounded by namespace count per trace). The silent-empty behavior is precisely how the capability dependency stayed unnoticed — an instrument that cannot see reports nothing, which reads as "nothing there".
- The function's two `unsafe` blocks are removed; no namespace switching means no restore-failure hazard on the recording thread.

Testing: 453/453 lib tests, including three new — `if_inet6` parsing against the real format (malformed lines skipped), `fib_trie` local-address extraction (dedup across sections, broadcast/routed excluded), and an own-namespace enumeration test asserting loopback is found via `/proc/self` with no namespace-entry privileges — that last one is the regression test for the capability-free property itself. `cargo fmt --check` and clippy clean. Userspace-only; no BPF, no schema change, `SCHEMA_VERSION` untouched.

### DIFF
--- a/src/session_recorder.rs
+++ b/src/session_recorder.rs
@@ -1,7 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::ffi::CStr;
 use std::net::{Ipv4Addr, Ipv6Addr};
-use std::os::unix::io::AsRawFd;
 use std::sync::{Arc, Mutex, OnceLock, RwLock};
 
 use anyhow::Result;
@@ -33,7 +32,7 @@ use perfetto_protos::process_descriptor::ProcessDescriptor;
 use perfetto_protos::process_tree::process_tree::Process as ProtoProcess;
 use perfetto_protos::system_info::Utsname;
 use perfetto_protos::thread_descriptor::ThreadDescriptor;
-use std::fs::{self, File};
+use std::fs;
 use std::os::unix::fs::MetadataExt;
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


josefbacik:

*Mechanics: this review was produced by executing the repository's rust-bpf-code-reviewer agent definition (under the repo's agents directory) over the full diff before the PR was opened, per this repo's review workflow. Author-run; findings are structured per the definition's output format.*

### Summary

Replaces `setns(CLONE_NEWNET)`-based interface enumeration with `/proc/<pid>/net/` reads (`if_inet6` + `fib_trie` local table), removing the `CAP_SYS_ADMIN` dependency and the silent-empty failure mode it caused under restricted capability sets. Adds a per-namespace warning when a non-host enumeration returns nothing. Two `unsafe` blocks are removed outright.

### Critical Issues

None.

- No schema change (`network_interface` record shape unchanged) — `SCHEMA_VERSION` correctly untouched.
- `Cargo.toml`/`Cargo.lock` untouched — version bump belongs to the merge workflow.
- No BPF surface.

### Improvements

None required. Considered and rejected: attempting to recover IPv4 interface names by correlating `fib_trie` with `/proc/<pid>/net/dev` — there is no per-address mapping in proc, and a guessed name is worse than an honest empty one; the body discloses the trade.

### Suggestions

- `parse_fib_trie_local`'s leaf detection relies on the `|-- ` marker and the `/32 host LOCAL` attribute string; both have been stable across kernel versions for years, and malformed content degrades to fewer addresses plus the new warning rather than wrong ones. Acceptable; the parser doubles up on safety by re-parsing the address text through `Ipv4Addr`.
- The empty-string interface grouping is documented at the parse site and the emit site; if a future consumer needs v4 names inside namespaces, netlink-via-namespace-entry remains the only source and would need the capability discussion this change avoids.

### Positive Notes

- The warning is placed at the emit site where namespace identity (name + representative pid) is in scope, and is bounded by namespace count — it converts a silent capability regression into a visible, attributable log line.
- The own-namespace loopback test pins the capability-free property itself, not just parser behavior: it fails on any future change that reintroduces a privileged path.
- Net `unsafe` count in the file goes down by two, and the thread-namespace restore-failure hazard (previously a WARNING-and-hope path) is gone entirely.

josefbacik:
No, we should just document that we need CAP_SYS_ADMIN for this and then complain loudly if it is not there for the network recorder. We can use this as a fallback if we don't have CAP_SYS_ADMIN, and we should log that fallback appropriately, but we should be steering towards utilizing CAP_SYS_ADMIN so that we have full functionality.

josefbacik:
Closing — superseded by granting CAP_SYS_ADMIN to the agent deployment, which restores the direct setns path. The /proc enumeration workaround is no longer needed.

PR #169, opened 2026-07-22
exit_snapshot: only pin ELF backing files

josefbacik:
## Summary

`exit_snapshot` currently pins any executable file-backed mapping's
backing file. Executable file-backed mappings are not always ELF
objects: runtimes that manage code or memory through memfd — sandbox
runtimes such as gVisor mapping guest memory arenas, JITs mapping code
files — produce maps entries that pass the executable/file-backed
screen but hold raw bytes symbolization can never resolve. This PR
screens the backing file at pin time: read four bytes, require the ELF
magic.

## Why pinning non-ELF files is worse than useless

- **It can never produce a name.** The snapshot's whole purpose is
  feeding `Source::Elf` at end-of-pass symbolization; a non-ELF file
  fails ELF parsing every time. The pin is pure cost.
- **Descriptor pressure scales with the wrong thing.** On hosts running
  ordinary workloads, executable mappings dedupe to a small set of real
  binaries (a few dozen unique files even under heavy fork/exec load).
  On hosts dense with sandboxed workloads, every sandbox's memory arena
  is a *distinct* memfd — the pinned set saturates its cap with useless
  files, and the per-pass descriptor peak (pins plus the symbolizer's
  own per-source descriptors) scales with sandbox churn. Under a
  1024-class `RLIMIT_NOFILE` that peak is reachable every pass.
- **A held descriptor extends dead processes' memory lifetimes.** The
  pins are real descriptors; holding one to an unlinked memfd keeps the
  arena's pages alive until the next reset — deferring reclaim of
  potentially large amounts of memory for files that were never
  symbolizable in the first place.

## The change

`open_mapping_file` now verifies the ELF magic on whichever path opened
the file (map_files primary or the verified fallbacks) before returning
it. A file that fails the screen fails on every path — they all reach
the same inode — so no further candidates are tried. A mapping whose
file fails the screen is simply not recorded: its addresses keep the
`[exited]` rendering, the same degradation contract as every other
capture failure. Cost: one 4-byte `pread` of an already-open descriptor
per unique file per snapshot pass.

## Validation

- New unit test `is_elf_screens_magic` (own exe passes; junk and empty
  files fail).
- New unit test `capture_skips_non_elf_backing_files` exercises the
  exact production path: an exec-mapped non-ELF memfd (which presents
  in maps as executable + file-backed, just like a runtime arena) is
  screened out of the snapshot while real ELF mappings still pin.
- Full lib suite green; clippy `-D warnings` clean.
  20260721-222030), including the gVisor/runsc recording target and the
  exited-process recovery e2e (trace-validation 33/33) — the recovery
  feature is verified unaffected for real binaries (the test binary is
  a real ELF and passes the screen).

### DIFF
--- a/src/exit_snapshot.rs
+++ b/src/exit_snapshot.rs
@@ -213,8 +213,8 @@ impl ExitSnapshots {
     }
 
     /// Ensure the backing file of a parsed mapping is pinned. Returns false
-    /// when the file cannot be opened (the segment is then not recorded and
-    /// its addresses keep the [exited] rendering).
+    /// when the file cannot be opened or is not an ELF object (the segment
+    /// is then not recorded and its addresses keep the [exited] rendering).
     fn pin_file(&mut self, tgid: i32, parsed: &ParsedMapsLine) -> bool {
         if self.files.contains_key(&parsed.file) {
             return true;
@@ -253,10 +253,16 @@ impl ExitSnapshots {
 /// path through the process's root, then the plain path — each verified
 /// against the mapping's (device, inode) so a recycled or namespace-mismatched
 /// path can never pin the wrong file.
+///
+/// Whatever path opened it, the file must pass the ELF-magic screen: a
+/// non-ELF backing file can never symbolize, so pinning it is pure cost
+/// (see [`is_elf`]). A file that fails the screen fails it on every path
+/// (they all reach the same inode), so there is no point trying the next
+/// candidate.
 fn open_mapping_file(tgid: i32, parsed: &ParsedMapsLine) -> Option<File> {
     let link = format!("/proc/{tgid}/map_files/{:x}-{:x}", parsed.start, parsed.end);
     if let Ok(file) = File::open(link) {
-        return Some(file);
+        return is_elf(&file).then_some(file);
     }
     for candidate in [
         format!("/proc/{tgid}/root{}", parsed.path),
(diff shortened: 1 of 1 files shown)


josefbacik:
## Code review

definition over the full diff (5129893 vs v1.11.28).

### Summary

Single-file change adding an ELF-magic screen at pin time in
`exit_snapshot`. Semantics follow the module's existing degradation
contract exactly (screened mapping = not recorded = `[exited]`
rendering); the screen closes a class of descriptor-pressure and
memory-lifetime hazards on hosts where executable mappings are backed
by non-ELF memfds, at the cost of one 4-byte pread per unique file per
pass.

### Critical Issues

None.

- Schema, `Cargo.toml` version, recorder criteria, Python-version
  criteria: no such surfaces touched.
- Degradation safety verified: `capture()` with every segment screened
  produces an empty snapshot, which `observe_tgid` declines to insert;
  `resolve()` on the missing tgid returns None; frames keep `[exited]`.
  Identical to the existing budget/eviction degradation paths.
- No behavior change for real binaries: ELF backing files pass the
  screen unconditionally, so populations where the recovery feature
  pays (fork/exec-heavy real-binary workloads) are unaffected.

### Improvements

- None required. A conceivable follow-up is short-circuiting the maps
  walk entirely for processes whose mappings are predominantly non-ELF,
  but the walk is already bounded (`MAX_SEGMENTS_PER_PROC`, per-second
  capture budget) and the screen removes the expensive part (pins and
  symbolizer feed), so the added complexity isn't warranted.

### Suggestions

None — comment density and placement match the module's house style
(mechanism rationale on the helper, one-line contract note on
`pin_file`).

### Positive Notes

- One screen closes three hazards at once: per-pass descriptor peaks,
  deadweight end-of-pass symbolization attempts, and dead-process
  memory-lifetime extension via held descriptors.
- `capture_skips_non_elf_backing_files` tests the true production path
  (exec-mapped memfd → map_files open → screen) rather than the helper
  in isolation.

PR #170, opened 2026-07-22
stack_recorder: bound symbolizer descriptors in the exited-process pass

josefbacik:
## Problem

In continuous mode, ticks on fork/exec-heavy hosts can fail with `Too many open files (os error 24)` opening `stack.parquet`, right after the "exited-process pass done" phase marker.

Since #165, the exited-process pass symbolizes dead processes' user addresses through `/proc/self/fd/<n>` paths of backing files pinned by the first-sample snapshot layer (`exit_snapshot.rs`). Those pins are bounded (512 files, oldest-first eviction) and released at the end of each tick. But blazesym's file caches open *their own* descriptor for every distinct binary symbolized and never evict (`FileCache`'s docs: "stale/old entries are never evicted"), so a single tick holds:

- up to 512 snapshot pins (by design, released after the pass), plus
- one *duplicate* descriptor per distinct binary the symbolizer touched — unbounded, held until the symbolizer drops at the end of the tick.

On hosts with heavy short-lived-process churn — in particular sandboxed (gVisor) workloads, where memfd/stub-region backings give each process a distinct binary set — the duplicates approach the pin cap, the descriptor table fills, and the first `stack.parquet` open after the exited pass fails with EMFILE. The symbolizer's own failed opens are swallowed by design, so symbolization degrades silently (`unknown ([exited])` frames) before the hard failure surfaces.

The existing RSS-budget rebuild valve (`maybe_rebuild`) does not help: it is gated on anon RSS, and this failure mode shows up at tens of MiB of anon RSS, far under any budget.

## Fix

Bump blazesym to v0.2.6, which provides `Symbolizer::evict()` (libbpf/blazesym#1613), and evict what the exited pass touches:

- `finish_inner` tracks the distinct exit-snapshot ELF paths the symbolizer actually symbolizes through;
- the batch bound is enforced after **every** symbolization attempt (even a failed attempt leaves a descriptor-holding cache entry): once `EXITED_ELF_EVICT_BATCH` (64) distinct paths accumulate they are evicted via `Evict::Elf`, closing blazesym's descriptors — one deep stack crossing many binaries cannot exceed the cap;
- a final sweep runs when the pass ends, including when it bails out early on a drain error;
- an eviction failure warns and continues: a leaked descriptor is strictly better than a failed tick.

Net effect: the symbolizer's exit-snapshot descriptors are bounded at 64 at any instant, so a tick's peak is the pins (<= 512) plus 64 plus runtime descriptors — comfortably under a 1024 soft limit. (When debug info is enabled, blazesym may additionally cache separate on-disk debug-info files under their own paths; those are deduplicated across binaries and absent entirely in the memfd-backed sandboxed-workload scenario that motivated this fix.)

The RSS rebuild valve is untouched (it guards an orthogonal resource). Evicting by path also removes the cache entry itself, so under eviction a stale entry cannot outlive a descriptor number's reuse; today each tick symbolizes with a fresh symbolizer, so staleness cannot cross ticks, but this matters for a future longer-lived symbolizer.

## Considered and deferred

- **Evicting the live pass's gVisor bridge (`/proc/<pid>/map_files/...`) Elf entries.** The live pass ran under the descriptor limit long before #165, so it is not the leak; its entries are bounded by the live-process count, and per-group eviction would thrash re-parses of ELFs shared across processes. If measurement after this fix ever shows pressure there, the right shape is screening at bridge-open (rejecting memfd arenas before a cache entry forms), not live-pass eviction.
- **Fallback without a dependency bump:** a descriptor-count trigger on `maybe_rebuild` — count `/proc/self/fd` entries against `getrlimit(RLIMIT_NOFILE)`'s soft limit and rebuild the symbolizer at about 0.8x of it. Coarser (a trip costs a full re-parse of everything cached) and it papers over the unbounded growth rather than removing it; recorded here in case the dependency bump ever needs reverting.

## Tests

- `test_evict_exited_elf_paths_closes_descriptors` (unit): symbolizing through a `/proc/self/fd` path holds a duplicate descriptor; evicting the path closes it, leaving only the pin. Descriptors are counted by readlink match against a uniquely named binary copy, so concurrent tests cannot perturb the count.
- `test_evict_exited_elf_paths_guards_fd_path_reuse` (unit): after eviction, re-pointing the same descriptor number at a different file (`dup2`) leaves no descriptor for the old file, and the path is re-read fresh. This pins safe behavior for a future longer-lived symbolizer; the test's doc comment explains why it cannot be hit today.
- `test_exited_pass_fd_bound` (integration, `#[ignore]`, needs root/BPF): 450 short-lived children, each running its own copy of a symbol-bearing binary, traced under a soft `RLIMIT_NOFILE` of 700; at least 80% of the children's exited frames must symbolize into their own binary. Without eviction the pass exceeds the limit and frames silently degrade, failing the assertion. (The child count deliberately stays under the snapshot layer's 512 pinned-file cap so its own oldest-first pin eviction never competes with the property under test.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -677,8 +677,8 @@ dependencies = [
 
 [[package]]
 name = "blazesym"
-version = "0.2.5"
-source = "git+https://github.com/libbpf/blazesym.git?rev=862c2cf5d424307456322d2c68fe86c591baece2#862c2cf5d424307456322d2c68fe86c591baece2"
+version = "0.2.6"
+source = "git+https://github.com/libbpf/blazesym.git?rev=8705da0a8cec2bb27f897ac732e27121f9be7936#8705da0a8cec2bb27f897ac732e27121f9be7936"
 dependencies = [
  "cpp_demangle",
  "crc32fast",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ bitfield = "0.19.0"
 # 0.2.x releases (where Sym::code_info became Option<Box<CodeInfo>>), so
 # `cargo install` without --locked picked an incompatible version. An exact pin
 # keeps fresh resolutions reproducible.
-blazesym = { git = "https://github.com/libbpf/blazesym.git", rev = "862c2cf5d424307456322d2c68fe86c591baece2" }
+blazesym = { git = "https://github.com/libbpf/blazesym.git", rev = "8705da0a8cec2bb27f897ac732e27121f9be7936" }
 clap = { version = "4.5.20", features = ["derive"] }
 ctrlc = "3.4"
 dashmap = "6"
--- a/src/stack_recorder.rs
+++ b/src/stack_recorder.rs
@@ -1,5 +1,6 @@
 use std::collections::hash_map::RandomState;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fs::File;
 use std::io::{BufReader, BufWriter, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
@@ -21,7 +22,7 @@ use crate::utid::{ThreadAwareRecorder, UtidGenerator};
 use blazesym::helper::{read_elf_build_id, ElfResolver};
 use blazesym::symbolize::source::{Elf, Kernel, Process, Source};
 use blazesym::symbolize::{
-    cache, Input, ProcessMemberInfo, ProcessMemberType, Resolve, Sym, Symbolizer,
+    cache, evict, Input, ProcessMemberInfo, ProcessMemberType, Resolve, Sym, Symbolizer,
 };
 use blazesym::Error as BlazeErr;
 use blazesym::Pid;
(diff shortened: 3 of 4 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — PASS (no critical issues)

*Mechanics: review executed with the repository's `rust-bpf-code-reviewer` agent definition (blob `f09e9b60`) over the full diff `e27bf22..2a37a65` (Cargo.toml, Cargo.lock, src/stack_recorder.rs, tests/exited_fd_bound.rs), prior to push. Two revisions followed the initially reviewed one, each interdiff read under the same criteria: a four-line comment in the batch-eviction block (documenting the benign path-collision case, added on reviewer request), and a recalibration of the integration test (child lifetime 0.4s → 10s plus a pre-recording baseline-descriptor canary assert) after gate runs on a slow emulated environment showed short child lifetimes lose the exit-snapshot race — which both starves resolution and deflates the very descriptor demand the limit is calibrated against, making the test blind to the property in either direction there. The descriptor limit was probed at both 700 and 800 during recalibration and kept at 700: a full patched run under 700 passed, so the original limit already holds the patched peak with margin and remains the stricter negative control. The recalibrated negative control (unpatched tree) and patched run are the gate evidence for this PR.*

### Summary
The change bounds symbolizer descriptor usage in the exited-process pass by tracking the exit-snapshot ELF paths the symbolizer touches and evicting them (blazesym v0.2.6 `Evict::Elf`) once 64 distinct paths accumulate, plus an end-of-pass sweep that also covers the bail-out path. The eviction family (Elf, not Process) matches where the descriptors are actually held, and the tests pin the load-bearing properties rather than incidental behavior.

### Critical Issues
None.
- `version` in Cargo.toml / the systing entry in Cargo.lock: untouched (verified — the dep-bump commit's Cargo.toml delta is the single blazesym `rev` line; the lock delta is blazesym `0.2.5 → 0.2.6` plus its source field only, with zero transitive churn).
- No schema surfaces touched → no SCHEMA_VERSION / SCHEMA_CHANGES.md / `[bump minor]` obligations.
- All three `unsafe` blocks (rlimit get/set and dup2, all in tests) are minimal and carry `// SAFETY:` comments.

### Improvements
None required. The two candidate concerns I chased both resolve in the diff itself:
- Mid-stack lock gap (drop → evict → re-acquire): safe by construction — the resolved snapshot data is owned at the drop point (`elf_path` is moved into the touched set first), so the compiler rules out references outliving the guard, and the next iteration resolves through the re-acquired guard.
- Eviction keyed by the `/proc/self/fd/<n>` path string requires blazesym to key its cache by the path as given (no canonicalization) — pinned explicitly by `test_evict_exited_elf_paths_closes_descriptors`, so a future blazesym behavior change fails the test instead of silently reintroducing the leak.

### Suggestions
- A one-line comment in the eviction block naming the benign collision case (a descriptor number recycled to a new snapshot pin during the lock gap → evicting the new file's entry → harmless re-parse) would save the next reader the derivation. Non-blocking.
- `symbolize_stack_frames` now carries 8 parameters behind `#[allow(clippy::too_many_arguments)]`; if it grows again, a small context struct is the natural refactor. Not this PR.

### Positive Notes
- The batch bound is enforced per symbolization attempt, not per stack, so a single deep stack crossing many binaries cannot exceed the cap — and the comment explains exactly why.
- Capturing the drain error and sweeping before re-raising keeps descriptor accounting uniform on every exit path without a drop guard that would have to hold the symbolizer.
- The failed-parse eviction assertion in the first unit test closes the subtle case where an unparseable entry still holds a descriptor.
- The integration test asserts the user-visible property (exited frames actually symbolize under a lowered RLIMIT_NOFILE) rather than internal counters, and its margins are documented in the constants.

PR #171, opened 2026-07-22
Revert exited-process recovery (#165, #169, #170)

josefbacik:
## Summary

Reverts the exited-process recovery feature and its two dependent
fixes, as three mechanical merge reverts (newest first):

- Revert #169 (`exit_snapshot: only pin ELF backing files`)
- Revert #170 (`stack_recorder: bound symbolizer descriptors in the
  exited-process pass`, including the blazesym v0.2.6 bump)
- Revert #165 (`Symbolize processes that exit before end-of-trace
  symbolization`)

## Why

#165's snapshot-hint path does too much work in probe context. On each
tgid's first stack sample, the sampling program updates the
`recovery_hinted_tgids` LRU map and reserves/submits into the hint
ring buffer — and that code runs under scheduler/IRQ-path locks
(`sched_switch`, the IRQ tracepoints), which must stay near-free. On
hosts whose process populations exceed the LRU's 8192 entries within
seconds of a capture starting, every subsequent insert also contends
LRU eviction, from IRQ-disabled context, across all cores at once.

This is the same failure class as the one fixed in b25c202
("pystacks: move symbol interning off the BPF probe path"): probe-side
map work → raw spinlock contention → on AWS guests, paravirtualized
spinlock waiting (`kvm_wait` halting with interrupts off) escalates a
long-held lock into CPUs that never come back. In production this
presented as hard CPU stalls and ENA `NETDEV WATCHDOG` transmit-queue
wedges beginning within seconds of the capture-stop transition on a
small percentage of many-core AWS hosts per capture cycle, with
affected hosts losing connectivity until reboot. Hosts running the
previous release through the identical restart cadence showed no
excursion above ambient.

## Scope: why all three

#169 and #170 exist only to serve the #165 machinery (#169 screens what
the snapshot layer pins; #170 bounds the symbolizer descriptors its
exited pass creates). With #165 gone they are dead code on deleted
files. Reverting the trio as pure `git revert` commits keeps every
change mechanically auditable against history rather than hand-crafting
a combined diff; the blazesym v0.2.6 bump goes with #170 since nothing
calls `Symbolizer::evict()` afterward — the replacement work will
re-introduce the pieces it needs.

The only hand-touched region is the conflict resolution in
`stack_recorder.rs` where #166 (`--symbolize-elide-generics`, which
stays) interleaved with #165's hunks. Resolution verified two ways: the
BPF source is byte-identical to v1.11.25, and the residual diff of
every reverted file against v1.11.25 is exactly the surviving elision
commit's changes (minus its one line that lived inside the removed
block).

## Accepted regression

Short-lived processes' user frames return to `unknown ([exited])` —
roughly doubling that unresolved share on fork/exec-heavy hosts
(measured on matched windows). That symbolization win is deliberately
given back until a replacement lands that does no per-sample probe-side
work and no per-tick full-population scanning. The replacement
direction (build-id-based offline symbolization, where the kernel's
`BPF_F_USER_BUILD_ID` stack walker provides frame identity and
resolution happens against a build-id-keyed store, out of band) is
being designed separately.

## Validation

- `cargo build`, 442/442 lib tests, `clippy -D warnings`, `fmt` clean.
  20260722-135437), including the gVisor/runsc recording target;
  trace-validation 32/32 with the reverted feature's e2e correctly
  absent.
- Byte-equivalence proofs above.


josefbacik:
## Code review

definition over the full diff (c47220d vs v1.11.30). Note: the definition at this
revision predates the maintainer's announced IRQ-path review additions; this review
applies the current in-repo instructions.

### Summary

Three mechanical merge reverts removing the exited-process recovery
machinery and its two dependent fixes. The only hand-authored content
is the three-hunk conflict resolution in `stack_recorder.rs` where the
surviving `--symbolize-elide-generics` feature interleaved with the
reverted hunks; everything else is the byte inverse of named upstream
merges.

### Critical Issues

None.

- Hand-resolution audited hunk-by-hunk (field block, constructor,
  exited None-branch) and proven by construction: the residual diff of
  all touched files vs v1.11.25 equals exactly the surviving elision
  commit's line set, minus its single line that lived inside the
  removed recovery block.
- BPF source byte-identical to v1.11.25 — the sampling-path hint
  emission (the IRQ-path cost this revert exists to remove) is fully
  gone; no partial state remains.
- Zero references to the removed module anywhere in sources, tests, or
  Cargo metadata; blazesym pin restored to its pre-bump revision;
  package version untouched (bump-on-merge owns it); no schema
  surfaces involved.

### Improvements / Suggestions

None — a revert should be exactly this shape.

### Positive Notes

- Trio-revert scope keeps the tree free of dead code on deleted files
  and keeps each commit independently auditable against history.
- The accepted-regression numbers in the body are measured (matched
  production windows), not estimated.

PR #172, opened 2026-07-22
systing_core: bracket every capture-stop step with timed log lines

josefbacik:
## Summary

Brackets every step of the capture-stop transition with timed log
lines: the tracing-gate flip, each consumer/worker join, the
socket-metadata load, BPF teardown (made an explicit step), the
recorder drain, and the symbolize-and-write phase.

## Why

Hard CPU stalls and NIC watchdog wedges have been observed beginning
within seconds of the capture-stop transition, and host logs could
localize them no further than "somewhere after `Stopping...`". Each
bracket prints its begin line BEFORE the step runs and an elapsed line
when it completes — so a host that wedges mid-step leaves the unclosed
bracket as its last shipped log line, and the next occurrence
self-localizes to the exact operation in flight with no further
instrumentation.

The BPF teardown becomes an explicit step in the process: the probe
links and skeleton previously detached implicitly at end of scope,
after unrelated log lines — exactly the ambiguity the brackets exist
to remove. Drop order is unchanged (probe links before skeleton, same
as the prior implicit reverse-declaration order); it is only made
explicit and timed.

## Cost

Logging only: one line pair per stop step per capture. No behavior
changes, no new threads, nothing in the sampling or IRQ paths.

## Validation

- `cargo build`, 442/442 lib tests, `clippy -D warnings`, `fmt` clean.
  20260722-154527) — every end-to-end target exercises the instrumented
  stop path. (The harness captures passing tests' stdout, so the phase
  lines are not visible in the gate logs; they are unconditional
  `println!` on the executed path.)

### DIFF
--- a/src/systing_core.rs
+++ b/src/systing_core.rs
@@ -17,7 +17,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::{channel, sync_channel, Receiver, Sender, SyncSender};
 use std::sync::{Arc, Mutex};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use crate::events::{EventKeyType, EventProbe, SystingProbeRecorder};
 use crate::network_recorder;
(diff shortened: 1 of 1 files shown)


josefbacik:
## Code review

definition over the full diff (60c87e8 vs its base).

### Summary

Logging-only change: an RAII bracket helper plus timed brackets around
each existing capture-stop step, and the previously-implicit BPF
teardown made an explicit, timed step. One file, +86/−21.

### Critical Issues

None.

- No BPF program changes; nothing added to sampling or IRQ paths.
- The one semantics-adjacent edit — explicit `drop` of the probe links
  and skeleton — preserves the prior implicit drop order exactly
  (reverse declaration: memory-alloc links, probe links, skeleton);
  verified against the block's declaration order.
- Schema, versioning, recorder, and Python-support criteria: no such
  surfaces touched.

### Improvements / Suggestions

None. No unit tests added — deliberate for a logging-only change; the
VM gate exercises every bracket in every end-to-end target, and the
phase lines appearing in target output is the functional proof.

### Positive Notes

- Begin-line-before-step is the load-bearing design choice: it converts
  a wedge from "somewhere in the stop path" into a named operation in
  shipped logs, with zero steady-state cost.

PR #173, opened 2026-07-23
stack_recorder: build-id user stacks for exited-process symbolization

josefbacik:
## Summary

Adds `--collect-build-id`: user stacks are captured as **(build-id, file
offset)** pairs instead of raw virtual addresses, so a frame's identity no
longer depends on its process being alive at end-of-trace symbolization.
Frames of exited processes resolve by name whenever a build-id-keyed
store knows the binary — and when it doesn't, they render as
`unknown ([buildid:<hex>]) <0x<offset>>`: a durable, aggregatable
identity that can be resolved offline later, unlike a raw ASLR-slid hex
address. Off by default; the off configuration is byte- and
instruction-identical to before this change.

## Why

Processes that exit before end-of-trace symbolization render as
`unknown ([exited]) <hex>` — on fork/exec-heavy hosts (CI) that is
20-40% of CPU samples. #165 attacked this with live capture (per-sample
hints, /proc sweeps, fd pins) and was reverted (#171): its per-sample
work ran under sched-path locks, which is exactly the overhead class
#126 exists to keep out of the probes.

This is the opposite approach: **let the kernel attach the identity at
walk time, and move all resolution to the end of the trace.**
`bpf_get_stack` already supports `BPF_F_USER_BUILD_ID` — the kernel's
stack walker (the same mechanism perf uses) resolves each user frame
against the mapped file's ELF build-id note and emits
`(build_id[20], file_offset)` instead of an IP. The extraction is
kernel-internal, bounded, and trylock-guarded: where it cannot walk
safely (contended mmap lock, missing note, anonymous mapping) it falls
back to the raw IP with `BPF_STACK_BUILD_ID_IP` status, so capture is
never worse than today. No systing code runs per frame; probes stay
lookup-only.

Symbolization then becomes a pure lookup:

    (build_id, offset) -> build-id-keyed store -> name

Dead processes stop being a special case at all. A side benefit: the
result cache is keyed by `(build_id, offset)` and is fully
process-independent — one resolution serves every process that ever
mapped that binary, where the per-process address caches could not
share anything across tgids.

Binaries with no host-visible backing file get the same durability:
sandbox runtimes that map guest executables through memfd (unopenable
from the host by path) still carry ELF notes the kernel can read at
walk time, so those frames gain a resolvable identity for the first
time.

## Design

**Capture (BPF):** the event carries **one** `user_stack` region,
tail-positioned and sized for the 32-byte build-id entries. A rodata
knob (`tool_config.collect_build_id`) selects the walk: build-id mode
fills the region with `bpf_stack_build_id`-shaped entries and reserves
it fully; raw-IP mode packs u64 addresses into the same region and
reserves only the bytes it uses (`offsetof(user_stack) + depth * 8` —
byte-for-byte the pre-change reservation), with the ring consumer
zero-extending short records (the memory ring's established
variable-length pattern). A `user_stack_format` flag on the event makes
every record self-describing: the capture mode is fixed per run, so
the flag is belt-and-braces for downstream readers rather than
load-bearing. Because the knob is `const volatile` rodata, the verifier
prunes the dead arm at load time (same mechanism the `no_irq` gate
documents), so flag-off adds zero executed instructions and zero
reservation bytes. A layout test pins the region's tail position and
u64 alignment (the packed-IP view depends on both) and the mirror
struct's equivalence to the kernel's `bpf_stack_build_id` (also
`_Static_assert`ed on the BPF side).

**Store (userspace, end-of-trace):** `src/build_id_store.rs`. Fill
sources in trust order:

1. `.build-id` debug directories (`/usr/lib/debug/.build-id/xx/rest`) —
   trusted: the path is derived from the id, so a traced process cannot
   choose which file answers for an id it doesn't own.
2. debuginfod (with `--enable-debuginfod`) — trusted for the same
   reason (queried by id). The client is now shared between the
   existing process dispatcher and the store.
3. Live binaries — opportunistic: one host-wide pass at symbolization
   start indexes every live process's executable mappings by their
   build-id note, read via namespace-immune `/proc/<pid>/map_files`
   links (falling back to the maps path where map_files needs
   privileges the caller lacks). Host-wide rather than
   sampled-processes-only because the binaries a dead process mapped
   are often held open only by processes that were never sampled — the
   `systing record -- short_cmd` shape, where the traced command dies
   and its binary lives on in the unsampled parent. Nothing here runs
   during capture.

Live fills **never override** a trusted entry and never overwrite
anything already stored: a binary carrying a copied build-id note can
only affect the rendering of ids that no trusted source knows, which is
the same capability class as forging sample content outright. The store
caps its entries (16384) with LRU eviction, so unique-build-id storms
bound memory instead of growing it.

**Ordering:** the live-process index runs before any symbolization
pass, so exited-process stacks resolve in the existing first pass with
its emit-immediately memory behavior — no pass restructuring.

**Rendering:** store hits render normally (module = binary basename for
live fills, `[buildid:<hex8>]` for id-derived sources; the trailing
`<...>` slot carries the file offset — the frame's identity within the
module). Store misses keep the full id:
`unknown ([buildid:<40-hex>]) <0x<offset>>`. The full id is the point —
truncating it would break offline re-resolution.

**Deliberately absent** (the entire reverted-#165 machinery class):
hint ringbufs, seen-LRU maps, /proc sweeps at capture time, map_files
pins, fd caches, exec-invalidation handlers. Nothing accumulates during
capture, so there is nothing to bound.

## Cost

Paid **only when the flag is on**:

- Stack-ring reservations grow from 2720B to 3584B per event (1.32x)
  in the pystacks build — the one region at 32B entries instead of 8B.
  Without pystacks compiled in the base record is much smaller, so the
  ratio is steeper (~640B -> ~1500B, ~2.4x); sizing the stack rings
  accordingly is part of opting in. (A dual-walk shape that emits raw
  IPs plus a per-binary sideband was considered and is the fallback if
  this ratio ever matters; the flat capture is simpler and the numbers
  are pinned by `test_stack_event_user_stack_region_layout`.)
- The kernel does the note extraction inside `bpf_get_stack` (page-
  cache reads, `mmap_read_trylock`, IP fallback on contention — no
  sleeping, no IO, no locks taken by our code).
- End-of-trace: one /proc maps read per live process plus one ELF-note
  read per distinct executable mapping (the host-wide index; hundreds
  of milliseconds on busy hosts, in the teardown path), and one
  `.build-id` stat (plus optional debuginfod fetch, memoized per id)
  per unique build-id.

Flag off: reservation sizes, BPF instruction streams (rodata prune),
and all userspace paths are unchanged; the spill format gains a 1-byte
frame tag (run-internal file, no compatibility surface).

Object-level check (base vs branch, `llvm-objdump -d`, x86_64): **zero
instruction delta in every BPF program section** — sched, perf, IRQ,
network, memory programs are unchanged except relocated call-target
constants. The gated arms live in `emit_stack_event_with_ts`'s block
graph in `.text` (net −14 instructions vs base — the single-region
layout generates tighter code), and the build-id arm is pruned at load
when `collect_build_id` is 0 — the same rodata dead-code elimination
the `no_irq` gate documents. The BPF object also compiles clean for
aarch64 (layout `_Static_assert` included).

## Validation

- 450/450 lib tests, `clippy -D warnings`, `fmt` clean. New unit
  coverage: store trust-precedence (live never overrides trusted,
  misses memoized, LRU cap eviction), spill round-trip of mixed
  build-id/IP frames, dedup-hash discrimination (an IP and an equal
  offset must not alias; different ids must not alias), deferred-render
  format, and the layout pin.
- `test_finish_inner_build_id_dead_frame_resolves_via_live_fill`: a
  dead-tgid stack whose build-id frame resolves to a named symbol
  because a live process (the test itself) maps the binary — the core
  promise, proven without a VM.
- New `build_id_record` integration test (VM): records a traced child
  that exits before symbolization, negative-control first (flag off:
  frames render `[exited]`, zero build-id vocabulary), then asserts the
  same dead process's frames resolve **by name** with the flag on. Its
  first run caught a real gap — a pid-filtered trace samples no live
  process, so a sampled-tgids-only fill had nothing to index (frames
  stayed in the deferred form, end-to-end capture proven but no
  names) — which is what made the live fill host-wide.
- Full VM gate: 9/9 existing targets pass with the flag off, plus the
  new target run explicitly (both at this head).

## Follow-ups (out of scope here)

- Memory-recorder stacks keep raw-IP capture in v0; adopting the same
  tail-array pattern there is mechanical once this shape settles.
- An artifact-store fill source (build-id -> symbols service) — the
  store's source ranking already leaves room for it.
- A post-hoc re-resolution pass over emitted traces (the rendered
  frames carry everything it needs).

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,15 @@ This will display all available recorders and their default states:
 
 Python stack symbolization is not a recorder; enable it with `--collect-pystacks`, which resolves Python frames in whichever stacks the active recorders collect.
 
+`--collect-build-id` captures user stacks as (build-id, file offset) pairs
+instead of raw addresses, so frames of processes that exit before end-of-trace
+symbolization still resolve — through a build-id-keyed store filled from
+`/usr/lib/debug/.build-id`, debuginfod (with `--enable-debuginfod`), and the
+binaries of still-running processes. Frames whose build-id no source knows
+render as `unknown ([buildid:<hex>]) <0x<offset>>`, a stable identity that can
+be resolved offline later. Costs ~40% larger stack-ring reservations while
+enabled; off by default and free when off.
+
 Note: `sleep-stacks` acts as a master switch for all sleep stack collection. When enabled,
 both uninterruptible (D state) and interruptible (S state) sleep stacks are collected by
 default. Use `--no-interruptible-stack-traces` or disable the `interruptible-stacks`
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -49,6 +49,22 @@
 #define SKIP_STACK_DEPTH 3
 #define NR_RINGBUFS 8
 
+/* Mirror of the kernel's struct bpf_stack_build_id (filled by bpf_get_stack
+ * under BPF_F_USER_BUILD_ID), with the offset/ip union flattened to one u64
+ * so the generated Rust type has no union member. status says which meaning
+ * the u64 carries: BPF_STACK_BUILD_ID_VALID = file offset (build_id is set),
+ * BPF_STACK_BUILD_ID_IP = raw instruction pointer (the kernel could not
+ * resolve the frame's VMA to a file with a build-id note, or could not walk
+ * safely in this context). */
+struct systing_stack_build_id {
+	s32 status;
+	u8 build_id[20];
+	u64 offset_or_ip;
+};
+_Static_assert(sizeof(struct systing_stack_build_id) ==
+		       sizeof(struct bpf_stack_build_id),
+	       "layout must match the kernel struct bpf_get_stack fills");
+
 #define SYS_ENTER_COOKIE 0xFFFFFFFFFFFFFFFEULL
 #define SYS_EXIT_COOKIE 0xFFFFFFFFFFFFFFFFULL
 
@@ -86,6 +102,14 @@ const volatile struct {
 	u32 no_sched;          /* Sched recorder off: sched_switch still runs for
 				* cpu_running_pid and sleep stacks, but task_event
 				* emission is suppressed. */
+	u32 collect_build_id;  /* User stacks carry (build_id, file offset) pairs
+				* instead of raw IPs (BPF_F_USER_BUILD_ID), and
+				* stack_event reservations include the
+				* user_stack_bid tail. rodata: with the knob off
+				* the verifier prunes the build-id arms entirely,
+				* so the off configuration is instruction- and
+				* reservation-identical to before the knob
+				* existed. */
 	u32 no_irq;            /* IRQ recorder off: the irq/softirq programs are
 				* not loaded at all in that case, so this rodata
 				* gate is belt-and-braces (and lets the verifier
@@ -196,18 +220,34 @@ struct marker_event {
 	u64 info;          // from args[3] (flags param of faccessat2)
 };
 
+/* user_stack_format values: which entry format the user_stack region
+ * carries. The capture mode is fixed for a whole run (rodata), so the
+ * per-event flag is belt-and-braces — it makes every record
+ * self-describing for downstream readers. */
+#define USER_STACK_FORMAT_IP 0
+#define USER_STACK_FORMAT_BUILD_ID 1
+
 struct stack_event {
 	enum stack_event_type stack_event_type;
 	u64 ts;
 	u32 cpu;
+	u32 user_stack_format;
 	struct task_info task;
 	u64 kernel_stack_length;
 	u64 user_stack_length;
 	u64 kernel_stack[MAX_STACK_DEPTH];
-	u64 user_stack[MAX_STACK_DEPTH];
 #ifdef SYSTING_PYSTACKS
 	struct pystacks_message py_msg_buffer;
 #endif
+	/* One region, tail by contract, sized for the larger (build-id) entry
+	 * format. Raw-IP mode packs u64 addresses from the region start and
+	 * uses less of it: reservations stop at
+	 * offsetof(user_stack) + MAX_STACK_DEPTH * sizeof(u64) — byte-for-byte
+	 * what the pre-build-id layout reserved — and userspace zero-extends
+	 * short records (the memory ring's variable-length pattern).
+	 * user_stack_format says how to parse; user_stack_length counts
+	 * entries of that format. */
+	struct systing_stack_build_id user_stack[MAX_STACK_DEPTH];
 };
 
 struct perf_counter_event {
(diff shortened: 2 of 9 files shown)


josefbacik:
## Code review

definition over the full diff at b50a10b (vs base 74165cd, v1.11.32).
An earlier revision of this branch was reviewed the same way; the final
delta (the host-wide live-process index replacing sampled-tgids-only
fills, prompted by the integration test's first VM run) was re-reviewed
against every criterion below before this verdict was finalized.

### Summary

Opt-in build-id capture for user stacks: one rodata-gated BPF arm walks
into a tail array with `BPF_F_USER_BUILD_ID`, and symbolization gains a
trust-ranked build-id store plus a process-independent result cache.
9 files, +1240/−97; flag-off behavior is unchanged by construction
(load-time pruning + prefix-sized reservations, both test-pinned).

### Critical Issues

None.

- Schema & versioning: no table/schema changes (new rendering is frame
  *strings* only), `Cargo.toml` version untouched, so no `[bump minor]`
  token is required.
- Recorder consistency: no new recorders or tables; utid handling
  untouched.
- BPF: probes remain lookup-only — the new arm is a flag change on the
  existing `bpf_get_stack` call into a differently-typed buffer; no map
  operations, ring reserves, or allocations were added to any hot path.
  The extraction work happens inside the kernel helper (trylock-guarded,
  IP fallback), not in program text.
- The two-size reservation copies the memory ring's established
  variable-length pattern, and the zero-extend consumer mirrors
  `create_memory_ring`. `user_stack_bid` tail position is enforced by
  both a `_Static_assert` (layout vs kernel struct) and a Rust layout
  test (tail position + reservation sizes), so the
  `offsetof`-prefix contract cannot rot silently.
- Store trust ranking matches its stated policy in code: trusted
  sources are probed before the live fill is consulted, live fills
  never overwrite anything, and the miss memoization plus entry cap
  (LRU) bound both repeated probing and adversarial unique-id storms.

### Improvements

Addressed during review:
- README gained the `--collect-build-id` paragraph (the flag was
  otherwise undocumented outside `--help`).
- `fetch_debuginfod` documents the zero-padding ambiguity for sub-20-byte
  build-id notes (padded form is queried; such ids miss and fall to the
  deferred rendering).

### Suggestions

- `symbolize_stack_frames` now takes seven parameters; a small context
  struct would tidy the signature. Deliberately left explicit in v0 so
  the data flow (which caches are global vs per-process) stays visible
  at the call sites.
- `BuildIdStore::lookup` clones a `ResolvedBinary` (PathBuf + String)
  per uncached `(id, offset)`; bounded by unique frames and fronted by
  the result cache, so not worth lifetime plumbing.
- The two walk branches in `emit_stack_event_with_ts` duplicate the
  length handling; kept separate so each mode's walk (and the
  prune-at-load reasoning) reads linearly.

### Positive Notes

- The integration test is negative-control-first: it proves the burner
  is sampled, dead by symbolization time, and mode-off is genuinely off
  before asserting the feature works — the control run inherits its
  validity assumptions to the feature run. Its first VM run caught a
  real population gap (a pid-filtered trace has no sampled live
  processes to fill from), which is what moved the live fill from
  sampled-tgids-only to the host-wide index.
- Running the live-process index before any pass keeps the existing
  two-pass structure and its emit-immediately memory behavior for
  exited stacks — no new spill tiers.
- The deferred rendering keeps the full 40-hex id: unresolved frames
  stay resolvable offline, which is the design's actual contract.

josefbacik:
Ack — collapsing to one region: `user_stack` sized for the build-id entries, raw-IP mode packs u64s into the same region and reserves only the bytes it uses (byte-for-byte the old reservation), plus a `user_stack_format` flag on the event so every record is self-describing. Sizes go 2720B/3584B (1.32x, flag-on only). Revision lands after the VM gate re-run.

josefbacik:
Done in e99e507 — one `user_stack` region sized for the build-id entries; raw-IP mode packs u64s into the same region and reserves only `offsetof(user_stack) + depth * 8` (byte-for-byte the old reservation), and each event carries a `user_stack_format` flag so records are self-describing. Reservations are now 2720B raw-IP / 3584B build-id (1.32x, flag-on only); the object-level check still shows zero instruction delta in every program section (the emit helper's `.text` actually shrank vs base). Full VM gate re-run green at this head, including the dead-process e2e; the layout test now also pins the region's u64 alignment, which the packed-IP view depends on.

PR #175, opened 2026-07-24
session_recorder: emit full container IDs, not 12-char prefixes

josefbacik:
`get_container_id` truncates every container ID to its first 12 hex chars, so the `container:<id>` namespace names in `network_interface` (and namespace track names) carry docker-short-id prefixes rather than the runtime's actual identifier. Any consumer that joins those IDs against runtime state keyed by the full 64-hex digest — cgroup scope directories, CRI APIs — can never match a truncated form.

This returns the ID exactly as the runtime wrote it, and factors the cgroup parse out of the `/proc` read into `parse_container_id` so the width behavior is pinned by unit tests across all three runtime formats (containerd, CRI-O, Docker).

No schema change: `network_interface.namespace` stays a string column; only the embedded ID length changes. Tests: 453 passing (`cargo test --lib`) including the two new width-pinning cases; fmt/clippy clean.

### DIFF
--- a/src/session_recorder.rs
+++ b/src/session_recorder.rs
@@ -437,7 +437,18 @@ fn all_proc_pids() -> impl Iterator<Item = u32> {
 fn get_container_id(pid: u32) -> Option<String> {
     let cgroup_path = format!("/proc/{pid}/cgroup");
     let content = fs::read_to_string(&cgroup_path).ok()?;
+    parse_container_id(&content)
+}
 
+/// Extracts a container ID from `/proc/<pid>/cgroup` content.
+///
+/// Returns the ID exactly as the runtime wrote it (the full 64-hex digest
+/// on containerd/CRI-O/Docker) — no truncation. Consumers that join these
+/// IDs against other runtime state (cgroup trees, CRI APIs) need the full
+/// digest; anything display-oriented can shorten at the display layer.
+/// The `>= 12` length floor only guards against matching stray short hex
+/// fragments in unrelated path components.
+fn parse_container_id(content: &str) -> Option<String> {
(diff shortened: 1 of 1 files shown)


josefbacik:
Code review (rust-bpf-code-reviewer, executed over the full diff at b1c30fa; mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition applied in-session, all five passes):

### Summary
Minimal behavioral fix with a testability refactor: `get_container_id` stops truncating container IDs to 12 chars and returns the runtime's full digest; the cgroup-content parse is factored into a pure `parse_container_id` pinned by unit tests across all three runtime formats. The change is small, correct, and self-documenting.

### Critical Issues
None.
- Schema gate: no table or column changes (`network_interface.namespace` remains a string column; only the embedded ID length changes) — no `SCHEMA_VERSION` bump, no `SCHEMA_CHANGES.md` entry, no `[bump minor]` token required.
- `Cargo.toml`/`Cargo.lock` untouched (version-bump workflow owns them).
- No recorder-table/`utid` surface touched.

### Improvements
Addressed in this revision: the parse was fs-coupled and untestable — it is now factored into `parse_container_id` with width-pinning tests. The shipped truncation is exactly the class an untested width invites.

### Suggestions
The three runtime branches (docker / cri-containerd / crio) share their validate-and-return tail and could fold into one helper over `(marker, terminator)` pairs — deferred to keep this fix reviewable; worth doing if a fourth runtime format ever lands.

### Positive Notes
The `>= 12` length floor is retained and its purpose is now documented (guards against stray short hex fragments), so runtimes with shorter-than-64 IDs keep working unchanged.

PR #176, opened 2026-07-25
stack_recorder: name the module on build-id store hits that miss the symbol

josefbacik:
Follow-up to #173. When build-id capture is on and a build-id frame's symbol does not resolve, name the module instead of emitting a bare `[buildid:<hex>]`.

## What changes

A user frame whose mapping carries an ELF build-id note is symbolized through the build-id store rather than the live process. On a store hit the binary path — and therefore its module name — is known. If the file-offset symbolization then fails to name the symbol, the frame was rendered as `[buildid:<hex>]`, dropping the module name the store had already resolved.

The live-process path already renders `<module>:unknown` (e.g. `libc.so.6:unknown`) when it can name the module but not the symbol. This change makes the build-id path do the same: render the module name on a store-hit-symbol-miss, and reserve the bare `[buildid:<hex>]` form for when only the id is known — a full store miss, or a trusted id-only entry (debuginfod / `.build-id` dirs) whose `display_module` is itself an id placeholder.

Resolved frames, full misses, and frame-labels-off output are unchanged.

## Why

Observed when running with build-id capture enabled: frames that the live-process path had named at module level rendered as bare build-id hashes on the build-id path, because that path discarded the module name the store had already resolved on the hit. The module name is the recovered identity here — the hash adds nothing over a name we already have, and the bare-hash form is strictly less useful for reading and aggregating a flame graph.

## Test

The miss-rendering is extracted into `render_unresolved_build_id` as a pure function and unit-tested for the four cases: known module → `<module>:unknown`; trusted id-only entry and full store miss → deferred `[buildid:<hex>]`; frame-labels-off → machine `buildid:<hex>+<off>` form. Full lib suite green (454 tests).

No schema change, so no `SCHEMA_VERSION` / `SCHEMA_CHANGES.md` bump; this is a symbolization-rendering change only.

### DIFF
--- a/src/stack_recorder.rs
+++ b/src/stack_recorder.rs
@@ -1286,14 +1286,16 @@ impl StackRecorder {
                     self.elide_generics,
                 );
             }
+            // Store hit, symbol unresolved: the module is still known.
+            return render_unresolved_build_id(
+                self.frame_labels,
+                Some(&bin.display_module),
+                &id,
+                offset,
+            );
         }
-        if self.frame_labels {
-            format!("unknown ([buildid:{}]) <{offset:#x}>", build_id_hex(&id))
-        } else {
-            // Even with labels off the identity must survive: a bare offset
-            // would masquerade as a virtual address.
-            format!("buildid:{}+{offset:#x}", build_id_hex(&id))
-        }
+        // Full store miss: only the build-id identity is available.
+        render_unresolved_build_id(self.frame_labels, None, &id, offset)
     }
 
     /// Symbolize a single stack and return frame names.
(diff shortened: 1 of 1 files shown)


josefbacik:
## Automated review (rust-bpf-code-reviewer, in-session)

Reviewed the full diff before push, per this repo's review protocol. Mechanics disclosed: the review criteria were applied by the authoring session over the complete `src/stack_recorder.rs` diff — a self-review by the change author; a separate human approval is still required.

### Summary
Small, well-scoped change: extracts the build-id miss-rendering into a pure function and adds a store-hit-symbol-miss branch that names the known module. No BPF, schema, or capture-path changes.

### Critical Issues
None.
- Schema/versioning: no schema touched, so no `SCHEMA_VERSION` / `SCHEMA_CHANGES.md` bump and no `[bump minor]` needed; `Cargo.toml` version untouched (bump-on-merge regime).
- Recorder consistency: no new tables, no `tid`/`utid` changes — N/A.
- BPF: none touched — N/A.

### Improvements
- The store-hit vs id-only distinction is made by `!display_module.starts_with("[buildid:")`, which couples to how the store builds its placeholder module name. A type-level distinction (the store signalling whether an entry is id-only) would be more robust. Noted in the function doc comment; reasonable as a follow-up rather than blocking this change.

### Suggestions
None blocking.

### Positive Notes
- The pure-function extraction makes the four rendering cases unit-testable without constructing a `Symbolizer`; the test covers known-module, id-only, full-miss, and labels-off.
- Behavior is preserved exactly for resolved frames, full misses, and labels-off output; only the store-hit-symbol-miss case with a known module changes.

Gates run before push: `cargo fmt --check` clean, `clippy` clean, full lib suite 454 passed / 0 failed.

PR #177, opened 2026-07-29
network: attribute flows to their network namespace inode [bump minor]

josefbacik:
Adds a `netns_inum` column to `network_socket` and `network_interface` (schema v13): for a flow row, the inode of the network namespace that owns the socket, read in BPF at event emission; for an interface row, the inode of the namespace the interface was enumerated in (the per-namespace walk already computes this inode for its namespace names).

## Why

Flow-to-namespace attribution currently matches the socket's `src_ip` against the per-namespace interface map. That match cannot disambiguate loopback, because every namespace has `127.0.0.1`/`::1` on `lo`, so intra-namespace localhost traffic is unattributable. Inbound flows (whose recorded source is the remote peer) and IPv4-mapped-IPv6 sockets also attribute wrongly or not at all. The namespace inode is unambiguous: a reader joins `network_socket.netns_inum = network_interface.netns_inum` and gets the owning namespace (and hence the container) directly.

## Mechanics

- BPF: a `read_sk_netns_inum()` helper — one null-guarded CO-RE read of `sk->__sk_common.skc_net.net->ns.inum` — stamped at all 16 socket-bearing emission sites (2 network events from `info->sk_ptr`, 14 packet events from `sk`). The 5 packet paths with no socket pointer leave the field 0; both event reserves memset their buffers, so "unknown" is a real 0, never garbage.
- A socket's netns is stable for its lifetime, so reading at emission equals reading at creation. Readers treat 0 as "unattributed" — the pre-change behavior. A wrong attribution is not possible, only a missing one.
- Rust: both records carry `netns_inum` (i64) through models → recorder → arrow schema → parquet writer → DuckDB DDL. `SCHEMA_VERSION` 12→13 with a `SCHEMA_CHANGES.md` entry. `Cargo.toml` untouched (version is assigned at merge).

## Testing

- `cargo build --tests` green — compiles the CO-RE read against vmlinux BTF, which validates the field chain exists with the right types.
- `cargo fmt --check` and `cargo clippy --workspace --no-default-features -- -D warnings`: clean.
- `cargo test --lib`: 454 passed (schema round-trip and fold tests).
- Runtime verification: the VM integration rig was unavailable in this environment, so runtime behavior is validated at first deployment on a multi-namespace node, per a written gate: BPF load-proof on the production kernel; sampled `network_socket.netns_inum` nonzero and distinct across namespaces; join against `network_interface.netns_inum` with ~0 mismatches (the BPF read and a userspace `stat` of `/proc/<pid>/ns/net` return the same nsfs inode by construction); exact match to a known workload's namespace inode; root-namespace sockets carrying the init-ns inode; and the no-socket packet paths reading 0. This PR is write-only — nothing consumes the new column yet — so the deploy-time validation carries no consumer risk. It covers the trace-level tables this PR writes; downstream consumers are validated separately when they land.

### DIFF
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -271,3 +271,28 @@ SCHEMA_VERSION bump); `callsites` tables written by older systing-util
 binaries from spec-compliant inputs carry the inverted parent/depth shape and
 cannot be distinguished in-band — re-import with a fixed binary if the tree
 direction matters.
+
+## Schema Version 13 (systing 1.12.0) — 2026-07-29
+
+`network_socket` and `network_interface` each gain a `netns_inum BIGINT` column:
+the inode of the network namespace that owns the socket (for a flow row) or the
+interface/IP (for an interface row). It lets a reader attribute a flow to its
+owning netns — and therefore its pod — by joining
+`network_socket.netns_inum = network_interface.netns_inum`, instead of matching
+the socket's `src_ip` against the interface map. The IP match cannot
+disambiguate loopback (every netns has `127.0.0.1`/`::1` on `lo`), so
+intra-pod-localhost flows were unattributable before; the netns inode is
+unambiguous. It also attributes inbound flows (whose `src_ip` is the remote
+peer, not a local interface) and IPv4-mapped-IPv6 sockets correctly.
+
+### New columns
+- `network_socket.netns_inum` (BIGINT): read in BPF from the socket's
+  `struct sock` (`sk->__sk_common.skc_net.net->ns.inum`) at send/recv event
+  emission; the netns is stable over the socket's lifetime. `0` when the sock
+  pointer was unavailable.
+- `network_interface.netns_inum` (BIGINT): the inode of the enumerated netns
+  (the same value userspace reads from `/proc/<pid>/ns/net`), emitted alongside
+  the existing `namespace` display string by the per-namespace interface walk.
+
+Pre-v13 databases and traces have neither column; readers join on `netns_inum`
+where present and fall back to the `src_ip`→interface heuristic otherwise.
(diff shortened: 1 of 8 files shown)


josefbacik:
Code review (rust-bpf-code-reviewer, executed over the full diff at 31dea50b; mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition applied in-session, all five passes):

### Summary
Adds namespace attribution to network flows: a `netns_inum` column on `network_socket` and `network_interface`, stamped in BPF via one null-guarded CO-RE read (`sk->__sk_common.skc_net.net->ns.inum`) at every socket-bearing emission site, and from the interface walk's already-computed namespace inode on the interface side. Schema v13. Write-only: no consumer of the column in this PR.

### Critical Issues
None.
- Schema gate PASS: `SCHEMA_VERSION` 12→13, `SCHEMA_CHANGES.md` v13 entry present, `[bump minor]` in the PR title, `Cargo.toml`/`Cargo.lock` untouched (the merge workflow owns the version).
- No `unsafe` added, no new maps; the helper is a straight-line probe read of structs the handlers already touch.
- Fail-safe holds: both event reserves memset their buffers and the 5 packet paths without a socket pointer never set the field, so unknown reads as 0 ("unattributed", the pre-change behavior) — never garbage and never a wrong namespace.

### Improvements
An earlier revision left 2 socket-bearing packet sites (`sk_stream_wait_memory`, `inet_sock_set_state`) unset. A TCP state change is often a socket's first-seen event, so those sockets would have latched netns=0 through the first-seen dedup. Fixed in this revision: all 16 socket-bearing sites now stamp.

### Suggestions
A cached per-socket read (capture once into socket metadata, echo at emission) was considered and rejected: the map lookup costs about what the probe read saves, and a cache is born unpopulated for sockets that predate the trace. The per-emission read is correct and the value is lifetime-stable.

### Positive Notes
The namespace inode (rather than the netns cookie) is the right key: it is the same value nsfs exposes as the st_ino of `/proc/<pid>/ns/net`, so trace rows join directly against the interface walk and against ground truth observable from userspace.

PR #179, opened 2026-08-03
stack_recorder: route live-process build-id frames through the live cascade

josefbacik:
## Problem

With `--collect-build-id`, the kernel stack walker tags every user frame whose mapping carries a build-id note — including frames of processes that are still alive at symbolization time. Those frames bypass the full live symbolization cascade (process memory view, pool-island bridging, guest-maps labels) and take a single-attempt `Elf`/`FileOffset` lookup through the build-id store instead, with no fallback.

Two consequences:

1. **Live-process regression.** Frames the normal live path resolves fine come back unresolved from the store path. In a 50/50 node A/B on a large gVisor-based cluster, the flag-on arm lost ~8 percentage points of leaf real-resolution vs control, concentrated in exactly the binaries you'd expect to resolve (libc, ld-linux, compilers, node/JVM/Python runtimes).
2. **Sandboxed processes can't work at all through this path.** For gVisor systrap processes, guest text lives in pool-memfd islands, so the kernel's vma-arithmetic "file offset" (`ip - vm_start + pgoff<<12`) is island-relative — translating it against the real ELF's program headers cannot succeed. These are precisely the frames the island bridge exists for, and the store path bypasses the bridge.

## Fix

Route build-id frames of a **live** process through the process's own view first: find the file-backed executable mapping that both contains the file offset and carries the frame's build-id note, reconstruct the virtual address (`vm_start + (offset - file_offset_of_mapping)`), and symbolize it through the existing live cascade — which re-engages the memory view, island bridging, and guest labels automatically. The store path remains the fallback when reconstruction declines, and stays the sole path for exited processes — the feature's design case, unchanged.

Reconstruction is conservative by construction:
- memfd and anonymous mappings never match (their offset fields are pool offsets, not ELF file offsets);
- offset ranges of unrelated files overlap freely, so candidates are confirmed by build-id note — an unconfirmable candidate (unreadable link, note mismatch) declines to the store path rather than guessing.

Live-path renderings are process-scoped (bridging and guest labels depend on the process), so they cache per process group; store-path results remain globally cached as before.

Flag-off behavior is untouched: `UserFrame::BuildId` frames only exist in build-id mode.

## Testing

- Unit tests: candidate containment math, anon/memfd exclusion, multi-candidate return, note zero-extend/truncate normalization, and the decline-to-store fallbacks.
- Full suite green locally; fmt + clippy clean.
- Deploy validation: the running A/B is the instrument — the flag-on arm's leaf real-resolution should converge to control, with build-id/module forms remaining only on genuinely-exited frames.

## Out of scope (noted seam)

Island-relative offsets (pool-backed frames of sandboxed processes) still decline to the store path — healing those end-to-end needs the inverse island bridge (island offset → original file offset) applied before store lookup. Follow-up material.

### DIFF
--- a/src/sandbox_maps.rs
+++ b/src/sandbox_maps.rs
@@ -231,6 +231,31 @@ impl ProcessMaps {
             .collect()
     }
 
+    /// File-backed executable mappings whose file-offset range contains
+    /// `file_offset`, as (candidate virtual address, namespace-immune
+    /// `map_files` link) pairs. The candidate address is where the file
+    /// offset is mapped in this process: `start + (file_offset - offset)`.
+    /// Memfd and anonymous entries never match — their offset fields are
+    /// pool offsets, meaningless as ELF file offsets (see [`MapEntry`]).
+    /// Offset ranges of unrelated files overlap freely, so callers must
+    /// disambiguate candidates (by build-id note).
+    pub fn virt_candidates_for_file_offset(&self, file_offset: u64) -> Vec<(u64, PathBuf)> {
+        self.entries
+            .iter()
+            .filter(|e| e.exec && matches!(e.backing, Backing::File(_)))
+            .filter(|e| e.offset <= file_offset && file_offset - e.offset < e.end - e.start)
+            .map(|e| {
+                (
+                    e.start + (file_offset - e.offset),
+                    PathBuf::from(format!(
+                        "/proc/{}/map_files/{:x}-{:x}",
+                        self.tgid, e.start, e.end
+                    )),
+                )
+            })
+            .collect()
+    }
+
     fn entry_for(&self, addr: u64) -> Option<&MapEntry> {
         // Entries are in address order as read from /proc.
         self.entries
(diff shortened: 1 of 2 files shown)


josefbacik:

*Mechanics: this review executed the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff (`src/sandbox_maps.rs` +70, `src/stack_recorder.rs` +152/−29), with the gate battery run locally: `cargo fmt --check` clean, `cargo build --tests` green, `cargo test --lib` 458/458 passed, `cargo clippy --tests` zero warnings.*

### Summary

A conservative, well-scoped fix: build-id frames of live processes are re-located in the process's own maps (candidate mapping must both contain the file offset and carry the matching build-id note) and symbolized through the existing live cascade, with the store path as fallback and as the unchanged sole handler for exited processes. The design declines to the store path on any ambiguity rather than guessing, and the cache scoping split (global store-path results vs process-scoped live-path results) is exactly right and documented at both declaration sites.

### Critical Issues

None.

- **Schema/versioning gates:** no schema change (`src/duckdb.rs` `create_schema()` and `src/trace/schema.rs` untouched) → no `SCHEMA_CHANGES.md` entry, `SCHEMA_VERSION` bump, or `[bump minor]` title token required. `Cargo.toml`/`Cargo.lock` version untouched (diff stat: only the two src files) — compliant with bump-on-merge.
- **Recorder consistency:** not a recorder or table change; no `utid` semantics touched. N/A.
- **BPF-side:** no BPF code in this diff; kernel-side build-id tagging is pre-existing. N/A.
- **Flag-off byte-identity (load-bearing for shipping this without any gating token):** verified by chain — `UserFrame::BuildId` is only constructed when `event.user_stack_format == USER_STACK_FORMAT_BUILD_ID` (stack_recorder.rs:1641, and spill-replay of frames that were format-gated at capture, :555); the BPF side only emits that format in build-id mode; the store fill is gated on `self.build_id_mode` (:1012, default `false`). Therefore with the flag off, the new dispatch arm and `reconstruct_build_id_addr` are unreachable and `virt_candidates_for_file_offset` has no caller on any executed path; the only flag-off deltas are one empty `HashMap::new()` per process group and the behavior-identical `note_to_id` extraction inside the (flag-gated) store fill. Rendering output flag-off is byte-identical.

### Improvements

None required. One consistency point was already handled during review: the `([u8; 20], u64)` key tuple is now the named `BuildIdFrameKey` alias at all five sites (including the two pre-existing ones), and the widened `user` parameter is the documented `LiveUserState` alias — this is what cleared the `clippy::type_complexity` warning the first version of the signature introduced.

### Suggestions

- `reconstruct_build_id_addr` performs `read_elf_build_id` per candidate per unique `(id, offset)` per process group. This is bounded by `live_bid_cache` (declined frames also cache, so no per-occurrence retries) and candidate sets are small, but if symbolization profiles ever show it, a per-link note memo inside the group would collapse the reads to one per mapped file. Defer until measured.
- The imperative loop in `reconstruct_build_id_addr` could be a `find_map`, but the early-return loop reads fine; style only.

### Positive Notes

- Decline-over-guess throughout: memfd/anon mappings never match (their offsets are pool offsets, not ELF file offsets), overlapping candidates must be confirmed by note, and an unreadable `map_files` link declines to the store path — which also makes exit races degrade to exactly the old behavior.
- Tests cover the load-bearing edges: containment math, anon/memfd exclusion on the gVisor stub fixture, multi-candidate overlap, note zero-extend/truncate normalization to the kernel's fixed field, and both decline paths (no maps; unconfirmable candidate).
- The doc comments on `symbolize_stack_frames` and the two cache declarations explain the live/store split and its process-scoping rationale — the next reader gets the design, not just the mechanics.

**Verdict: PASS — no critical issues; ship as-is.**

PR #180, opened 2026-08-04
skills: update trace/analyze docs to match the current CLI and schema

christian-heusel:
Both `SKILL.md` files had drifted from the code. Everything below was verified against `systing --list-recorders` / `--help` (v1.11.37), `src/duckdb.rs`, `src/analyze/`, and `src/mcp.rs`, and the SQL examples were run against a local trace database.

## `skills/systing-trace/SKILL.md`

- **Recorder table was stale**: missing `irq` (on by default), `memory`, `memory-alloc`, and the network recorder split into `network` (connection state only) / `network-syscalls` / `network-packets`. Documented the companion and dependency rules from `process_recorder_options()` — notably that `--add-recorder network` also enables `network-packets` while `--only-recorder network` is state-only.
- Noted that `--output` defaults to `trace.pb`, plus the `.systing` / `.systing.gz` profile-export format.
- Added the flags that were missing entirely: `--sample-freq`, `--perf-counter`, `--cpu-frequency`, `--continuous`, `--stream`, `--collect-build-id`, `--symbolize-names-only`, `--symbolize-elide-generics`, `--no-frame-labels`, `--no-gopclntab`, `--no-gvisor-guest-maps`, `--resolve-addresses`, the `--memory-*` tuning set, and the `--no-sched` / `--no-irq` / `--syscalls` toggles. Options are now grouped by area instead of one flat table.
- Recorded the two constraints the CLI enforces: marker thresholds require `--continuous`, and `--tpu-profile` requires a fixed `--duration`.

## `skills/systing-analyze/SKILL.md`

- **`flamegraph` stack types were wrong**: `uninterruptible` / `interruptible` are rejected by `StackTypeFilter::from_str`; the accepted values are `cpu`, `uninterruptible-sleep`, `interruptible-sleep`, `all-sleep`, `all`. Also `min_samples` → `min_count`, and `top_n` / `start_time` / `end_time` were undocumented.
- **Stack schema was stale**: `stack` holds `frame_ids BIGINT[]` + `leaf_name` (the `frame_names` column now lives in the `stack_frames` view), the join key is `stack.id` rather than `stack_id`, and frames are ordered **root-to-leaf** — so the old `frame_names[1] AS leaf` example named the outermost caller, not the leaf.
- **`sched_slice` has no `end_state_str`**, and `end_state` is a bitmask, so the `end_state = 2` example silently missed compound states. Documented the bits and the `NULL` = preempted case.
- Joins now include `trace_id` throughout, which multi-trace databases require.
- New sections: memory tables (with the `memory_rss.member` encoding, including the synthetic `-1`/`-2` values), counters/slices, `sysinfo.sample_event` / `sample_period` weighting with `cpu_info`, the `systing-analyze` CLI subcommands, and a "memory growing?" investigation pattern.

Docs only — no code or schema changes, so no version bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/systing-analyze/SKILL.md
+++ b/skills/systing-analyze/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: systing-analyze
-description: Analyze a systing trace database (.duckdb). Use when the user asks about a systing trace — flamegraphs, scheduling latency, CPU hotspots, network behavior, off-CPU time, TPU op/metric data, or any question about what's in a trace.duckdb file. Orchestrates the systing-analyze MCP tools (trace_info, query, flamegraph, sched_stats, cpu_stats, network_*).
+description: Analyze a systing trace database (.duckdb). Use when the user asks about a systing trace — flamegraphs, scheduling latency, CPU hotspots, network behavior, memory/allocation activity, off-CPU time, TPU op/metric data, or any question about what's in a trace.duckdb file. Orchestrates the systing-analyze MCP tools (trace_info, query, flamegraph, sched_stats, cpu_stats, network_*).
 ---
 
 # Analyzing systing traces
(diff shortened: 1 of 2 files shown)


josefbacik:
## Review: docs match the current CLI and schema ✅

Recommendation: merge. This is an accurate, well-scoped documentation fix — it corrects several places where the SKILL.md files had drifted from the code, and the load-bearing claims check out against the source.

This review was produced by an automated agent reviewer running this repo's Rust/BPF review rubric over the full diff; a human maintainer should still give it a normal pass.

### What it does

Updates `skills/systing-trace/SKILL.md` and `skills/systing-analyze/SKILL.md` to match the current recorder set, MCP tool surface, and DuckDB schema. Documentation only — no code, schema, or dependency changes.

### Verified against the source

I checked the claims that would most mislead a reader if they were wrong, and they are correct:

- **Stack ordering is root-to-leaf, and `leaf_name` is the innermost frame.** Confirmed in `src/duckdb.rs` (the `stack` table comment: "frame_ids are ordered root-to-leaf (outermost caller first) … leaf_name names the last (innermost) frame") and `src/trace/models.rs` ("leaf_name: Innermost (executing) frame's name — the last entry"). The previous docs' `frame_names[1] AS leaf` was genuinely wrong (that indexes the outermost caller), so this is a real correction, not a cosmetic one.
- **`sched_slice.end_state` is a raw kernel task-state bitmask, and there is no `end_state_str` column.** Confirmed: `end_state INTEGER` in the schema with no string sibling, and the model documents it as "Raw kernel task state value." The advice to test bits (`end_state & 2 != 0`) rather than compare for equality is correct, and the old enum-style docs would have produced wrong results on compound states.
- **Memory schema structure** (`member` as the RSS category with synthetic negative values for periodic `mm_struct` samples) matches `src/memory_recorder.rs` and the `memory_rss` table definition.

### Findings

None blocking. Two small notes:

- The `query` examples all switched to explicit `trace_id`-qualified joins (e.g. `ON s.trace_id = ss.trace_id AND s.id = ss.stack_id`) instead of `USING (stack_id)`. That is the correct and safer form for multi-trace databases — good change, and worth the one-line callout you already added under "Multi-trace databases."
- The exact integer values in the memory `member` mapping and the SQL examples I did not each execute; you noted you ran them against a local trace DB, and the structural claims they rest on are confirmed above.

### Gates

- No `Cargo.toml`/`Cargo.lock` version change, no schema change — nothing that interacts with the version-bump-on-merge workflow. Clean.

PR #181, opened 2026-08-04
cli: add shell completion generation

christian-heusel:
Adds `clap_complete`-backed shell completion scripts for all three binaries.

## What

- `systing --completions <SHELL>` — a flag, since `systing` has a flat CLI.
- `systing-analyze completions <SHELL>` and `systing-util completions <SHELL>` — subcommands, matching those tools' subcommand-based CLIs.

Supported shells: bash, zsh, fish, elvish, powershell.

The `systing` flag is handled at the top of `main()`, before the BPF capability check that re-execs under `systemd-run`, so generating completions never needs privileges.

`scripts/generate-completions.sh` emits bash/zsh/fish files for all three binaries into an output directory (default `./completions`). `SYSTING_BIN_DIR` points it at built artifacts, so distro packagers don't need the binaries on `$PATH`.

README gains a "Shell completions" section with per-user install paths and the packager one-liner.

Version bumped 1.11.37 → 1.11.38 (patch; no schema change).

## Testing

- `cargo build --bins`, `cargo fmt --check`, `cargo clippy --all-targets` — clean
- `cargo test` — run separately by the author
- Generated bash and zsh scripts source without error; `scripts/generate-completions.sh` produces all nine files

## Note for packaging

Completions are generated at runtime from the clap definitions, so they stay in sync automatically — but a package build must execute the binaries to produce them, which needs a workaround when cross-compiling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,6 +873,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.6.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f84a88507dbd05c695f2cb5e8558e747179134005e9893882dec964190ed89"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.5.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3443,6 +3452,7 @@ dependencies = [
  "bitfield",
  "blazesym",
  "clap",
+ "clap_complete",
  "ctrlc",
  "dashmap",
  "debuginfod",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ tonic = { version = "0.14", features = ["transport"] }
 tonic-prost = "0.14"
 tokio-stream = "0.1"
 tonic-reflection = { version = "0.14", default-features = false }
+clap_complete = "4.5"
 
 [lib]
 name = "systing"
--- a/README.md
+++ b/README.md
@@ -31,6 +31,39 @@ This will generate a `trace.pb` file which can be uploaded to a
 `.systing.gz`) writes a lightweight profile export any tool can parse without
 DuckDB — see [docs/PROFILE_EXPORT_FORMAT.md](docs/PROFILE_EXPORT_FORMAT.md).
 
+## Shell completions
+
+All three binaries can print their own completion script. `systing` uses a
+flag, the subcommand-based tools use a `completions` subcommand:
+
+```bash
+systing --completions bash
+systing-analyze completions zsh
+systing-util completions fish
+```
+
+Supported shells: `bash`, `zsh`, `fish`, `elvish`, `powershell`.
+
+To install them for your user (bash and zsh shown):
+
+```bash
+# bash
+mkdir -p ~/.local/share/bash-completion/completions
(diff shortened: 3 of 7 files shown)


josefbacik:
## Review: add shell completion generation ✅ (one change needed before merge)

Recommendation: merge once the version bump is reverted (details below). The feature itself is correct and well-built — completion generation for all three binaries, with the right CLI shape for each and a sensible packager script.

This review was produced by an automated agent reviewer running this repo's Rust/BPF review rubric over the full diff; a human maintainer should still give it a normal pass.

### What it does

Adds `clap_complete`-backed shell completion output: `systing --completions <SHELL>` (a flag, matching systing's flat CLI) and `systing-analyze completions <SHELL>` / `systing-util completions <SHELL>` (subcommands, matching those tools). Ships a `scripts/generate-completions.sh` for packagers and documents installation in the README. Shells: bash, zsh, fish, elvish, powershell.

### One thing to fix before merge

**The PR changes the package version (`1.11.37` → `1.11.38`) in both `Cargo.toml` and `Cargo.lock`.** This repo bumps the version and tags automatically on every merge to main, so a version change carried in a PR collides with that workflow and produces a double bump. Please revert just the two `version = "…"` lines (and the matching `name = "systing" version` line in `Cargo.lock`) back to what main has. The dependency additions in the same files — the new `clap_complete` entry in `[dependencies]` and its `Cargo.lock` package block — are correct and should stay; only the version lines need to come out.

This is a workflow convention that isn't obvious from the outside, so nothing wrong with the change itself — just the one revert.

### What's correct

- **`systing --completions` is handled before the privilege re-exec.** The flag is processed at the top of `main()`, ahead of the BPF capability check that re-execs under `systemd-run`, so generating completions never needs privileges. This is the right placement and matches the PR description.
- **The CLI shape fits each binary**: a flag for the flat `systing` CLI, a `completions` subcommand for the subcommand-based tools. Standard `clap_complete::generate(shell, &mut cmd, name, &mut stdout())` usage in each.
- **`scripts/generate-completions.sh`** uses `set -euo pipefail`, branches correctly between the flag form and the subcommand form, and honors `SYSTING_BIN_DIR` for packagers building from `target/release`. Executable bit is set.
- `clap_complete = "4.5"` resolves to 4.6.8 in the lock, consistent with the workspace's clap 4.x.

### Findings

- **Critical (workflow, not code): the `Cargo.toml`/`Cargo.lock` version bump** — revert as above.
- **Minor (optional): no test coverage for the completion paths.** A tiny smoke test that each binary's `Cli::command()` produces completions without panicking would guard against a future CLI change breaking generation, but this is low-risk plumbing and not a merge blocker.

### Gates

- No schema change (`SCHEMA_VERSION` untouched) — correct, this doesn't touch the trace format.
- The version-bump gate is the one flagged above.

josefbacik:
Re-reviewed at the new head (`2bad661b`): the version change is gone — you reverted just the two `version = "…"` lines and kept the `clap_complete` dependency additions, which is exactly right. That was the only blocker, so from the review side this is good to merge.

The earlier optional note still stands and is still optional: a small smoke test that each binary produces completions without panicking would guard against a future CLI change silently breaking generation, but it's not a merge requirement.

Automated agent re-review, same Rust/BPF rubric as the earlier comment; a human maintainer should still give it a normal pass.

PR #182, opened 2026-08-05
network: keep one socket identity across state transitions

josefbacik:
A socket's opens and closes can land on different flow rows, because the socket identity key is (sk pointer, 4-tuple) and state transitions fire with degraded tuples at both ends of a connection's life: connect-start fires before the ephemeral port is assigned, and transitions to CLOSE observably arrive with the bound port already zeroed — every anonymous close row reads `src_port=0` in production traces. The tuple-keyed lookup therefore mints a second, anonymous `src_port=0` identity for terminal transitions. On a production cluster with aggressive readiness probing, the anonymous rows were 68% of all socket rows and closes exceeded opens by ~500x on the busiest port.

**The fix, two halves.**
1. **BPF**: `inet_sock_set_state` reuses the socket's existing identity via the sk-keyed map (`sk_socket_id_map`) before falling back to tuple-keyed creation, so the whole lifecycle lands on one identity; the fallback still creates an identity for sockets whose first observed event is a state change. When the tracepoint fires with a raw CLOSE — both the real-close path and the TIME_WAIT rewrite destroy this sk — the sk-keyed entry is deleted so a recycled allocation mints a fresh identity. Net map traffic goes down on the common path: one lookup replaces get-or-create's lookup/insert/re-lookup. The timewait kprobes are unaffected: they capture the id at `tcp_time_wait` entry, before `tcp_done` fires this tracepoint.
2. **Recorder**: a socket created at connect-start emits its first record with a port-less tuple, so the recorder now emits ONE upgraded record when a later event reveals the bound tuple (tracked by flipping the seen-set to a map of socket_id → carried-a-bound-port). Readers keeping one row per socket prefer the row with the nonzero port; the `NetworkSocketRecord` doc states the contract.

**Disclosed behavior notes.**
- An application that retries `connect()` on the same failed socket now merges its attempts onto one flow row (the tuple-keyed metadata entry from the first attempt matches the retry's identical pre-bind tuple): counts accumulate on one row rather than spreading across rows, which is the desired aggregate for re-dialers. Sockets recycled by the allocator get fresh identities via the delete-at-destroy, and a socket that merges into a persistent tuple entry through client-port reuse is re-seeded into the sk-keyed map (guarded on a bound port) so its terminal transitions still resolve to the merged identity.
- The sk-keyed map has no UDP destroy hook, so dead UDP socket entries persist for the lifetime of a capture (bounded: ~10s continuous, minutes for on-demand); a very UDP-heavy node could fill the map's 10240 slots during a long on-demand trace, degrading TCP seeding to the fallback path. A `udp_destroy_sock` hook or an LRU map is the follow-up if that shows up in practice.
- A connect-start that hits a predecessor's persistent port-less tuple entry attributes that single event to the predecessor's identity (one polluted event, invisible to open/close accounting) — the bound-port guard on hit-path seeding keeps the pollution to that one event rather than the socket's whole lifecycle.

**Validation.** `cargo test` full suite green (461 lib tests including three new emission-lifecycle tests; integration suites green), fmt and clippy clean, and the BPF change compiled through skeleton generation against real vmlinux BTF (straight-line map operations, no new loops). Runtime behavior is validated at the first deployment the same way this codebase's recent BPF changes were: state-change rows carrying stable socket identities end-to-end — opens and closes joining on one `socket_id` with a bound tuple — checked on real traffic before any consumer depends on it.

No schema change: table shapes are untouched (`SCHEMA_VERSION` unchanged); the upgrade record is a content-level contract documented on the record type.

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -1103,8 +1103,22 @@ static __always_inline u64 get_or_create_socket_id(
 
 	// Check if we already have metadata for this socket
 	existing = bpf_map_lookup_elem(&socket_metadata_map, &identity_key);
-	if (existing)
+	if (existing) {
+		// Seed the sk-keyed map on the hit path too: a socket that
+		// merges into a persistent tuple entry (client port reuse
+		// within the capture) must still resolve sk-first at its
+		// terminal transitions, or its close falls back to the
+		// anonymous degraded-tuple identity. Guarded on a bound
+		// src_port — seeding from a port-less connect-start hit
+		// would pin the socket to a shared per-destination port-0
+		// entry for its whole lifecycle.
+		if (src_port != 0) {
+			u64 sk_ptr = (u64)sk;
+			bpf_map_update_elem(&sk_socket_id_map, &sk_ptr,
+					    &existing->socket_id, BPF_ANY);
+		}
 		return existing->socket_id;
+	}
 
 	// Create new socket metadata
 	metadata.socket_id = generate_socket_id();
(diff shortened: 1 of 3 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** (executed in-session over the full diff per the repo's review protocol; author and reviewer are the same session — the review ran the definition's passes rigorously and its mechanics are disclosed here).

### Summary
Two coordinated halves fixing the split-identity defect: the BPF state-change handler now reuses a socket's existing identity via the sk-keyed map before falling back to tuple-keyed creation, and deletes the sk-keyed entry when the socket is destroyed; the recorder emits one upgraded socket record when a socket first seen pre-bind later reveals its bound tuple. Minimal, well-commented, and the emission decision is factored into a pure, unit-tested function.

### Critical Issues
None.
- Schema gate: `src/duckdb.rs` and `src/trace/schema.rs` untouched — no schema change, so no SCHEMA_VERSION bump and no `[bump minor]` required. `Cargo.toml`/`Cargo.lock` untouched (verified via git status).
- Recorder-consistency (utid rules): not applicable — no per-thread attribution surface changes.

### Improvements
- Same-socket reconnects (an application calling connect() again on a socket that failed) now merge onto one flow row: the tuple-keyed metadata entry from the first attempt persists and matches the identical pre-bind tuple of the retry. This is disclosed in the PR body as intended aggregate behavior (a re-dialer's attempts belong on one row with accumulating counts); a full per-lifecycle split would need the metadata entry deleted at destroy time, which requires carrying the creation tuple and is not worth the map traffic today.

### Suggestions
None material.

### Positive Notes
- The sk-first lookup REDUCES hot-path map traffic for known sockets (one lookup replaces get-or-create's lookup-insert-relookup sequence).
- The delete keys on the raw tracepoint `newstate` rather than the rewritten event state, with the two-destruction-paths reasoning stated where it matters; ordering against the timewait kprobe (which captures the id at `tcp_time_wait` entry, before `tcp_done` fires this tracepoint) was verified in-source.
- `should_emit_socket_record` as a pure seam with exhaustive-by-construction match and three lifecycle tests is exactly the right testable shape.

Gates at review time: `cargo build --tests` green (BPF skel-gen compiled the modified program against real vmlinux BTF), `cargo test --lib` 461/461, `cargo fmt --check` clean, `cargo clippy --workspace --no-default-features -D warnings` clean.

**Post-review addendum (delta at `eb368f30`):** the second-reviewer pass (BPF owner) caught that the metadata-hit path never seeded the sk-keyed map, leaving the tuple-reuse class's closes on anonymous identities — fixed with a bound-port-guarded seed-on-hit (the guard prevents pinning connect-starts to shared port-less entries). The kernel-ordering parenthetical was softened to the observed fact. Verdict unchanged: PASS, no Critical issues.

PR #183, opened 2026-08-05
stack_recorder: teach build-id live dispatch about gVisor pool mappings

josefbacik:
## What

#179 routes live-process build-id frames through the live symbolization cascade by locating the frame's (build-id, file offset) in the process's own file-backed executable mappings. That search deliberately matches only file-backed mappings, so a frame whose offset is not a real ELF file offset declines to the store path. gVisor guest frames are exactly that case, and the store can't save them: under systrap the kernel's build-id walker can report a guest frame against the `runsc-memory` pool memfd with a pool-relative offset, so the store looks up the right ELF at a meaningless offset and renders `module:unknown`.

This PR adds a pool-aware second search. When no file-backed mapping matches, executable memfd mappings whose offset range contains the frame's offset yield candidate addresses. A candidate is accepted only if the island bridge (#137) maps it to a congruent file-backed neighbour AND that neighbour's own ELF build-id note matches the frame's build-id. Accepted frames route through `symbolize_user_addr` — the same live cascade that already resolves these frames when build-id mode is off. Anything that doesn't bridge, or bridges to a binary whose note mismatches, declines to the store path exactly as before.

## Why the offsets are pool-relative

Verified on stock runsc (release-20260622.0) in a VM: guest text is mostly file-backed (donated FDs, real file offsets — those resolve fine through #179), but systrap patches syscall sites through 4KB `runsc-memory` memfd islands interleaved in guest text. The islands carry a copy of the guest ELF — opening an island via `map_files` returns ELF magic — so the kernel's build-id read on an island frame finds the binary's real note, while the island vma's pgoff is pool-relative (e.g. `0x3fe3e000` into the pool, not into the ELF). The result is `BuildId { id: <real note>, offset: <pool offset> }`: the file-candidate search cannot match it by construction, and the store hits the right module but misses the symbol.

On a gVisor-based Kubernetes cluster this renders the guest application set (compilers, node, libpython, libjvm — binaries that resolve correctly with the flag off) as `module:unknown` in build-id mode, leaving the treated arm ~6pp worse in leaf resolution than control in a node-level A/B.

## Design constraints

- Flag-off behavior is untouched: `UserFrame::BuildId` frames only exist in build-id mode, and nothing outside that path changed.
- The store path stays the sole handler for exited processes — this only adds a live-process recovery before falling back.
- Decline-safe: the pool path can never resolve a frame against an unconfirmed binary. The bridged neighbour's ELF note is the authority; the memfd's own note is a patched copy and is never trusted.
- `virt_candidates_for_file_offset` keeps its file-backed contract; the pool search is a separate method so non-build-id callers keep clean semantics.

## Open question (bounded)

On stock runsc the islands are ~2% of guest text — enough to prove the mechanism, but possibly not the whole gap if a managed-Kubernetes runsc build pool-backs more guest text than stock. The search handles both cases identically (any executable pool mapping containing the offset is a candidate), so this changes nothing about the fix; a post-deploy A/B on the same cluster will show whether any residual class remains.

## Testing

- Unit fixtures for the pool search: containment math against island mappings, non-executable pool mappings excluded, real ELF file offsets never claimed by the pool path.
- Decline-safety: a pool candidate that bridges but whose neighbour note cannot be verified falls back to the store path (asserted with a fixture whose bridge target cannot exist).
- The positive resolution path needs a live runsc guest, so it is exercised by the deploy A/B rather than a unit fixture.
- cargo fmt / build / clippy (`-D warnings`) / test green locally: 459 passed / 0 failed in the main suite; the kernel-requiring integration tests are environment-ignored locally and covered by CI.

### DIFF
--- a/src/sandbox_maps.rs
+++ b/src/sandbox_maps.rs
@@ -256,6 +256,29 @@ impl ProcessMaps {
             .collect()
     }
 
+    /// Virtual addresses of executable memfd mappings whose offset range
+    /// contains `offset`. Exists for gVisor's `runsc-memory` pool: under
+    /// systrap the kernel's build-id walker reports a guest island frame's
+    /// offset relative to the pool memfd (not the original ELF), so the
+    /// `virt_candidates` lookup finds nothing; the address
+    /// `start + (offset - e.offset)` is where that pool offset is mapped,
+    /// which is an island inside the guest image. Deliberately name-agnostic
+    /// (any memfd, matching [`Self::bridge_for`]'s scope, not
+    /// `GVISOR_MEMFDS`): correctness comes from the caller bridging each
+    /// candidate and verifying the bridged neighbor's build-id note, which
+    /// makes a coincidental match on some other runtime's exec memfd
+    /// harmless. Returns addresses to bridge, never (offset, link) pairs —
+    /// the pool offset itself is meaningless as an ELF file offset, and only
+    /// the island's file-backed neighbors identify the real binary.
+    pub fn pool_candidates_for_offset(&self, offset: u64) -> Vec<u64> {
+        self.entries
+            .iter()
+            .filter(|e| e.exec && matches!(e.backing, Backing::Memfd(_)))
+            .filter(|e| e.offset <= offset && offset - e.offset < e.end - e.start)
+            .map(|e| e.start + (offset - e.offset))
+            .collect()
+    }
+
     fn entry_for(&self, addr: u64) -> Option<&MapEntry> {
         // Entries are in address order as read from /proc.
         self.entries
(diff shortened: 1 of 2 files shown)


josefbacik:
Review performed per `.claude/agents/rust-bpf-code-reviewer.md` (in-repo definition, blob `f09e9b6`) over the full diff of this PR.

### Summary

The change extends the build-id live dispatch (`reconstruct_build_id_addr`) with a pool-memfd candidate search so gVisor guest frames route through the live symbolization cascade instead of failing to the store path on pool-relative offsets. It is small, decline-safe by construction, mirrors the existing candidate-search idiom, and touches no schema, version, recorder, or BPF-side surface.

### Critical Issues

None found. Checked explicitly:
- No `Cargo.toml`/`Cargo.lock` version modification (bump-on-merge preserved).
- No schema change — symbolization-only, so no `SCHEMA_VERSION`/`SCHEMA_CHANGES.md`/`[bump minor]` obligations attach.
- No new `unsafe`.
- No recorder surface (no `utid`/attribution concerns).

### Improvements

- Considered and rejected during review: name-filtering pool candidates to `GVISOR_MEMFDS`. `bridge_for`'s memfd scope is deliberately name-agnostic, and the bridged neighbour's build-id-note equality is the correctness authority — filtering by memfd name would split one mechanism across two scopes and add gVisor-version brittleness for no safety gain. The method doc states this so a future reader doesn't "tighten" it into a divergence.

### Suggestions

- None material. The nested `if let Ok(Some(..))` + note-equality shape in the pool loop intentionally matches the existing file-candidate loop rather than restructuring both.

### Positive Notes

- The pool path cannot resolve a frame against an unconfirmed binary: acceptance requires bridge congruence AND the bridged neighbour's ELF note matching the frame's build-id, and the unit fixture asserts the decline path when note verification is impossible.
- Doc comments carry the two non-obvious constraints (a pool offset is not an ELF file offset; the memfd's own note is a patched copy and untrusted) instead of restating the code.

Gates: cargo fmt --check / build --tests / clippy --tests `-D warnings` / cargo test all green at the reviewed head — 459 passed / 0 failed in the main suite (my four new fixtures included), kernel-requiring integration tests environment-ignored locally (CI covers them). Bound to the committed text on the v1.11.38 base.

PR #185, opened 2026-08-06
Gate munmap memory events on syscall success

josefbacik:
The munmap leg of the memory recorder emitted its event at `sys_enter_munmap` with `size` taken from the raw `len` argument — before the kernel validated anything. A failed `munmap(addr, bogus_len)` unmaps nothing and returns EINVAL, but the recorder published a row carrying the bogus length as the unmap size. In production this shows up as address-magnitude `size` values (callers passing pointer-scale values in the length position — above 2^60 in the worst rows), frequent enough on some hosts that downstream 64-bit aggregation overflows.

mmap and brk already pair enter→exit through the per-thread scratch map and drop failed calls. This converts munmap to the same shape: enter stashes `{addr, len}`, exit emits only when the syscall returned 0, and `systing_munmap_exit` joins the memory recorder's program enable list. Successful unmaps produce identical rows to before (same fields, captured at exit within the same syscall context); failed calls now produce nothing, which is what actually happened.

One captured-stack limitation, documented at the emit site: a thread that unmaps its OWN stack (musl's thread-exit path) reaches sys_exit with those pages gone, so the user stack reads empty for exactly that row class where enter-time capture had frames. The size field — what aggregations consume — comes from the stash and stays correct, and stashing full stacks per thread would multiply the scratch map's footprint by the stack-buffer size, so the narrow attribution loss is accepted.

Tested in the vmtest VM (v6.12-bpf kernel): `memory_recorder` and `trace_validation` suites green.

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -394,7 +394,7 @@ struct epoll_event_bpf {
 enum memory_event_type {
 	MEMORY_RSS_STAT,      // kmem:rss_stat - per-mm RSS counter change
 	MEMORY_MMAP,          // sys_exit_mmap - new VM area
-	MEMORY_MUNMAP,        // sys_enter_munmap - VM area removed
+	MEMORY_MUNMAP,        // sys_exit_munmap (success-gated) - VM area removed
 	MEMORY_BRK,           // sys_exit_brk - heap boundary change
 	MEMORY_PAGE_FAULT,    // exceptions:page_fault_user - demand fault (sampled)
 	MEMORY_MM_SAMPLE,     // periodic mm_struct snapshot (from perf_event_clock)
@@ -959,7 +959,7 @@ struct {
 	},
 };
 
-/* Per-thread scratch map for pairing mmap/brk enter→exit. */
+/* Per-thread scratch map for pairing mmap/munmap/brk enter→exit. */
 struct memory_syscall_args {
 	u64 addr;
 	u64 size;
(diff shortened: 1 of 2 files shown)


josefbacik:
Review per the repo's `rust-bpf-code-reviewer` agent definition, executed over the full diff before commit (mechanics: the definition's checklist applied in-session; findings below).

**Verdict: PASS — no critical issues.**

- **Pattern fidelity:** the enter-stash/exit-emit conversion mirrors `systing_mmap_enter`/`systing_mmap_exit` line-for-line (same scratch map, same tgidpid key, same lookup-copy-delete sequence, same missing-stash early-return); the success gate (`ctx->ret != 0`) is strictly simpler than mmap's MAP_FAILED window check and matches munmap's 0-or-negative-errno contract.
- **Scratch-map semantics:** `BPF_ANY` overwrite per thread is the existing sharing model (mmap/brk already coexist on this map; syscalls cannot nest within a thread at these tracepoints), and the delete-after-copy prevents stale reuse.
- **Verifier surface:** the new exit program's shape (map lookup, fixed-size copy, ringbuf reserve/submit, the shared `memory_capture_stack` inline) is byte-pattern-identical to the already-verified `systing_mmap_exit`; no new helpers, no new map types.
- **Enable list:** `systing_munmap_exit` added to `MEMORY_BPF_PROGRAMS` — without it the program would silently never attach and munmap events would stop entirely (the enter handler no longer emits); the integration suite's munmap coverage is the against-silence control.
- **Schema:** no field added/removed/retyped; successful-unmap rows are byte-identical to before. No `SCHEMA_VERSION` bump required. `Cargo.toml` untouched (bump-on-merge).
- **Event-loss consideration:** failed munmaps no longer emit — deliberate and correct (nothing was unmapped); a scratch-map fill (65536 entries) drops pairing the same way mmap/brk already accept.
- **Tests:** memory_recorder + trace_validation suites in the vmtest VM (numbers in the PR body); host build/fmt/clippy clean.

PR #186, opened 2026-08-06
Use per-attach-point kernel stack skip; perf-event samples skip zero

josefbacik:
Kernel "leaves" in sampled CPU profiles are currently the third-level caller of what actually executed: the shared `SKIP_STACK_DEPTH = 3` is right for tracepoint-attached captures, where it skips BPF/tracing entry glue, but wrong for the perf-event sampler, whose stack walk starts at the sample's interrupted registers with no glue to skip. This PR passes the skip per attach point: 0 on the perf-event path, the unchanged 3 everywhere else.

Evidence from production data (2-hour windows):

- Pure-leaf kernel functions (`clear_page_erms`, `rep_movs_alternative`, `pv_native_safe_halt`, qspinlock slowpath) appear as CPU-sample kernel leaves in 48,886 of 2.39 billion rows — 0.002%, effectively never. They cannot all be absent; the skip was consuming them.
- The 3-frame shift reconciled an active incident's numbers exactly: lock slowpath spin sits three frames below each lock site, so spin rendered as the lock sites' callers (`dd_dispatch_request`'s slowpath spin as `__blk_mq_sched_dispatch_requests`, and the same pattern at the insert and merge sites).

Deliberately out of scope: the tracepoint-side skip stays 3. Sleep-stack data shows the glue depth above sched tracepoints varies by kernel version (two distinct residual-glue leaf patterns in production), so no constant is right on every kernel; that residue is handled at fold time by name, version-independently, on the consumer side.

Consumer note: leaf-keyed pattern matching downstream is calibrated against today's shifted data, so kernel-leaf spectra change where this deploys and leaf-prefix detectors get re-keyed alongside; that work is tracked and lands with or immediately behind the deployment.

Tested: cargo check/clippy/fmt clean; `trace_validation` and `perf_counter` integration suites green in the vmtest VM (v6.12-bpf kernel). Post-deploy verification: pure-leaf functions appear as leaves, and the leaf spectrum agrees with an interrupted-IP sampler over the same window.

Mechanism-precedent check: the only prior change to skip semantics is #184 (stop capturing kernel stacks on fault events — the same family, a shared constant applied to a context it does not fit); no per-attach-point differentiation previously existed.

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -46,7 +46,18 @@
 #define PF_KTHREAD 0x00200000
 
 #define MAX_STACK_DEPTH 36
+/* Kernel-stack frames to skip on TRACEPOINT-attached captures, covering
+ * the BPF/tracing entry glue between the traced site and the program
+ * (bpf_trace_run*, __traceiter_*). The exact glue depth varies by
+ * kernel version, so some kernels leak residual glue frames past this
+ * skip; fold-side leaf handling owns that remainder. Perf-event
+ * sampling captures must NOT use this: there the walk starts at the
+ * sample's interrupted registers -- no glue exists, and a nonzero skip
+ * discards the interrupted IP plus real return addresses (every kernel
+ * "leaf" becomes its third-level caller).
+ */
 #define SKIP_STACK_DEPTH 3
+#define PERF_EVENT_SKIP_STACK_DEPTH 0
 #define NR_RINGBUFS 8
 
 /* Mirror of the kernel's struct bpf_stack_build_id (filled by bpf_get_stack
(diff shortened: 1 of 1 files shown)


josefbacik:
rust-bpf-code-reviewer review — repo protocol: the definition at `.claude/agents/rust-bpf-code-reviewer.md` executed over the full diff; mechanics disclosed: run in-session by the authoring seat, all five passes, prior to commit.

### Summary
Single-purpose BPF change parameterizing the kernel-stack skip per attach point. Minimal, verifier-safe (`bpf_get_stack`'s flags argument accepts a non-constant value), and zero runtime cost (per-site constants). Correctly scoped: only the unambiguous perf-event path changes behavior.

### Critical Issues
None. Schema untouched (`stack_event` layout unchanged, so no SCHEMA_VERSION / `[bump minor]` gate applies); `Cargo.toml` untouched per the bump-on-merge workflow; no recorder or table changes.

### Improvements
None material. An enum wrapping the two skip values would be over-abstraction for two constants.

### Suggestions
None.

### Positive Notes
The define-site comment documents the tracepoint-versus-perf semantic split at the source of truth, including the version-varying glue caveat that justifies leaving the tracepoint constant alone; the perf-site comment records the exact hazard of reintroducing a skip there. Both are why-comments the code cannot express itself.

PR #187, opened 2026-08-06
Sample mmap/munmap/brk memory events with a per-type 1:N knob

josefbacik:
The memory recorder's map legs (mmap/munmap/brk) record every call, on the assumption that map-call rates are low. A production host broke that assumption: a Python workload unmapping ~90 times a second put one node at the top of the fleet's memory-row emitters (~90x the cluster median), and the resulting volume OOM-looped the capture slice — the memory cgroup killed the recorder 12 times in 20 hours, each time while the capture's database-generation phase peaked on top of the volume-proportional baseline. Faults already sample (`memory_fault_sample_rate`, default 97); this gives the map legs the same knob.

`memory_map_sample_rate` defaults to 1 — record everything, exactly today's behavior — so nothing changes until it is set.

Two design choices worth the reviewer's attention:

- The sample decision is made at exit, not enter. Every enter still stashes its args and every exit still consumes and deletes them, so the enter/exit scratch pairing is untouched — there is no path where a sampled-out enter leaves a stash for a later exit of a different syscall to misread. The decision also lands on the successful-event stream (after the failure gate), so the rate applies to real events, and the gated work is the expensive part: the per-event stack capture and the ringbuf emit.
- Each of the three event types has its own per-CPU counter. A single shared counter would correlate the streams: a workload doing a periodic mmap/munmap pair against a modulo rate could sample one type almost exclusively and starve the other. Per-type counters make each an independent 1:N sample.

Byte aggregates over sampled events scale by the rate; per-event address pairing (an mmap with its own munmap) is unreliable at rates above 1 — the same documented tradeoff as the alloc knob. SCHEMA_CHANGES.md's `memory_map` entry documents this alongside the sibling knobs.

Tested in the vmtest VM (v6.12-bpf kernel): the memory_recorder suite is green. The default path (sampling off) is byte-for-byte today's emission behavior, which is what every existing test exercises; the enabled path is per-CPU 1:N modulo counting, the same shape as the fault and alloc knobs.

Mechanism precedent: the fault knob and the alloc knob are this exact pattern; this is the third leg of the family, closing the one remaining unsampled memory-event class.


josefbacik:
rust-bpf-code-reviewer review — repo protocol: the definition at `.claude/agents/rust-bpf-code-reviewer.md` executed over the full diff; mechanics disclosed: run in-session by the authoring seat, all five passes, prior to commit. Two review-bot rounds were absorbed before this verdict (noted below).

### Summary
Third instance of the established memory-sampling-knob pattern (fault, alloc, now map events): a rodata rate field, per-CPU counter maps, a short-circuiting sample helper, and the gate on the successful-event stream. Default 1 preserves today's behavior exactly and touches no map on that path. The sampling decision lives at exit, and each of the three event types has its own counter.

### Critical Issues
None. No schema change — row shapes are untouched; the knob changes only how many rows exist — so no SCHEMA_VERSION / `[bump minor]` gate applies, and `Cargo.toml` is correctly untouched (the bump-on-merge workflow owns the version and tag). No new recorder or table, so the utid-consistency criteria do not apply. The SCHEMA_CHANGES.md `memory_map` entry is updated to document the sampling alongside its fault/alloc siblings.

### Improvements
None material. A generic helper shared across the three knobs, and a statement-expression macro for the per-type gate, were both considered and declined: the macro obscures the BPF codegen for no real dedup, and the per-site `rate > 1` guard is the same shape the fault knob already uses.

### Suggestions
None.

### Positive Notes
Two decisions are the right ones and worth a reviewer's confirmation. The gate is at exit, not enter: every enter still stashes and every exit still consumes and deletes, so a sampled event cannot leave a scratch stash for a later exit of a different syscall to misread — the enter/exit pairing invariant holds by construction. And each event type has its own counter: a single shared counter would correlate the three streams, so a periodic mmap/munmap pair against a modulo rate could sample one type near-exclusively and starve the other. The helper comment records the broken assumption (low map rates) with the observed counterexample.

### Review-bot rounds absorbed
Round 1: enter-side placement risked a sampled-out enter leaving a stash for a later exit to misread (moved the gate to exit, dissolving it), and a single shared counter correlated the three types (split to per-type counters). Round 2: the counter map was looked up even at the default rate (guarded each lookup behind `rate > 1`, matching the fault pattern), and a SCHEMA_CHANGES sentence overstated that a specific mmap/munmap could not both be recorded (corrected). The recurring version-bump finding is a false positive against this repo's bump-on-merge workflow.

PR #188, opened 2026-08-06
Pin clap_complete to 4.6.1

josefbacik:
The ^4.5 requirement resolves to 4.6.8, but downstream consumers build hermetically against a curated crate mirror that currently carries 4.6.1, so shell-completion builds 404 there. The 4.6.2-4.6.8 delta is immaterial for the completions systing generates.

Side effect of re-resolution: two dependency-list entries (socket2, tempfile) now reference the older windows-sys already present in the lock; both windows-sys versions remain resolved, and neither affects Linux builds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -874,9 +874,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.6.8"
+version = "4.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f84a88507dbd05c695f2cb5e8558e747179134005e9893882dec964190ed89"
+checksum = "406e68b4de5c59cfb8f750a7cbd4d31ae153788b8352167c1e5f4fc26e8c91e9"
 dependencies = [
  "clap",
 ]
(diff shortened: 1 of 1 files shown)


josefbacik:
Code review (rust-bpf-code-reviewer protocol over the full diff at 8cbc426d):

### Summary
Lock-only dependency pin: clap_complete 4.6.8 → 4.6.1, staying inside the existing ^4.5 requirement. No source, schema, or build-system changes; the diff is four lines each way in Cargo.lock.

### Critical Issues
None. The one criterion with teeth for a lockfile diff — the prohibition on PRs touching systing's own version in Cargo.toml or its Cargo.lock entry (the version-bump workflow owns those) — passes: systing's stanza is untouched. Only clap_complete's version and checksum move, plus two dependency-list references (socket2, tempfile) re-pointing to the older windows-sys that was already resolved in the lock; both windows-sys versions remain present, and neither affects Linux builds.

### Improvements
None applicable — no code changed.

### Suggestions
When the downstream mirror catches up to a newer clap_complete, this pin drops with a plain `cargo update -p clap_complete`; nothing else anchors it.

### Positive Notes
The commit message states the re-resolution side effect explicitly instead of leaving it for the reviewer to discover in the diff.

Review mechanics: the repo's rust-bpf-code-reviewer agent definition executed in full over the diff by the systing deployment automation; the Rust, BPF, schema/versioning, recorder-consistency, and Python-version criteria are vacuously satisfied (no such files touched), and the version-line criterion was checked affirmatively. 4.6.1 verified downloadable from crates.io with the checksum cargo resolved from the registry index.

PR #189, opened 2026-08-07
Bound recorded munmap size by the process mapped size

josefbacik:
The memory recorder's `munmap` events record the syscall's length argument as the unmap size, but that is the requested range, not the amount actually unmapped. The kernel removes whatever is mapped in `[addr, addr+len)` and returns 0 even where nothing is mapped, so a caller that unmaps an oversized range of a sparse address space records a size far larger than anything that could have been freed. gVisor sentry processes do exactly this: on a production cluster they recorded munmaps up to 17.59 TB against a process virtual size of at most 7.97 TB (p99 5.8 GB) — physically impossible, since a process cannot unmap more virtual memory than it has mapped — and these values sum to petabytes of "unmapping" per hour in the churn aggregate.

This reads the process's mapped size (`total_vm`) at munmap enter, stashes it alongside the existing `{addr, len}`, and at exit bounds the recorded size by it: a munmap cannot free more than the process had mapped. Impossible values collapse to the process's total mapped size (an upper bound — no munmap can remove more than the process has mapped); real unmaps, which are within the mapped size by definition, are unchanged.

Scope: `munmap` only. `mmap` legitimately reserves more virtual space than the process currently holds (that is what mmap does), so there is no equivalent enter-time bound for it; large mmap sizes are real VA reservation and are a display-labeling matter for consumers, not a recorder bug. This composes with the existing success gate (failed calls emit nothing) — that handles one garbage class, and this handles the successful-but-oversized class the success gate cannot see because the call genuinely succeeded.

The size stays a VA-range figure, not resident/physical bytes; a consumer that needs memory-actually-freed needs a different signal (an RSS delta), which is a separate metric question.

Tested in the vmtest VM (v6.12-bpf kernel): the memory_recorder suite is green. The enter-time `mm` field read mirrors the existing `brk` enter path (which already reads `mm->brk`), so the hot-path cost is one additional field read per munmap on traced tasks.


josefbacik:
rust-bpf-code-reviewer review — repo protocol: the definition at `.claude/agents/rust-bpf-code-reviewer.md` executed over the full diff; mechanics disclosed: run in-session by the authoring seat, all five passes, prior to commit.

### Summary
Bounds the recorded `munmap` size by the process's mapped size, read at enter and applied at exit. Single-purpose, minimal: one added `u64` on the shared enter→exit scratch struct, one `mm->total_vm` read at enter, one comparison at exit. Fixes a class of physically-impossible sizes (unmap larger than the whole process) that the existing success gate cannot catch, because these calls genuinely succeed.

### Critical Issues
None. No schema change — the `memory_map` row shape is unchanged; only the `size` field's value is bounded for `munmap` — so no SCHEMA_VERSION / `[bump minor]` gate applies, and `Cargo.toml` is correctly untouched (bump-on-merge owns the version). No new recorder or table. SCHEMA_CHANGES.md's `memory_map` entry documents the refined `munmap` size semantics.

### Improvements
None material.

### Suggestions
None. The bound is deliberately `munmap`-only: `mmap` legitimately reserves more than the current mapped size, so no enter-time analog exists, and large `mmap` values are real VA reservation (a consumer-labeling matter, not a recorder bug).

### Positive Notes
The `mm->total_vm` read at enter mirrors the existing `brk` enter path's `mm->brk` read, so the pattern and its hot-path cost are already precedented in this file — no new BPF idiom. Reading `total_vm` at enter (pre-unmap) is the correct choice: at exit the pages are already gone, so the exit-time value would under-bound. The `vm_limit && size > vm_limit` guard both avoids zeroing the size on the kernel-thread edge (total_vm reads 0) and leaves every within-bounds unmap byte-for-byte unchanged, so only impossible values move. The comment states the requested-vs-freed distinction the code cannot express itself.

### Hazard-precedent
The change adds one per-munmap `mm` field read on traced tasks. Precedent: `systing_brk_enter` already reads `mm->brk` at the same point in the same file — identical shape, so no new hot-path scanning class is introduced.

josefbacik:
On the version-bump review note: this is intentional and correct as-is. This repo bumps the patch version and tags on every merge to main via the bump-on-merge workflow (#154), so PRs deliberately do not touch `Cargo.toml` — a manual bump here would collide with the merge-time bump. The CLAUDE.md "bump the patch version" line predates that workflow. No schema change either, so no SCHEMA_VERSION increment.

The docstring-drift note was valid and is fixed in the latest push: `MemoryMapRecord` in `models.rs` now documents that for munmap, `size` is the requested length bounded by the process's total mapped size — an upper bound, not necessarily a region of that size at `addr` — matching the SCHEMA_CHANGES.md entry.

PR #190, opened 2026-08-07
stack_recorder: symbolize fully pool-backed gVisor guest text under build-id mode

josefbacik:
## What

The earlier build-id pool fix routes a gVisor guest frame through the live symbolizer only when its reconstructed address bridges to a congruent file-backed neighbour (a syscall-patch island: a few 4 KB `runsc-memory` pages interleaved in otherwise file-backed text). But when a binary runs from a memory-backed overlay — the container's writable layer, or a tmpfs mount — its *entire* text lives in the `runsc-memory` pool memfd with no file-backed neighbour at all. That covers exactly the binaries a build pod or interpreter extracts and runs at runtime (compilers, node, libpython, libjvm, and friends), and for them the reconstructed address has no neighbour to bridge, so the frame declines to the store path and renders `module:unknown`.

This PR adds that case. When a pool candidate has no file neighbour, it authenticates the image against its own ELF build-id note — read from the live process memory at the image base (`/proc/<tgid>/mem`, which reads by address and so works for the anonymous pool memfd that carries no path link) — and on a build-id match returns the reconstructed address for the same live symbolizer the flag-off path already uses. For a fully pool-backed image the memfd bytes are the real binary, not a patched island copy, so the image's own note is authoritative.

## Why this is safe

- Decline-safe: no image base, unreadable memory, a malformed ELF, or a build-id mismatch all fall through to `None` — the store path, i.e. today's behaviour with zero regression.
- Flag-off is byte-identical: `UserFrame::BuildId` frames only exist under `--collect-build-id`, and nothing outside that path changed.
- No new capability: the profiler already reads the process's memory on the live symbolization path; this reads the same memory to confirm a build-id.
- The ELF note parser reads the guest's own (semi-trusted) bytes, so it is bounded: every buffer is length-checked before indexing, addresses use checked arithmetic, and the program-header count and note size are capped. A malformed header can only make it decline.

## Testing

- `parse_build_id_from_elf` is split from the `/proc/<tgid>/mem` read so the note walk is unit-tested with synthetic ELF bytes: a well-formed image yields its build-id; bad magic and short reads decline.
- `pool_image_base` derivation, and the full-pool reconstruction declining when the process memory is unreadable (so a coincidental pool-offset match never resolves against an unconfirmed image).
- cargo fmt / build / clippy (`-D warnings`) / test green: 464 lib tests pass, the three new fixtures included.

### DIFF
--- a/src/stack_recorder.rs
+++ b/src/stack_recorder.rs
@@ -673,9 +673,101 @@ fn note_to_id(bytes: &[u8]) -> [u8; 20] {
 ///    it only if it bridges to a real file — [`ProcessMaps::bridge_for`]'s
 ///    neighbour-congruence is the disambiguator, and the bridged neighbour's
 ///    own ELF note must match the frame's build-id (the pool memfd's note is
-///    unreliable, but the neighbour file's is authoritative). A coincidental
-///    pool-offset match that does not bridge, or bridges to a different
-///    binary, falls through to the store path unchanged.
+///    unreliable, but the neighbour file's is authoritative).
+///
+/// 3. Fully pool-backed guest text: on a memory-backed overlay (the container
+///    writable layer, or tmpfs) the whole image lives in the pool memfd with
+///    no file-backed neighbour, so case 2 finds nothing to bridge. Here the
+///    image's own ELF note IS authoritative — the bytes are the real binary,
+///    not a syscall-patch copy — so read it from the process's live memory at
+///    the image base and, on a build-id match, return the reconstructed
+///    address for the caller to symbolize through the live cascade. A
+///    coincidental pool-offset match in an unrelated image fails the note
+///    match and falls through to the store path unchanged.
+///
+/// Every non-match falls through to the store path, so this only ever adds a
+/// live-process resolution ahead of the unchanged fallback.
+fn read_build_id_from_process_mem(tgid: i32, image_base: u64) -> Option<[u8; 20]> {
+    use std::os::unix::fs::FileExt;
+    let mem = std::fs::File::open(format!("/proc/{tgid}/mem")).ok()?;
+    parse_build_id_from_elf(image_base, |addr, len| {
+        let mut buf = vec![0u8; len];
+        mem.read_exact_at(&mut buf, addr).ok()?;
+        Some(buf)
+    })
+}
+
+/// The ELF-parsing core of [`read_build_id_from_process_mem`], split from the
+/// `/proc/<tgid>/mem` read so it is testable with synthetic bytes. `read`
+/// returns `len` bytes at an absolute virtual address, or `None`.
+///
+/// Parses the guest's own Elf64 image at `image_base` with the `object` crate
+/// and returns the first `NT_GNU_BUILD_ID` note descriptor, or `None` on any
+/// malformed or short input. The image is read through object's bounds-checked
+/// parsers segment by segment rather than loaded whole: a running image keeps
+/// no section-header table in memory (sections are not `SHF_ALLOC`), so the
+/// high-level `Object::build_id` — which needs it — cannot be used here; the
+/// program headers, which do live in memory, reach the note on their own. Each
+/// read is size-capped because the bytes are another process's (semi-trusted)
+/// memory.
+fn parse_build_id_from_elf(
+    image_base: u64,
+    read: impl Fn(u64, usize) -> Option<Vec<u8>>,
+) -> Option<[u8; 20]> {
+    use object::elf::{FileHeader64, NT_GNU_BUILD_ID, PT_NOTE};
+    use object::read::elf::{FileHeader, NoteIterator, ProgramHeader};
+    use object::Endianness;
+    use std::mem::size_of;
+
+    type Ehdr = FileHeader64<Endianness>;
+
+    // A malformed header must not drive a huge read of another process's
+    // memory: real images carry a few program headers and a tiny build-id note.
+    const MAX_PHNUM: u64 = 256;
+    const MAX_NOTE_BYTES: u64 = 64 * 1024;
+
+    // Header first — object validates the magic, class, version and endianness
+    // and declines on anything malformed; `parse` reads only the fixed header.
+    let ehdr_bytes = read(image_base, size_of::<Ehdr>())?;
+    let header = Ehdr::parse(&*ehdr_bytes).ok()?;
+    let endian = header.endian().ok()?;
+
+    // Read the header plus the program-header table (which sits just past the
+    // header, at e_phoff) in one span so object can index the table.
+    let phnum = header.e_phnum(endian) as u64;
+    if phnum > MAX_PHNUM {
+        return None;
+    }
+    let table_end = header
+        .e_phoff(endian)
+        .checked_add(phnum.checked_mul(header.e_phentsize(endian) as u64)?)?;
+    let front = read(image_base, usize::try_from(table_end).ok()?)?;
+
+    for ph in header.program_headers(endian, &*front).ok()? {
+        if ph.p_type(endian) != PT_NOTE {
+            continue;
+        }
+        // A PT_NOTE segment maps at image_base + p_offset; read exactly its
+        // bytes and let object walk the notes (it handles the note alignment
+        // and every bound).
+        let filesz = ph.p_filesz(endian);
+        if filesz == 0 || filesz > MAX_NOTE_BYTES {
+            continue;
+        }
+        let seg = read(
+            image_base.checked_add(ph.p_offset(endian))?,
+            usize::try_from(filesz).ok()?,
+        )?;
+        let mut notes = NoteIterator::<Ehdr>::new(endian, ph.p_align(endian), &seg).ok()?;
+        while let Some(note) = notes.next().ok()? {
+            if note.n_type(endian) == NT_GNU_BUILD_ID && note.name() == b"GNU" {
+                return Some(note_to_id(note.desc()));
+            }
+        }
+    }
+    None
+}
+
 fn reconstruct_build_id_addr(
     maps: Option<&ProcessMaps>,
     id: &[u8; 20],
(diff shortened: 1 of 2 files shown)


josefbacik:
Review performed per `.claude/agents/rust-bpf-code-reviewer.md` (in-repo definition, blob `f09e9b6`) over the full diff by automated agent tooling; a human maintainer should still give it a normal pass.

### Summary

Extends `reconstruct_build_id_addr` with a third case: fully pool-backed gVisor guest text (a binary on a memory-backed overlay or tmpfs, whose whole image lives in the `runsc-memory` pool memfd with no file-backed neighbour). #179's pool path required `bridge_for` to find a congruent file neighbour and declined without one; that is exactly the dominant failing class on GKE. The fix authenticates such an image against its own ELF build-id note read from the live process memory and returns the reconstructed address for the existing live symbolizer. Small, decline-safe, no unsafe, no schema or version change.

### Critical Issues

None found. Checked explicitly:
- No `Cargo.toml`/`Cargo.lock` version modification (bump-on-merge preserved).
- No schema change — symbolization-only, so no `SCHEMA_VERSION`/`SCHEMA_CHANGES.md`/`[bump minor]`.
- No new `unsafe`. The process-memory read uses safe `std::os::unix::fs::FileExt::read_exact_at`.
- No recorder/`utid` surface.

### Improvements (caught and applied during review)

- The ELF note parser reads the guest's own semi-trusted bytes, so it was hardened: every buffer is length-checked before indexing (`ehdr.len() < 64`, `ph.len() < 56`), every address is computed with checked arithmetic (`checked_add`/`checked_mul`), and the segment count and note size are capped (`e_phnum <= 256`, `p_filesz <= 65536`). A malformed header can only make the parse decline — never panic, wrap, or over-read.

### Suggestions

- The parser is split into `read_build_id_from_process_mem` (the `/proc/<tgid>/mem` open + read) and a pure `parse_build_id_from_elf(image_base, read)` core, so the note walk is unit-tested with synthetic bytes rather than a live process. This is why the positive path has real coverage without a fixture process.

### Positive Notes

- Decline-safe by construction: no image base, unreadable memory, malformed ELF, or build-id mismatch all fall through to `None` → the store path, i.e. current behaviour with zero regression. Flag-off is untouched (`UserFrame::BuildId` only exists in build-id mode).
- The authentication is correct-by-authority: for a fully pool-backed image the memfd bytes ARE the real binary (unlike a syscall-patch island copy), so the image's own note is authoritative — the fix reads it from live memory rather than trusting a coincidental pool-offset match, and a mismatch declines.
- `pool_image_base` derives the ELF-header address as `start - pool_pgoff`, the affine constant across an image's pool mappings — documented with the `pool_offset = vaddr - image_base + K` relationship it rests on.

### Security note

This adds a parse of guest-controlled ELF bytes in the host-side profiler. It is bounded (length-checked reads, checked arithmetic, size/count caps), performs no code execution, holds no new capability (the profiler already reads the process's memory on the live path), and is decline-safe. The bytes are the guest's own binary, which the guest already runs.

Gates: cargo fmt --check / build --tests / clippy --tests `-D warnings` / test — all green; 464 lib tests pass including the three new fixtures (synthetic-ELF note parse + malformed-input decline, image-base derivation, full-pool decline-when-unreadable). Bound to the committed text.

src/stack_recorder.rs
```diff
@@ -673,9 +673,90 @@ fn note_to_id(bytes: &[u8]) -> [u8; 20] {
...
+/// `NT_GNU_BUILD_ID` descriptor. `None` on any malformed or short input.
+fn parse_build_id_from_elf(
+    image_base: u64,
+    read: impl Fn(u64, usize) -> Option<Vec<u8>>,
+) -> Option<[u8; 20]> {
+    // The bytes are the guest's own (semi-trusted) ELF image, so every read
+    // is length-checked and every address computed with checked arithmetic —
+    // a malformed header can only make this decline, never panic or wrap.
+    let ehdr = read(image_base, 64)?;
+    if ehdr.len() < 64 || ehdr[0..4] != *b"\x7fELF" || ehdr[4] != 2 || ehdr[5] != 1 {
```
josefbacik: No magic numbers please, lets use constants so we know what the numbers mean.

src/stack_recorder.rs
```diff
@@ -673,9 +673,90 @@ fn note_to_id(bytes: &[u8]) -> [u8; 20] {
...
+        return None;
+    }
+    let e_phoff = u64::from_le_bytes(ehdr[32..40].try_into().ok()?);
+    let e_phentsize = u16::from_le_bytes(ehdr[54..56].try_into().ok()?) as u64;
+    let e_phnum = u16::from_le_bytes(ehdr[56..58].try_into().ok()?) as u64;
+    if e_phentsize < 56 || e_phnum == 0 || e_phnum > 256 {
```
josefbacik: Same here, no magic numbers.

src/stack_recorder.rs
```diff
@@ -673,9 +673,90 @@ fn note_to_id(bytes: &[u8]) -> [u8; 20] {
...
+            .checked_add(e_phoff)?
+            .checked_add(i.checked_mul(e_phentsize)?)?;
+        let ph = read(ph_addr, 56)?;
+        // PT_NOTE == 4.
+        if ph.len() < 56 || u32::from_le_bytes(ph[0..4].try_into().ok()?) != 4 {
+            continue;
+        }
+        let p_offset = u64::from_le_bytes(ph[8..16].try_into().ok()?);
+        let p_filesz = u64::from_le_bytes(ph[32..40].try_into().ok()?);
+        if p_filesz == 0 || p_filesz > 65536 {
```
josefbacik: This whole section as well.

src/stack_recorder.rs
```diff
@@ -673,9 +673,90 @@ fn note_to_id(bytes: &[u8]) -> [u8; 20] {
...
+        }
+        // PT_NOTE lives in the first PT_LOAD (file offset 0 mapped at
+        // image_base), so its in-memory address is image_base + p_offset.
+        let notes = read(image_base.checked_add(p_offset)?, p_filesz as usize)?;
+        let mut pos = 0usize;
+        while pos + 12 <= notes.len() {
+            let namesz = u32::from_le_bytes(notes[pos..pos + 4].try_into().ok()?) as usize;
+            let descsz = u32::from_le_bytes(notes[pos + 4..pos + 8].try_into().ok()?) as usize;
+            let ntype = u32::from_le_bytes(notes[pos + 8..pos + 12].try_into().ok()?);
+            let name_start = pos + 12;
```
josefbacik: Here again.

src/stack_recorder.rs
```diff
@@ -1794,6 +1888,78 @@ mod tests {
...
+        let mut b = vec![0u8; 120];
+        b[0..4].copy_from_slice(b"\x7fELF");
+        b[4] = 2; // ELFCLASS64
+        b[5] = 1; // little-endian
+        b[32..40].copy_from_slice(&64u64.to_le_bytes()); // e_phoff
+        b[54..56].copy_from_slice(&56u16.to_le_bytes()); // e_phentsize
+        b[56..58].copy_from_slice(&1u16.to_le_bytes()); // e_phnum
+        b[64..68].copy_from_slice(&4u32.to_le_bytes()); // p_type = PT_NOTE
+        b[72..80].copy_from_slice(&120u64.to_le_bytes()); // p_offset
+        let note_len = 12 + 4 + 20; // nhdr + "GNU\0" + 20-byte id
```
josefbacik: Same here for this whole function.

josefbacik (commented):
You've left a bunch of magic numbers, please address this.

src/stack_recorder.rs
josefbacik: Done — the header check now reads `elf::MAGIC`, `elf::EI_CLASS`/`elf::ELFCLASS64`, and `elf::EI_DATA`/`elf::ELFDATA2LSB` from a new `mod elf` that names every ELF offset, type tag, and sanity cap in one place.

src/stack_recorder.rs
josefbacik: Done — `elf::E_PHOFF`, `elf::E_PHENTSIZE`, `elf::E_PHNUM`, `elf::PHDR_SIZE`, and the `elf::MAX_PHNUM` cap. Field widths now come from `size_of::<uN>()` rather than literal 2/8.

src/stack_recorder.rs
josefbacik: Done — the segment loop uses `elf::P_TYPE`/`elf::PT_NOTE`, `elf::P_OFFSET`, `elf::P_FILESZ`, and the `elf::MAX_NOTE_BYTES` cap.

src/stack_recorder.rs
josefbacik: Done — the note walk uses `elf::NHDR_SIZE`, `elf::N_NAMESZ`/`elf::N_DESCSZ`/`elf::N_TYPE`, `elf::NOTE_ALIGN`, `elf::NT_GNU_BUILD_ID`, and `elf::GNU_NAME`/`elf::GNU_NAMESZ`.

src/stack_recorder.rs
josefbacik: Done — `synth_elf_with_build_id` now builds the synthetic image from the same `elf::*` constants (`use super::elf::*`), so the test and the parser can't drift apart on any offset.

josefbacik:
Good call — switched the parse to the `object` crate. It's already a direct dependency here, and it's the same crate the rest of the tree uses to read ELF (`object::File::parse` in `systing_core.rs` and `pystacks/discovery.rs`). The hand-rolled header/program-header/note walk and its constants are gone: `parse_build_id_from_elf` now uses object's `FileHeader64` + `program_headers()` + `NoteIterator`, and the test builds its fixture from the same object structs so the two can't drift on a field offset.

Two notes, since the obvious library choices don't quite fit this spot:

- **Not `read_elf_build_id`** (blazesym's helper, already imported in this file): it takes a file path. This code path exists precisely because fully pool-backed guest text has no backing file — there's nothing on disk to hand it — so we read the build-id note out of the process's own memory (`/proc/<tgid>/mem`) instead.
- **Not object's high-level `Object::build_id()`** either: it needs the section-header table, and a running image doesn't keep one mapped (sections aren't `SHF_ALLOC`). The program headers, which are mapped, are enough to reach the `PT_NOTE` segment — so the parse goes through object's low-level program-header/note API rather than loading the whole file.

Behavior is unchanged (same tests pass). The reads stay bounded and decline on anything malformed, since the bytes are another process's memory. Branch is updated.

PR #192, opened 2026-08-08
stack_recorder: bound the file-backed symbol cache with evict()

josefbacik:
## What this does

The symbolization phase already has a memory valve that rebuilds the blazesym
symbolizer when it grows too large, but it watches only anonymous RSS. The
dominant cost is the file-backed symbol cache: blazesym mmaps every ELF and
`.so` it symbolizes and never releases them, and the anon watch cannot see
that. In practice the cache runs about 3 GiB and is the largest single term of
the tracer's roughly 3.9 GiB peak. This change adds a second budget on the
file-backed axis, and when that budget is exceeded it evicts the fattest cached
binaries with blazesym's `evict()` (fattest-first, releasing their mmaps)
before falling back to the existing full rebuild. The tracer's per-capture peak
drops by roughly 3 GiB.

## Why it matters

Two out-of-memory classes need this headroom, both on an 8 GiB memory slice:
- Whole-system captures on high-core hosts, where the symbol cache alone
  approaches the slice.
- Long on-demand captures (720 seconds) on busy nodes, where the tracer stacks
  the symbol cache on top of the DuckDB build phase and is killed partway
  through, so users get only partial results.

## Re-land history

This re-lands work that was reverted, so the precedent is on the record:
- The `evict()` API this consumes was added upstream in
  [blazesym#1613](https://github.com/libbpf/blazesym/pull/1613) (merged).
- It was first consumed by
  [#170](https://github.com/josefbacik/systing/pull/170), which bumped blazesym
  and added a count-based watermark on the exited-process snapshot.
- [#170](https://github.com/josefbacik/systing/pull/170) was reverted the same
  day by [#171](https://github.com/josefbacik/systing/pull/171) as collateral in
  an unrelated rollback (the #165 machinery), not because of `evict()`.

This change is broader than a re-land of #170. #170 bounded only the
exited-process snapshot, by entry count; it never touched the live-process
cache, which is where the roughly 3 GiB actually lives. This bounds the live
cache on the file-backed RSS axis and evicts by resident size.

## The eviction key

blazesym keys its ELF cache by each mapping's `/proc/<pid>/map_files/<start>-<end>`
path (its `EntryPath::maps_file`, the cache's `actual_path`), not the
`/proc/<pid>/maps` text path. Evicting by the wrong path is a silent no-op. This
change evicts by the map_files link that `ProcessMaps::exec_file_links` yields,
which is byte-for-byte the format blazesym constructs. An integration test
(`tests/symbolizer_evict.rs`, run under the privileged vng rig) confirms at
runtime that evicting by that link releases the cached file descriptor.

## Mechanism

<details>
<summary>Why the anon watch was blind to the file cache</summary>

`current_anon_rss_bytes()` computes resident minus shared from
`/proc/self/statm`. The shared term it subtracts is exactly the file-backed
symbol cache, because mmap'd ELFs are resident-shared rather than heap. So the
existing `symbolizer_memory_budget` check never fires on the file cache no
matter how large the cache grows. The confirming signature is that the tracer's
`rss_file` is flat at about 3 GiB across hosts; a sawtooth would show if a valve
were bounding it.

The fix watches the file-backed axis directly. `current_file_rss_bytes()` reads
the statm shared term, and `symbolizer_file_budget()` (cgroup/8) is held well
under the anon budget so the symbolization footprint leaves cgroup headroom for
the later DuckDB phase, which is where long captures run out of memory. When the
file budget is exceeded the valve evicts the fattest touched binaries by on-disk
size until back under budget, then falls back to a full rebuild only for anon
growth that a targeted evict cannot reach.

The file valve deliberately has no growth-floor anti-thrash (unlike the anon
rebuild path): a tight bound is correct for OOM prevention, and re-parsing a
re-touched evicted binary is the accepted cost, mitigated by fattest-first.
</details>

## Testing

- Unit tests for the eviction planner (fattest-first order, stops once under
  budget, deterministic tie-break) and the budget/RSS helpers.
- `tests/symbolizer_evict.rs` (vng, privileged): evicting by the map_files link
  releases blazesym's cached fd — the runtime counterpart to the source-verified
  key.
- `memory_recorder` passes under vng: the change is benign in the normal
  recording path (the valve does not fire below budget).

The roughly 3 GiB per-capture reduction is measured on a real node after deploy.
A vng VM lacks the binary diversity to grow the cache to fleet scale, so this
change proves the mechanism here, not the fleet magnitude.

## Note on Cargo.lock

The blazesym 0.2.6 bump re-resolved `windows-sys` to 0.59.0 across the tree.
Those are `cfg(windows)`-only dependencies with no effect on the Linux build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -677,8 +677,8 @@ dependencies = [
 
 [[package]]
 name = "blazesym"
-version = "0.2.5"
-source = "git+https://github.com/libbpf/blazesym.git?rev=862c2cf5d424307456322d2c68fe86c591baece2#862c2cf5d424307456322d2c68fe86c591baece2"
+version = "0.2.6"
+source = "git+https://github.com/libbpf/blazesym.git?rev=8705da0a8cec2bb27f897ac732e27121f9be7936#8705da0a8cec2bb27f897ac732e27121f9be7936"
 dependencies = [
  "cpp_demangle",
  "crc32fast",
@@ -1150,7 +1150,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1225,7 +1225,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2709,7 +2709,7 @@ dependencies = [
  "once_cell",
  "socket2 0.5.10",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3046,7 +3046,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.60.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3518,7 +3518,7 @@ dependencies = [
  "fastrand",
  "once_cell",
  "rustix 0.38.41",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ bitfield = "0.19.0"
 # 0.2.x releases (where Sym::code_info became Option<Box<CodeInfo>>), so
 # `cargo install` without --locked picked an incompatible version. An exact pin
 # keeps fresh resolutions reproducible.
-blazesym = { git = "https://github.com/libbpf/blazesym.git", rev = "862c2cf5d424307456322d2c68fe86c591baece2" }
+blazesym = { git = "https://github.com/libbpf/blazesym.git", rev = "8705da0a8cec2bb27f897ac732e27121f9be7936" }
 clap = { version = "4.5.20", features = ["derive"] }
 ctrlc = "3.4"
 dashmap = "6"
--- a/src/stack_recorder.rs
+++ b/src/stack_recorder.rs
@@ -21,7 +21,7 @@ use crate::utid::{ThreadAwareRecorder, UtidGenerator};
 use blazesym::helper::{read_elf_build_id, ElfResolver};
 use blazesym::symbolize::source::{Elf, Kernel, Process, Source};
 use blazesym::symbolize::{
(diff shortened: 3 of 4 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — bound the file-backed symbol cache with evict()

_Automated pre-merge review against the repo's `rust-bpf-code-reviewer` criteria (`main` @ `aa657c80`) plus the pending systing#191 ELF-parsing section. A human review is still required._

### Summary
The change adds a second symbolization memory valve on the file-backed RSS axis and evicts blazesym's mmap'd symbol cache fattest-first when it exceeds budget, before the existing full-rebuild fallback. The core policy is extracted into a pure, unit-tested `plan_evictions`; the eviction key is source-verified against blazesym and runtime-confirmed under vng. No Critical issues.

### Critical Issues
None.
- Versioning: `Cargo.toml` `version` is not modified (only the blazesym git rev); the systing entry in `Cargo.lock` is untouched.
- Schema: no schema surface touched (no `create_schema`, `schema.rs`, or table defs), so no SCHEMA_VERSION / SCHEMA_CHANGES.md / `[bump minor]` obligation.
- ELF parsing (systing#191): no hand-rolled ELF parsing — statm text and map_files path formatting only; all ELF handling is delegated to blazesym.
- Unsafe: both `unsafe` blocks (sysconf, malloc_trim) carry `// SAFETY:` comments.

### Improvements
- **Tight bound vs re-parse thrash (intentional, documented in code):** the file valve evicts whenever file-RSS exceeds budget, with no growth-floor anti-thrash (unlike the anon rebuild path's `rebuild_floor`). This is deliberate: the goal is OOM prevention, so a tight bound is correct, and re-parsing a re-touched evicted binary is the accepted cost, mitigated by fattest-first (evict big-cold binaries, keep small-hot ones). Flagged so it reads as a conscious choice.

### Suggestions
- `plan_evictions` uses on-disk file size as an upper bound on the resident drop per evict, so a pass can under-shed if on-disk size much exceeds resident; documented, and the valve re-fires next interval.
- `current_file_rss_bytes` (statm shared) counts the tracer's own file-backed mappings too, not only the blazesym cache — but that constant term is small relative to the budget, so bounding the total effectively bounds the cache.
- `exec_file_links` dedups per backing-file (first exec mapping's range); a binary with more than one executable segment caches under more than one maps_file in blazesym, so a deduped evict is partial. The bulk (single-r-xp common binaries, which is the 3 GiB) matches; the rebuild fallback and next-interval re-fire bound any residual.

### Positive Notes
- The eviction key is traced, not guessed: blazesym's `actual_path` is `EntryPath::maps_file`, byte-identical to `exec_file_links`' first element, and runtime-confirmed under vng (evicting by the map_files link drops the cached fd). The review caught an initial wrong-key no-op (keying by the display path) before it shipped.
- Clean refactor: the eviction policy is a pure `plan_evictions` with deterministic unit tests; the closure only actuates it.
- The trigger-axis fix (watch file-backed RSS, which the anon budget subtracts) is the root cause the prior valve missed, and it is documented in the code.

PR #193, opened 2026-08-11
Fix rss_stat attribution in reclaim context; add passive thrash sampling [bump minor]

josefbacik:
Reclaim-driven RSS changes were being dropped or misattributed: `kmem:rss_stat` fires from the *reclaimer's* context (kswapd, khugepaged, direct reclaim, a `memory.reclaim`/`process_madvise` writer), but both attach paths attributed the event to `current`. In cgroup-targeted mode the reclaimer fails the task filter and the victim's RSS drop is silently lost; in whole-system mode the drop is charged to the reclaimer, with the threshold-batching state keyed under the wrong tgid. This PR attributes by the tracepoint's mm instead, labels external-reclaim updates, and adds passive per-task thrash counters — so "process released memory" and "reclaim evicted its pages" become distinguishable, and steady-state thrash (evict + refault churn with flat RSS) becomes visible.

Three legs:

1. **Attribution by mm→owner** (`tp_btf/rss_stat`): when `current->mm != mm`, resolve `mm->owner` (CO-RE field-exists guard for `CONFIG_MEMCG`) and attribute the event, the trace-task filter, and the threshold batching to the owner. The common self-update case still uses `current` and pays one pointer compare. Exit teardown stays self-attributed on both of its shapes: same-tgid owner updates (`exit_mm()` nulls `current->mm` before `exit_mmap()` drops the counters), and the last-user case where `mm_update_next_owner()` has already cleared `mm->owner` — recognized by the exiting task still carrying the mm as `active_mm` — so per-process RSS falls to zero at exit. Ownerless updates from anyone else (a reclaimer hitting an mm whose owner is gone) are dropped as unattributable.

2. **`external` label**: a flag bit set when the updater is outside the mm's thread group, surfaced as a `memory_rss.external` BOOLEAN column. Splitting member deltas by it gives a per-process evicted-bytes timeline. Exit-flush residuals stay unflagged (mixed provenance), and under threshold batching the flag reflects the update that crossed the threshold — both documented in the schema notes.

3. **Passive thrash counters**: a companion `MEMORY_THRASH_SAMPLE` event at the existing periodic mm snapshot samples `task->maj_flt` and delayacct's `thrashing_count`/`thrashing_delay`, emitted as synthetic members -3/-4/-5. No new hooks, no per-page tracking. The event reuses the header's per-type scratch fields, so the shared event header does not grow. Whether delayacct was readable is recorded per sample (a flag bit), which makes the semantics definitive: zero-valued -4/-5 rows mean "delayacct on, no thrash", absent -4/-5 rows alongside -3 mean delayacct is off on the host (boot param or `kernel.task_delayacct` sysctl), and `maj_flt` works everywhere.

Classic fallback (`tracepoint/kmem/rss_stat`): the raw tracepoint carries a hashed `mm_id` plus a `curr` flag, no mm pointer, so remote updates cannot be resolved there — `curr==0` events are dropped (documented bounded loss; they were never correctly attributed on this path anyway, and the path is only selected on kernels whose BTF lacks `btf_trace_rss_stat`).

Out of scope, per the proposal: per-page tracking (rmap-side hooks are too hot), and munmap `size` stays caller-requested by design.

Schema: version 13→14 (`external` column + three synthetic members); SCHEMA_CHANGES.md carries the reader-facing semantics.

Testing: full suite green (522 passed, 0 failed), `cargo fmt --check` and `cargo clippy --all-targets` clean. On-kernel BPF load-verify under virtme-ng follows as a comment on this PR before any merge ask.

### DIFF
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -296,3 +296,51 @@ peer, not a local interface) and IPv4-mapped-IPv6 sockets correctly.
 
 Pre-v13 databases and traces have neither column; readers join on `netns_inum`
 where present and fall back to the `src_ip`→interface heuristic otherwise.
+
+## Schema Version 14 (unreleased) — 2026-08-11
+
+`memory_rss` gains an `external BOOLEAN` column and three synthetic members
+(-3/-4/-5), fixing rss_stat attribution in reclaim context and adding passive
+per-task thrash signals.
+
+### Attribution fix (no column)
+rss_stat events are now attributed to the mm's owner rather than to the task
+that happened to be running. Reclaim (`try_to_unmap_one`) decrements the
+victim's counters from the reclaimer's context, so reclaim-driven RSS drops of
+a traced process were previously dropped in cgroup-targeted mode (kswapd fails
+the filter) and misattributed to the reclaimer in whole-system mode (with the
+threshold batching state keyed by the wrong tgid). The common self-update case
+is unchanged, and a process's own exit teardown stays self-attributed — even
+when `mm_update_next_owner()` has already cleared `mm->owner` — so per-process
+RSS still falls to zero at exit. On the classic (non-BTF) fallback attach path
+the raw tracepoint carries no mm pointer, so remote updates are dropped there
+instead of misattributed.
+
+### New columns
+- `memory_rss.external` (BOOLEAN): true when the counter update was performed
+  from outside the process's own thread group — an external reclaimer (kswapd,
+  khugepaged, another process's direct reclaim, a `memory.reclaim` /
+  `process_madvise` writer) evicted or migrated the process's pages. False for
+  the process's own faults/maps/unmaps/exit teardown, for synthetic members,
+  and for exit-flush residuals (mixed provenance). Splitting cumulative member
+  deltas by this flag yields a per-process evicted-bytes timeline (under
+  threshold batching the flag reflects the update that crossed the threshold).
+
+### New synthetic members (periodic mm snapshots, `utid` = sampled thread)
+- `-3` maj_flt: cumulative major fault count for the sampled thread — a page
+  the process needed came back from disk/swap. Count, not bytes.
+- `-4` thrashing_count / `-5` thrashing_delay_ns: delayacct's thrash metrics
+  (stalls on `PG_workingset` pages — pages that refaulted soon after reclaim).
+  Emitted (zero values included) whenever delayacct was readable for the
+  sampled task, so zero-valued rows mean "delayacct on, no thrash", and their
+  complete absence while -3 rows are present means delayacct is not enabled
+  on the host (`CONFIG_TASK_DELAY_ACCT` plus the `delayacct` boot parameter
+  or the `kernel.task_delayacct` sysctl — runtime-enableable, but only tasks
+  forked after enablement carry the counters).
+
+"RSS flat while maj_flt and thrashing_delay climb" is a direct per-process
+thrash signature; `external` tells you whose pages reclaim is taking. Pre-v14
+databases and traces have neither the column nor the members: merging a
+pre-v14 DuckDB into a v14 database (`systing-util convert`) NULL-fills
+`external` via the column-intersection import, and readers of raw pre-v14
+parquet must treat the column as absent.
--- a/skills/systing-analyze/SKILL.md
+++ b/skills/systing-analyze/SKILL.md
@@ -113,7 +113,7 @@ ORDER BY retransmits DESC;
 ### Memory
 From the `memory` / `memory-alloc` recorders. `stack_id` columns join to `stack.id` like sample stacks.
 
-- `memory_rss(ts, utid, member, size)` — RSS tracking. `member`: `0`=file, `1`=anon, `2`=swap, `3`=shmem, `-1`=hiwater_rss, `-2`=total_vm (the negatives are synthetic periodic `mm_struct` samples).
+- `memory_rss(ts, utid, member, size, external)` — RSS tracking. `member`: `0`=file, `1`=anon, `2`=swap, `3`=shmem, `-1`=hiwater_rss, `-2`=total_vm, `-3`=maj_flt, `-4`=thrashing_count, `-5`=thrashing_delay_ns (negatives are synthetic periodic samples). Units vary by member: `size` is bytes for `0..3`/`-1`/`-2`, a cumulative fault count for `-3`, a stall count for `-4`, and nanoseconds for `-5` — never aggregate across members. `external` is true when the update came from an external reclaimer (kswapd, direct reclaim by another process) rather than the process itself; `-4`/`-5` rows appear only on delayacct-enabled hosts (zero-valued rows there mean "no thrash").
 - `memory_map(ts, utid, event_type, addr, size, prot, flags, stack_id)` — mmap/munmap/brk.
 - `memory_fault(ts, utid, addr, error_code, stack_id)` — user page faults (sampled 1-in-N, default 97).
 - `memory_alloc(ts, utid, op, addr, size, old_addr, stack_id)` — malloc/calloc/realloc/free uprobes.
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -412,6 +412,7 @@ enum memory_event_type {
 	MEMORY_MM_SAMPLE,     // periodic mm_struct snapshot (from perf_event_clock)
 	MEMORY_ALLOC,         // uretprobe malloc/calloc/realloc/... - heap object allocated
 	MEMORY_FREE,          // uprobe free - heap object released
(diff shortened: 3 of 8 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `agents/rust-bpf-code-reviewer.md` definition over the full diff, run by the authoring session before commit, per this repo's review protocol.

### Summary
Three-leg change: mm→owner attribution for rss_stat, an external-reclaim flag, and passive thrash sampling via a companion event. Hot-path cost in the common self-update case is one pointer compare; schema and versioning gates are satisfied. One structural issue found during review was fixed before commit (below).

### Critical Issues
None outstanding. Checked explicitly: SCHEMA_VERSION bumped 13→14 with a SCHEMA_CHANGES.md entry; `Cargo.toml`/`Cargo.lock` untouched; `[bump minor]` present in the PR title; thread identity uses `utid` throughout (no raw tid/pid columns); no hand-rolled ELF parsing introduced; no source-line references in comments.

### Improvements (found and fixed pre-commit)
- The initial implementation added three `u64` fields to `memory_event_header` for the thrash counters — +24 bytes on every memory event, including the high-frequency stack-less types the split-header design exists to keep at ~80 bytes, for fields only the low-rate periodic sample fills. Reworked to a dedicated `MEMORY_THRASH_SAMPLE` event reusing the per-type scratch fields (`addr`/`size`/`old_addr`), which is the file's established idiom (those fields already change meaning by event type). Header size unchanged; the recorder names the reused fields with local bindings for readability.

Four automated pre-push findings were also adjudicated and folded in before this PR opened:
1. *Exit-teardown regression (real bug)*: the first attribution shape dropped NULL-owner events, but `mm_update_next_owner()` clears `mm->owner` for the last user before `exit_mmap()` decrements the counters — so a process's own exit RSS fall would have been lost. Fixed by recognizing the exiting task via `active_mm == mm` (plus a `PF_KTHREAD` exclusion) and keeping that case self-attributed; ownerless updates from other tasks stay dropped.
2. *False converse in the delayacct docs*: "absence of -4/-5 means delayacct is off" ignored "delayacct on, nothing thrashed". Fixed structurally, not editorially: the BPF side now records whether `task->delays` was readable (a flag bit), the recorder emits the pair — zeros included — exactly when it was, and absence is now a definitive delayacct-off signal.
3. *Hollow compat sentence*: the schema notes claimed old parquet gets NULL-filled by an import path that never actually sees old files. Rewritten to the verified truth: DuckDB-to-DuckDB merges NULL-fill `external` via the column-intersection import; raw pre-v14 parquet simply lacks the column.
4. *Stale in-repo doc*: `skills/systing-analyze/SKILL.md` still described `memory_rss` as four columns with only -1/-2 members; updated with the new column, the new members, and explicit per-member units so analysts don't aggregate counts with bytes.

### Suggestions
- The sequential `add_memory_rss` + `report_write_error` blocks in the MM_SAMPLE/THRASH_SAMPLE arms could fold into a small emit helper. Kept as-is to match the file's existing idiom; the borrow structure makes a closure form awkward for marginal gain.

### Positive Notes
- The tgid-based external discriminator classifies exit teardown correctly: `exit_mm()` nulls `current->mm` before the teardown decrements fire, so a bare mm compare would mark every process's own exit as external reclaim. The tgid comparison handles it, and the comment explains the trap.
- The classic-fallback change states what the raw tracepoint actually carries (hashed `mm_id` + `curr`, no mm pointer) and takes a documented bounded loss over silent misattribution.
- `bpf_core_field_exists` guards on both `mm->owner` (`CONFIG_MEMCG`) and `task->delays` (`CONFIG_TASK_DELAY_ACCT`) keep load-time portability; NULL-owner and NULL-delays paths degrade to drop/zero rather than fault.
- The threshold-batching flag semantics (flag reflects the update that crossed the threshold) are stated in the record docs rather than left for readers to discover.

josefbacik:
On-kernel load-verify (virtme-ng, v6.12 kernel, QEMU TCG): the full memory_recorder integration target passed against this exact revision (9fb47ac) — 7/7 tests including `test_memory_rss_threshold_tp_btf` and `test_memory_rss_threshold_classic_fallback`, so both attach paths load, attach, and record end-to-end with the new attribution logic and the companion thrash event. This closes the load-verify follow-up promised in the PR body.

josefbacik:
**Independent second review (BPF side)** — full-diff read at `9fb47ac2`, focused on the attribution three-way, the classic-fallback drop, and the companion thrash event, with the kernel-side claims checked against kernel source rather than taken from the commit message.

### Verdict
No structural findings. The attribution logic is correct against kernel source at both ends of the practically relevant stable range (6.6.y and 6.18.y), the documented bounded losses are real but correctly scoped, and the schema/docs half is consistent with the BPF half everywhere I cross-checked. Everything below is confirmation detail or non-blocking notes.

### Kernel-side claims, independently verified
- `mm->owner` exists (CONFIG_MEMCG-gated) at both 6.6.y and 6.18.25 — the fix is live on modern kernels, not silently compiled away behind the `bpf_core_field_exists` guard. Worth knowing because there have been upstream proposals to remove this field; if a future kernel drops it, this code degrades to drop-remote-updates (the documented MEMCG=n behavior) rather than misattributing, which is the right failure direction — but the reclaim visibility would quietly vanish, so a release-note check on major kernel bumps is warranted.
- The exit-teardown ordering claim is exactly right: `exit_mm()` calls `mm_update_next_owner()` (kernel/exit.c), which does `WRITE_ONCE(mm->owner, NULL)` on both no-candidate paths — the `mm_users <= 1` fast path and the search-failed path — before the final mmput reaches `exit_mmap()`'s counter decrements. So the NULL-owner + `active_mm == mm` + `!PF_KTHREAD` branch is not defensive coding; it is the only thing keeping last-user exit teardown attributed. The pre-push finding that added it caught a real loss.
- Classic-fallback semantics: `curr` in `trace_event_raw_rss_stat` is literally `current->mm == mm` at the emit site, so the `!ctx->curr` drop is scoped to exactly the events this path cannot attribute. Correct trade.
- The "absent or NULL leaves the init-zeroed values" comment on the thrash block is load-bearing (ringbuf reservations are not zeroed by the kernel), and it holds: `reserve_memory_event_header()` runs `memory_header_init()`, which zeroes `member`, `flags`, `size`, and `old_addr`. Without that, a delayacct-off host could emit garbage flags that fake the DELAYACCT bit — checked because it would have been the one silent-wrong-data path in the design; it is closed.
- `record_task_info()` is fully parameterized (reads its task argument, not current), so owner attribution records the victim's identity, not the reclaimer's.

### Shapes probed beyond the commit message (all fine, none blocking)
- **kthread_use_mm users** (legacy vhost and similar): during the adopted window `current->mm == mm`, so those updates take the common path and are attributed to the kthread — dropped in cgroup mode, kthread-attributed in whole-system mode. That is byte-identical to pre-change behavior, and the population is small since vhost moved to user-thread workers. If it ever matters, the discriminator is cheap (a `PF_KTHREAD` check on the common path routing to the owner walk); fine as a future follow-up, wrong to grow this PR for.
- **CLONE_VM non-thread sharers** (including the vfork window): a sharer's own updates attribute to the updating task's tgid via the common path (the external flag correctly never evaluates there), while reclaim against the shared mm attributes to `mm->owner`. One mm's timeline split across tgids for sharing groups is pre-existing behavior, not something this change introduces or worsens, and the flag stays truthful in both directions.
- **Deferred final mmput** (mm outliving the thread group via an elevated reference — procfs readers, `get_task_mm` holders): the teardown decrements then arrive with `owner == NULL` and an `active_mm` that doesn't match, so they are dropped as unattributable and the victim's timeline never shows the final fall to zero. Pre-change, those events went to whichever task dropped the last reference — the victim's timeline was equally zero-less. Same end state, minus the garbage attribution; acceptable documented loss.
- **Unlocked `mm->owner` read**: BPF tracepoint context is RCU-protected, so a concurrently-exiting owner is memory-safe to read, and a stale-by-a-window attribution is tracing-grade tolerance, not a correctness hole. Deliberate and fine.

### One cheap suggestion (non-blocking)
The per-type field-reuse contract (`addr`=maj_flt, `size`=thrashing_count, `old_addr`=thrashing_delay_ns, pair gated on the flag) is now load-bearing in three places — the BPF comment, the recorder decode, and the schema doc. A small decode-path unit test on the recorder side (one synthetic THRASH header with the flag set → three rows; without it → maj_flt row only) would pin that contract against future field-reuse edits for a few lines of test.

### Cross-checks on the userspace half
SCHEMA_VERSION 13→14 in `src/duckdb.rs` with the DuckDB DDL, parquet writer, Arrow schema, and both docs all carrying the `external` column in the same position; the SCHEMA_CHANGES.md entry's units table and the delayacct fork-after-enablement caveat match the kernel behavior; the threshold-batching provenance caveat ("the flag reflects the update that crossed the threshold") is stated where readers will find it. The `[bump minor]` title tag is present and `Cargo.toml` is untouched, per this repo's release automation.

PR #194, opened 2026-08-13
[bump minor] memory_map: record the signed RSS delta of each mmap/munmap/brk

josefbacik:
## What

`memory_map` rows gain `rss_delta_bytes`: the signed change in the process's resident set across each recorded mmap/munmap/brk call, measured in the BPF handlers from the mm's resident counters at syscall enter and success-gated exit. Schema v15.

## Why

The `size` column names the virtual-address range the arguments described, and for churn accounting that number misrepresents reality on both sides: mmap commits nothing (page faults do), so mapping a large sparse reservation records as if that memory appeared; munmap frees only what was resident in the range, so unmapping one records far more than the machine got back. The resident-counter delta is the number that tracks pages actually committed or freed.

## How

- The map legs' existing enter-stash (`memory_syscall_args`) gains `enter_rss`: file+anon+shmem pages via the same `read_mm_rss_pages` read the rss_stat handler uses, so the two share one error model (worst case ±(percpu batch × nr_cpus) pages of counter-caching slop).
- The delta rides the event header's per-type `old_addr` field (previously always 0 on map legs), so the event size is unchanged and no new maps or map operations are added. `S64_MIN` is the "no read" sentinel; the recorder decodes it to NULL.
- All three enter-side reads sit behind the existing `trace_task` gates, so untraced tasks pay nothing. History check on these paths: the recorded hazard was per-event cost, cured by the trace_task gating in a11016f and batching in 5638255 — this change stays behind those gates. The 2026-07 reverts (#171) concerned per-tick whole-population scanning, a different mechanism; nothing here runs per-tick.
- `rss_delta_bytes BIGINT` nullable in parquet and DuckDB. NULL means the read was unavailable (or a pre-v15 row) — never zero. SCHEMA_CHANGES.md documents the read rules: NULL semantics, the three bounded approximations (concurrent faults in the syscall window, percpu counter slop, sampling composition — a sampled row carries its own call's full delta; the rate scales aggregate mass).

## Validation

- Full suite green on the final tree; rustfmt and clippy clean.
- VM e2e on a real kernel (v6.12): the map-touch-free workload now also asserts the new column — a non-NULL control first (a failed BPF-side read would NULL every row and pass vacuously), then at least 25 of the 50 fully-touched 2 MiB frees showing drops of 1 MiB or more. The negative bound also catches an enter/exit sign inversion. Result posted in a comment below.

### DIFF
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -344,3 +344,38 @@ databases and traces have neither the column nor the members: merging a
 pre-v14 DuckDB into a v14 database (`systing-util convert`) NULL-fills
 `external` via the column-intersection import, and readers of raw pre-v14
 parquet must treat the column as absent.
+
+## Schema Version 15 (unreleased) — 2026-08-13
+
+`memory_map` gains `rss_delta_bytes BIGINT` (nullable): the SIGNED change in
+the process's resident set across each mmap/munmap/brk call — pages actually
+committed or freed, where `size` only names the virtual-address range the
+arguments described.
+
+Why: the VA number misrepresents reality on both sides. mmap commits nothing
+(page faults do), so huge mmaps of sparse reservations recorded as if memory
+appeared; munmap frees only what was resident in the range, so sparse unmaps
+recorded far more than the machine gave back. `rss_delta_bytes` is measured
+from the mm's resident counters at syscall enter and success-gated exit:
+munmap rows are typically negative (the real release), mmap rows typically
+~0, brk either sign.
+
+Read rules, stated once here and in the record docs:
+- NULL means the resident read was unavailable on either side of the call —
+  and every pre-v15 row. Never treat NULL as zero.
+- Three approximations, all bounded and deliberate: concurrent faults or
+  reclaim on the process's other threads inside the syscall window land in
+  the delta (microsecond window); the resident reads carry the rss_stat
+  sampling's worst-case ±(percpu batch x nr_cpus pages) error (the kernel's
+  counter caching — the same slop rss_stat rows already document); and under
+  `--memory-map-sample-rate` 1:N sampling only sampled events carry a delta.
+- Sampling composition: a sampled-in row carries the FULL delta of its own
+  call — the rate scales aggregate event mass (multiply sums by N), never
+  divides a row's delta. At rates above 1 the scaled estimate is unbiased
+  but heavy-tail-noisy: one huge munmap missed by sampling is invisible,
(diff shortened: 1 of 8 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — the repo's reviewer agent definition (`.claude/agents/rust-bpf-code-reviewer.md`) executed over the full diff before push; mechanics disclosed: run in-session by the authoring assistant, all five passes (structure / Rust / BPF / schema & versioning / integration).

**Summary**: a small change following the existing enter-stash / success-gated-exit pattern; per-type header-field reuse keeps the event size unchanged; schema discipline complete.

**Critical issues**: none found. Explicitly checked: SCHEMA_VERSION 14→15 with a SCHEMA_CHANGES.md entry; no Cargo.toml/Cargo.lock version edits (`[bump minor]` carried in the title instead); thread attribution still via utid; no hand-rolled ELF parsing; no line-number references in comments.

**Improvements (found during review, fixed before push)**: the record field was `rss_delta` while the column is `rss_delta_bytes` — renamed to match; the repo's precedent is exact field/column parity (`timestamp_ns`, `hbm_size_bytes`).

**Claims verified against source**: `current_task_rss_bytes` genuinely mirrors the rss_stat handler's read (same `read_mm_rss_pages` helper, same negative clamp, same page-size scaling), so the shared-error-model comment is accurate; the `MEMORY_MM_*` indices mirror the kernel's `MM_*` values, so the member switch cannot silently return zero; the `S64_MIN` sentinel cannot collide with a real delta (that would need an 8 EiB resident-set change); all three enter-side reads are behind the existing `trace_task` gates.

**Positive notes**: `enter_rss` is copied out of the scratch entry before `bpf_map_delete_elem` in brk_exit; an unreadable mm yields the absent sentinel rather than a fake zero; the e2e assertion runs its negative control (the non-NULL count) before the semantic check.

josefbacik:
VM e2e result (real kernel v6.12, TCG, memory_recorder target): **PASS — 7 passed / 0 failed** in 798s, including the new `rss_delta_bytes` assertion: non-NULL rows present (the negative control against a silently failed BPF-side read), and at least 25 of the 50 fully-touched 2 MiB frees recorded drops of 1 MiB or more (a threshold that would also catch an enter/exit sign inversion). One disclosure: the VM run was built just before a pure Rust field rename (`rss_delta` → `rss_delta_bytes` on the record struct); the BPF object and the tested column name are byte-identical across that rename, and the full suite on the final tree is green — 522 passed / 0 failed, rustfmt and clippy clean.

PR #196, opened 2026-08-18
stack_recorder: count CPU samples that land in KVM guest execution as [guest]

josefbacik:
CPU time spent inside virtual machines currently vanishes from systing profiles. When a sampling interrupt lands while a vCPU thread (Firecracker, QEMU, cloud-hypervisor, gVisor's KVM platform) is executing guest code, the kernel deliberately returns empty callchains — `perf_callchain_kernel()` and `perf_callchain_user()` both bail out under `perf_guest_state()` ("We don't support guest os callchain now", on x86 and arm64 alike). The BPF side submits the stack event anyway, frameless, and `StackRecorder::handle_event` only acted on events carrying at least one frame, so every such sample was silently discarded. A host whose CPUs are mostly busy running guests profiled as mostly idle: no error, no counter, just missing samples.

This change counts those samples. BPF tags each stack event with whether the task was in guest context (`task->flags & PF_VCPU` — the same test the kernel's cputime accounting uses to bill a tick as guest time in `/proc/stat`), using four bytes of existing padding in `stack_event`. Userspace turns a frameless, guest-tagged event into a one-frame stack rendered `unknown ([guest])`, attributed to the vCPU thread like any other sample. Guest internals stay opaque — there is nothing on the host to symbolize — but the machine's utilization, and which VMM process and thread is burning it, become truthful. Frameless events that are *not* guest-tagged (both unwinds failed for an ordinary task) are still dropped, but are now counted and reported at trace end instead of disappearing.

**Observed on a real deployment** (Kubernetes nodes that are themselves cloud VMs, running Firecracker microVMs through nested virtualization, sampled with the software cpu-clock event): comparing systing's per-capture sample totals against node_exporter for the same nodes and hour, systing accounted for roughly a fifth of busy CPU; on Firecracker-heavy nodes its total tracked node_exporter's *system* time while *user* time — where guest execution is billed — produced nothing. Of ~2×10⁸ samples on `fc_vcpu` threads, 99.8% had `vcpu_run` on the stack and exactly zero had `vmx_vcpu_run`: no sample was ever recorded between VM-entry and VM-exit. What was recorded on those threads was host-side KVM work only (VMCS loads, TDP MMU faults, TLB-flush IPIs, halt polling).

<details><summary>What changes on the wire and on disk</summary>

- `stack_event` gains `u32 sample_flags` in what was alignment padding after the leading enum: no other field moves, the record does not grow, and the offsetof-based raw-IP reservation is unchanged.
- No new BPF work per sample beyond one flag test; `task->flags` is now loaded once and shared with the existing kthread check. Ring traffic is identical — these events were already being submitted and consumed; userspace simply stops discarding them — so sample volume on VM hosts rises to reflect actual utilization.
- Stack spill format: user-frame tag 2 = guest (no payload), alongside 0 = raw IP and 1 = build-id.
- Frame string `unknown ([guest])`: label in the module slot like `[exited]` / `[gvisor:*]`, no `<0xaddr>` suffix (there is no address); in-tree consumers (`parse_module_name`, profile export) already treat the suffix as optional. Documented in `docs/PROFILE_EXPORT_FORMAT.md`. Emitted regardless of `--no-frame-labels`, since it is not an address-derived label.
- No schema change (frame strings are values), so no `SCHEMA_VERSION` bump and no `[bump minor]`; `Cargo.toml` untouched.
- Precedent check: nothing in this tree's history handles guest/vCPU samples; the nearest cost-class precedent (b25c202, work moved off the BPF probe path) does not apply, since this adds no per-event BPF work beyond one flag test on an already-loaded value.
</details>

<details><summary>Why PF_VCPU is the right mark</summary>

KVM sets `PF_VCPU` in `guest_timing_enter_irqoff()` and clears it in `guest_timing_exit_irqoff()`, which it deliberately runs only *after* servicing the interrupt that forced the VM-exit ("so that any ticks that occurred while running the guest are properly accounted to the guest"). A sampling interrupt that hits guest execution is handled inside exactly that window, with `perf_guest_state()` set — hence frameless. A guest-tagged sample that does carry frames (for example an NMI landing in the host's entry/exit glue on bare metal) keeps its real frames; the synthetic frame is used only when there is nothing else. Exit handling proper (EPT faults, emulation, halt polling) runs after `guest_timing_exit_irqoff()` and profiles normally, as the `fc_vcpu` spectrum above shows.
</details>

<details><summary>Tests</summary>

- `test_guest_sample_is_counted_and_labeled`: BPF-shaped frameless events carrying the guest flag, for a live and an exited VMM tgid → counted, interned (one dedup key per tgid), streamed, and symbolized to `unknown ([guest])` through `finish_inner`.
- `test_frameless_unmarked_sample_is_dropped_and_counted`: unmarked frameless event → dropped and counted; guest-marked event *with* a kernel frame → real stack kept, no synthetic frame.
- Spill round-trip and user-frame filter tests extended with the guest tag / variant.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, full `cargo test --lib`.
</details>

A follow-up this enables but does not attempt: visibility *inside* the guest needs cooperation from within the VM (an in-guest agent, or PMU/LBR passthrough) and is a separate design.

### DIFF
--- a/docs/PROFILE_EXPORT_FORMAT.md
+++ b/docs/PROFILE_EXPORT_FORMAT.md
@@ -125,6 +125,12 @@ Frame strings are exactly systing's symbolized frame names:
   may be a label rather than a binary basename: `[kernel]`,
   `[gvisor:runtime]`, `[gvisor:guest]`, `[jit:<runtime>]`, `[exited]`,
   `[vdso]` and similar. Unresolvable frames use the function name `unknown`.
+- Guest execution: `unknown ([guest])` — a CPU sample that interrupted a
+  hypervisor vCPU thread while it was running its virtual machine. There is
+  no host address (the kernel does not walk callchains for guest-mode
+  samples), so there is never an `<0xaddr>` suffix; the frame exists so that
+  CPU time spent inside VMs is counted and attributed to the VMM process
+  rather than dropped. It is emitted regardless of `--no-frame-labels`.
 - Python (pystacks): `name (python) [file.py:line]` — no address suffix.
 - With `--no-frame-labels`: bare hex addresses (`0x7f95bfdb6e12`).
 
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -42,6 +42,13 @@
 
 #define TASK_COMM_LEN 16
 
+/* PF_VCPU: the task is a hypervisor vCPU thread currently accounted as
+ * running its guest (set from guest_timing_enter_irqoff() until
+ * guest_timing_exit_irqoff(), which KVM deliberately runs only after the
+ * host has serviced the interrupt that forced the VM-exit). A sampling
+ * interrupt that lands while a CPU is executing guest code is handled in
+ * exactly that window. */
+#define PF_VCPU 0x00000001
 #define PF_WQ_WORKER 0x00000020
 #define PF_KTHREAD 0x00200000
 
@@ -239,8 +246,21 @@ struct marker_event {
 #define USER_STACK_FORMAT_IP 0
 #define USER_STACK_FORMAT_BUILD_ID 1
 
+/* stack_event.sample_flags bits. */
+/* The sample interrupted a vCPU thread while it was running its guest
+ * (task->flags & PF_VCPU — the same test the kernel's own cputime code uses
+ * to bill a tick as guest time). The kernel returns EMPTY callchains for
+ * such samples by design (perf_callchain_kernel()/perf_callchain_user()
+ * bail out under perf_guest_state(), on x86 and arm64 alike), so the event
+ * arrives with no frames; the flag lets userspace account the sample as
+ * guest execution instead of discarding it. */
+#define SAMPLE_FLAG_IN_GUEST (1U << 0)
+
 struct stack_event {
 	enum stack_event_type stack_event_type;
+	/* Occupies what was alignment padding before `ts`: no other field
+	 * moves and the record does not grow. */
+	u32 sample_flags;
 	u64 ts;
 	u32 cpu;
 	u32 user_stack_format;
@@ -1787,7 +1807,17 @@ static void emit_stack_event_with_ts(void *ctx, struct task_struct *task,
 
 	event->stack_event_type = type;
 
-	if (!(task->flags & PF_KTHREAD)) {
+	unsigned int task_flags = task->flags;
+	/*
+	 * A vCPU thread sampled while its guest was executing: the callchain
+	 * walks below come back empty for it (see SAMPLE_FLAG_IN_GUEST), and
+	 * without this mark userspace could not tell "CPU busy running guest
+	 * code" from an unwind failure — it would drop the sample and the
+	 * guest's CPU time would vanish from the profile entirely.
+	 */
+	event->sample_flags = (task_flags & PF_VCPU) ? SAMPLE_FLAG_IN_GUEST : 0;
+
+	if (!(task_flags & PF_KTHREAD)) {
 		/*
 		 * Both modes walk into the same user_stack region; the entry
 		 * format (and how much of the region was reserved) differs.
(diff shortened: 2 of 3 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `agents/rust-bpf-code-reviewer.md` definition over the full diff of this branch (3 files: `src/bpf/systing_system.bpf.c`, `src/stack_recorder.rs`, `docs/PROFILE_EXPORT_FORMAT.md`), all five passes. Findings raised during the review were fixed before this commit; what follows is the verdict on the code as pushed.

### Summary
A small, well-contained fix for a real accounting hole: guest-mode CPU samples arrive frameless by kernel design and were being discarded in `handle_event` with no trace. The change marks them at the source with one flag test, carries the mark in existing struct padding, and accounts them as a synthetic `[guest]` frame through the normal intern/stream/symbolize path. No new BPF work of consequence, no record growth, no schema change.

### Critical Issues
None.
- Schema & versioning: no table/schema definitions touched (frame strings are values, not schema) → no `SCHEMA_VERSION` bump and no `[bump minor]` required; `Cargo.toml`/`Cargo.lock` untouched. Verified.
- Recorder consistency: samples still flow through `utid_generator.get_or_create_utid(tid)`; no raw tid/pid columns introduced.
- No `unsafe` added; no ELF parsing involved.
- BPF map/context pairing (NMI safety) checked: this diff adds no map operation — it writes one field of an already-reserved ring record — and the maps the `perf_event` program touches today are ring buffers (`stack_ringbufs` / `perf_counter_ringbufs`, ARRAY_OF_MAPS of RINGBUF) and per-CPU arrays (`missed_events`, `last_perf_counter_value`, plus `cpu_running_pid` which it only reads); no `BPF_MAP_TYPE_LRU_HASH` map exists under `src/bpf/`.

### Improvements (raised in review, resolved before push)
1. BPF redundant read: `task->flags` was loaded twice (guest mark + existing kthread check) → now loaded once into `task_flags` and shared.
2. Test magic number: `stack_event_type.0 = 1` → replaced with the generated `stack_event_type::STACK_RUNNING`, and the helper builds the event in one initializer (clippy `field_reassign_with_default`).
3. Over-scoped visibility: `GUEST_FRAME_NAME` was `pub(crate)` with no out-of-module user → plain `const`.
4. Arch wording: comments said "on x86"; `perf_callchain_kernel/user` bail out under `perf_guest_state()` on arm64 identically (checked in-tree) → wording made arch-generic.

- The spill user-frame tags (0/1/2) are bare literals in `push`/`read_spill_record`, following the pre-existing pattern for 0 and 1; naming all three would be a reasonable follow-up but is out of scope for this fix.
- The frameless-drop count is surfaced as an end-of-trace stderr note. If that residual class ever turns out to be non-trivial in the field, promoting it to a counter track alongside the BPF missed-event classes would make it queryable; today it is expected to be ~zero and the note keeps it from being silently so.

### Positive Notes
- The `PF_VCPU` choice is exactly the predicate the kernel's cputime code uses to bill guest time, so `[guest]` sample mass corresponds to what `/proc/stat` calls guest — a checkable external reference rather than a heuristic.
- `sample_flags` occupies former alignment padding (verified in the generated skeleton: `u32` after the 4-byte enum, `ts` still at offset 8, no pad field), so the offsetof-based raw-IP reservation contract is untouched.
- Guest-marked events that *do* carry frames keep their real stack — the synthetic frame is strictly a last resort — and that behavior is pinned by a test.
- Tests exercise the actual entry point (`handle_event` with a BPF-shaped event) through to `finish_inner` symbolization for both live and exited tgids, plus spill round-trip of the new tag.

Gates on the pushed tree: `cargo fmt --check` clean; `cargo build` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 472 passed / 0 failed (includes the 2 new tests and 2 extended ones).

PR #197, opened 2026-08-18
network: upgrade a socket record whose first sighting had no peer

josefbacik:
On kernels before 6.15, every inbound TCP connection accepted during a capture is recorded with no peer: its `network_socket` row carries the listener's `0.0.0.0:0` / `[::]:0` and is never upgraded, so readers see a listener-shaped row instead of a connection. This PR makes the recorder emit its one-time upgraded record when a later event on the same socket carries the real peer, using the mechanism it already has for the port-less connect-start case. On 6.15+ kernels nothing changes. Userspace only, no schema change.

**Why it happens.** `inet_sock_set_state` fires for the LISTEN→SYN_RECV child of a passive open from `inet_csk_clone_lock()`. Before Linux 6.15 that call sat before the kernel copied `inet_dport` / `sk_daddr` into the child, so the tracepoint reports `dport=0 daddr=0.0.0.0` (upstream commit a3a128f611a9, "inet: consolidate inet_csk_clone_lock()", shows the exact before/after trace in its message). The recorder mints the socket's identity from that event and emits the record with the degraded tuple; because `src_port` is the bound listener port, `should_emit_socket_record` never upgrades it. Since eb368f3 the SYN_RECV→ESTABLISHED transition (full tuple, microseconds later) resolves to the same socket_id through the sk map, so the peer is never written anywhere.

**Observed in production traces.** On 6.12 hosts, roughly one fifth of per-capture socket rows were listener-shaped, and 99.5–99.8% of those carried a transition to ESTABLISHED — a listening socket never does that. Hosts on 6.18 kernels emit essentially none and show the same connections with real peers.

**The change.**
- `socket_tuple_is_complete(af, src_port, dest_addr, dest_port)`: bound `src_port` AND peer not (unspecified address, port 0), decoded family-aware through `parse_ip_addr` so it agrees with the record's `dest_ip` string (an IPv4 event fills only the first 4 bytes of the 16-byte buffer).
- `should_emit_socket_record` takes that bool: emit on first sight; emit ONCE more the first time a sighting is complete when the record was not; never after. A degraded sighting never downgrades a complete mark; a socket that only ever shows a degraded tuple (real listener, unconnected UDP socket) emits exactly one record.
- Doc comments on `NetworkRecorder::seen_sockets` and `NetworkSocketRecord` state the two-row reader contract for both degraded shapes.

**Reader contract.** A `socket_id` can have two rows — degraded, then complete — for either degraded shape; readers keeping one row per socket keep the complete one. That was already the contract for the port-less upgrade. Two in-tree consumers, unchanged here: `analyze/network_connections.rs` groups by (socket_id, tuple), so a passive-open child on an old kernel shows two connection rows the same way the port-less double already does (a dedupe-to-complete-row follow-up if wanted); the Perfetto converter creates a track per record and attaches events to the last one, which is the complete row.

**Testing.** Unit tests: completeness matrix (both families, dirty IPv4 buffer tail, UDP dest_port 0 with a real address, both halves of the peer test required), the port-less lifecycle (unchanged behavior), the new peer-less lifecycle (one upgrade, nothing after), degraded-only lifecycle (one record), complete-first (one record), independence. `cargo fmt --check`, `cargo test --lib` (473 pass), `cargo clippy --lib --tests -D warnings` clean. No BPF change; trace size grows by one ~100-byte record per in-window inbound connection on <6.15 kernels. Precedent search over the recorder's history for this class: none beyond eb368f3, which introduced the upgrade mechanism this generalizes.

### DIFF
--- a/src/network_recorder.rs
+++ b/src/network_recorder.rs
@@ -312,14 +312,19 @@ pub struct NetworkRecorder {
 
     // Streaming support fields
     /// Track which sockets have had their NetworkSocketRecord emitted,
-    /// and whether that record carried a bound (nonzero) src_port. A
-    /// socket whose identity is created by a pre-bind state transition
-    /// (connect-start fires before the ephemeral port is assigned)
-    /// emits its first record with src_port=0; when a later event
-    /// arrives carrying the bound tuple, ONE upgraded record is
-    /// emitted for the same socket_id so the trace carries the real
-    /// tuple. Readers keep one row per socket by preferring the
-    /// bound-port row.
+    /// and whether that record carried a COMPLETE tuple (see
+    /// [`socket_tuple_is_complete`]). Two event shapes create a socket's
+    /// identity with a degraded tuple: a pre-bind state transition
+    /// (connect-start fires before the ephemeral port is assigned, so
+    /// src_port=0) and — on kernels before 6.15 — the LISTEN→SYN_RECV
+    /// transition of a passive open, which fires before the kernel has
+    /// copied the peer into the child socket (dest = the listener's
+    /// wildcard `0.0.0.0:0` / `[::]:0`; Linux commit a3a128f611a9
+    /// reordered it in 6.15). Either way the first record carries the
+    /// degraded tuple; when a later event on the same socket_id
+    /// arrives with the complete tuple, ONE upgraded record is emitted
+    /// so the trace carries the real endpoints. Readers keep one row
+    /// per socket by preferring the complete row.
     seen_sockets: HashMap<SocketId, bool>,
     /// Collector for streaming records during recording
     streaming_collector: Option<Box<dyn RecordCollector + Send>>,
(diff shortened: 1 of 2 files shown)


josefbacik:
**Code review — rust-bpf-code-reviewer pass over the full diff** (mechanics: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session by the author over `git diff` of the two files, five passes as the definition lists them; not an independent human reviewer).

### Summary
Small, userspace-only change that generalizes an existing one-shot record upgrade from a src_port-only completeness test to a two-endpoint test, with the decision function kept pure and the address decode routed through the existing `parse_ip_addr`. Tests cover the matrix and each lifecycle. No BPF, schema, or version changes.

### Critical Issues
None.
- Schema & versioning: no `create_schema()` / `schema.rs` change → no `SCHEMA_VERSION` bump, no `SCHEMA_CHANGES.md` entry, no `[bump minor]` needed. `Cargo.toml` `version` untouched.
- Recorder consistency: no new table or per-event column; utid handling untouched.
- BPF: untouched. (Map/context pairing check: no map writers change; the perf_event/tp_btf pairing is as before.)
- ELF/binary parsing: none.

### Improvements
- (Kept as-is, noted for the human reviewer) `should_emit_socket_record` now takes a `complete: bool`. A bool parameter is a mild readability cost, but the only call site computes it on the line above from `socket_tuple_is_complete`, and keeping the decision function free of address representation is what lets the tests name the three sighting shapes (`portless` / `peerless` / `full`) instead of raw tuples. A two-variant enum would be more type-safe but is over-engineering for a private function with one caller.
- The in-tree analyzers group `network_socket` by (socket_id, tuple); a passive-open child on a <6.15 kernel now yields two rows there, exactly as the port-less double already does. Not introduced by this change; a dedupe-to-complete-row follow-up in `analyze/` is the natural next step if that output matters.

### Suggestions
- None beyond the above; comments explain why (kernel ordering, family-aware decode), reference no source-line numbers, and the test constants document what the byte layouts stand for.

### Positive Notes
- The completeness test is decoded through `parse_ip_addr` rather than a raw all-zero-bytes check, so it cannot disagree with the `dest_ip` string the record carries, and the `UNSPEC_V4_DIRTY_TAIL` case pins that an IPv4 event's unused buffer tail cannot flip the verdict.
- The lifecycle tests keep the original port-less contract byte-for-byte in behavior and add the peer-less, degraded-only, and complete-first arms, so the "exactly one upgrade, never a downgrade" invariant is pinned from every side.

PR #198, opened 2026-08-18
analyze: render folded flamegraph frames with the continuous-profiling grammar

josefbacik:
`systing analyze` (the `stacktrace flamegraph` CLI subcommand and the MCP `flamegraph` tool) rendered frames in its own grammar, `func [module]`, while systing's continuous-profiling lane renders `file:function:line`. The two disagreed on every frame, and the folded writer wrapped the recorder's already-bracketed labels a second time, so kernel frames read `unknown [[kernel]]`. This PR makes the folded writer use the continuous lane's grammar, byte for byte, so a frame renders the same in every mode and the double bracket cannot occur. What you are approving is the grammar switch below plus one open question at the end (minor version bump or not).

**Per frame class — recorder text → what analyze wrote before → what it writes now (= what continuous profiling already writes)**

| Frame class | Recorder frame text | Before | Now |
|---|---|---|---|
| symbolized, source known | `tcp_sendmsg (vmlinux [net/ipv4/tcp.c:1234]) <0xffffffff8ec7559e>` | `tcp_sendmsg [vmlinux]` | `net/ipv4/tcp.c:tcp_sendmsg:1234` |
| symbolized, file without line | `wait_until_cold (myserver [registry.rs]) <0x57ee0e367975>` | `wait_until_cold [myserver]` | `registry.rs:wait_until_cold` |
| symbolized, no source | `main (myapp) <0x401234>` | `main [myapp]` | `myapp:main` |
| kernel, symbolized | `do_syscall_64 ([kernel]) <0xffffffff96f461b1>` | `do_syscall_64 [[kernel]]` | `[kernel]:do_syscall_64` |
| kernel, unsymbolized | `unknown ([kernel]) <0xffffffff8e000000>` | `unknown [[kernel]]` | `[kernel]:unknown` |
| pseudo-mapping labels (`[guest]`, `[exited]`, `[gvisor:*]`, `[jit:*]`, `[anon]`, `[buildid:…]`) | `unknown ([guest])` | `unknown [[guest]]` | `[guest]:unknown` |
| bare address, no mapping | `0x5dbfa1` | `0x5dbfa1 [unknown]` | `:0x5dbfa1` |
| unparseable text / empty name | `something weird` / `` | `something weird` / `[unknown]` | `:something weird` / `:0x0` |

Two rows look unusual on first read and are deliberate: a frame with neither mapping nor symbol renders with an empty file slot and a leading colon (`:0x5dbfa1`), and a kernel frame that carries source info renders the file rather than the `[kernel]` label. Both are exactly what the continuous lane renders today, and both are pinned by tests.

**What changes for readers**
- CLI `systing-analyze stacktrace flamegraph`: folded lines are still `frame;frame;… count`, root→leaf; only the per-frame text changes.
- MCP `flamegraph` tool: `stacks[].frames` and `folded` carry the new strings; nothing else in the result changes.
- Anything that reads frames from the trace database or from continuous-profiling output directly is unaffected — the recorder's frame text is not touched.
- Consumers that string-match the old spelling (in particular the doubled `[[kernel]]` / `[[exited]]` forms) see the new spelling from this version on. Consumers that treat frames as opaque text need no change.

<details>
<summary>How the parse works (why a regex, why anchored at the end)</summary>

The recorder writes one of three shapes: `func (mapping [file:line]) <0xpc>`, `func (mapping) <0xpc>`, `0xpc`. Function names can contain unbalanced parens and angle brackets (Rust generics, C++ templates), so the previous left-to-right split on the first `" ("` truncated such names. `parse_frame` now strips the trailing `<0xpc>`, then matches ` (mapping [file:line])` anchored at the end of the text — the same pattern the continuous lane uses — and everything before the match is the function name. `format_frame` then renders `file:function:line` with `file` falling back to the mapping and `function` falling back to `0x{pc}`. The parts borrow from the input; the only allocation is the output string. `regex` was already a dependency; the pattern is compiled once via `LazyLock`, the pattern used elsewhere in the tree.
</details>

<details>
<summary>Checks</summary>

- `cargo fmt --all -- --check`, `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`: clean; 488 passed / 0 failed. 25 tests in `analyze::flamegraph`, 20 new or rewritten: one per recorder shape (including a Rust generic name, a Go `<autogenerated>:1` source, file-without-line, a label with no address, unparseable text) and one per rendered class, plus folded-stack cases mixing all of them and a missing-frame-id case.
- The `tests/analyze_commands.rs` folded-output validator is grammar-agnostic (frames non-empty, count parses) and runs in CI against a recorded trace.
- Reviewer pass per the repo's `rust-bpf-code-reviewer` definition: no critical issues; verdict posted as a comment on this PR.
- No BPF or map changes; no trace-database schema change; `Cargo.toml` untouched.
</details>

**Open question:** this is a user-visible change to `systing analyze` output but not a database schema change, so per the repo's bump policy the title carries no `[bump minor]`. Say so if you want a minor bump anyway and I'll add the token.

### DIFF
--- a/src/analyze/flamegraph.rs
+++ b/src/analyze/flamegraph.rs
@@ -1,8 +1,10 @@
 use anyhow::{bail, Result};
+use regex::Regex;
 use serde::Serialize;
 use std::collections::HashMap;
-use std::fmt;
+use std::fmt::{self, Write as _};
 use std::str::FromStr;
+use std::sync::LazyLock;
 
 use super::AnalyzeDb;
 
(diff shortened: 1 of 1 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this branch (1 file: `src/analyze/flamegraph.rs`), all five passes. One finding raised during the review was fixed before the commit (noted below); what follows is the verdict on the code as pushed.

### Summary
The folded-flamegraph writer's `format_frame` is replaced by a two-step port of the continuous-profiling frame grammar: `parse_frame` splits the recorder's frame text (`func (mapping [file:line]) <0xpc>` / `func (mapping) <0xpc>` / `0xpc`) into borrowed parts by anchoring on the trailing pc and mapping, and `format_frame` renders `file:function:line` with the same fallbacks the continuous lane uses (mapping when the source location is unknown, `0x{pc}` when unsymbolized, `:line` omitted when zero). The `[[kernel]]` double bracket disappears as a consequence — labels are never re-wrapped — and every systing mode now renders a given frame the same way. Test coverage goes from five shape-pins to twenty-five cases covering each recorder shape and each rendered class.

### Critical Issues
None.
- Schema & versioning: no table/schema definitions touched (this changes the text of `systing analyze` folded output and the MCP `flamegraph` tool's `frames`, not the trace database) → no `SCHEMA_VERSION` bump and no `[bump minor]` per the repo's policy; `Cargo.toml`/`Cargo.lock` untouched (diff stat: one file).
- BPF: no BPF code, no map operations, nothing under `src/bpf/` → the map/context-pairing (NMI-writer) check does not apply.
- Recorder consistency, Python version support, ELF parsing: not applicable. No `unsafe`.

1. `parse_frame` first computed the function slice as `&head[..c.get(0).map_or(head.len(), |m| m.start())]` — a fallback for a capture that cannot be absent. Simplified to `&head[..head.len() - c[0].len()]` with a one-line comment stating the invariant it relies on (the pattern is `$`-anchored, so the match ends at the end of `head`).

- Two rendered classes look different from before and are worth a sentence in the PR body so nobody reads them as bugs: a frame with neither mapping nor symbol renders `:0x5dbfa1` (empty file slot, then the address — the continuous lane's shape), and a kernel frame that carries source info renders `net/ipv4/tcp.c:tcp_sendmsg:1234` (the file wins over the `[kernel]` label, again as the continuous lane does). Both are pinned by tests.
- `Frame`/`parse_frame`/`format_frame` are private to the module. If a later change wants one implementation shared with the continuous lane instead of two identical ones, this pair is the natural crate API to expose; nothing in this diff prevents that.
- `let _ = write!(out, …)` on a `String` discards an error that cannot occur; `.expect("write to String")` would say the same thing more loudly. Style only.

### Positive Notes
- Anchoring the parse on the trailing `<0xpc>` and `(mapping …)` instead of the first `" ("` fixes a latent truncation for function names that contain `" ("` (Rust generics, C++ templates) — the previous left-to-right parse cut such names at the first paren; `parse_frame_full` pins the correct behavior on a real Rust generic name.
- The regex is compiled once through `LazyLock` (the house pattern in `trace/constants.rs`) and the parts borrow from the input; the only allocation is the output string.
- The doc comment carries the input shapes and a worked table of outputs, so a reader can see the grammar without running it.

Gates on the pushed tree: `cargo fmt --all -- --check` clean; `cargo build` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 488 passed / 0 failed (25 tests in `analyze::flamegraph::tests`, 20 new or rewritten). The `tests/analyze_commands.rs` folded-output validator is grammar-agnostic (frames non-empty, `count` parses) and runs in CI.

PR #200, opened 2026-08-20
analyze: add sched aggregate, a per-capture scheduler summary for A/B comparisons

josefbacik:
This adds `systing-analyze sched aggregate` (and the matching `sched_aggregate` MCP tool): one per-capture scheduler summary you can put side by side for two schedulers or two hosts — wakeup latency, preempt wait, on-CPU slice length, context-switch and migration rates, a time-weighted runqueue-length distribution, per-CPU load vectors with imbalance metrics, log2 histograms next to every percentile, and the threads that dominate each latency tail. It reads the tables the existing `sched` recorder already writes; no recorder, BPF, or schema change. What you are deciding is the set of statistics and their definitions (below); the open choices are at the end.

**Why this shape.** `sched stats` answers per-thread questions and `sched cpu-stats` per-CPU utilization; an A/B of scheduler policy needs one number per question per capture, mergeable across captures. The "one CPU gets everything" failure is invisible to aggregate percentiles, so the summary carries per-CPU vectors and three imbalance metrics (busiest-CPU share, coefficient of variation, and the time some CPU had a queue while another idled) alongside the averages.

**Definitions** (all ns; the module docs carry the same text):
| statistic | definition |
|---|---|
| wakeup latency | first `sched_slice.ts` of the thread at/after its runnable marker (`thread_state.state=0`, with the wakeup's target CPU) minus the marker ts; split same-CPU / cross-CPU by whether it ran on the target |
| preempt wait | slice ended with `end_state` NULL (still runnable) → wait until the thread's next slice |
| run delay | wakeup latency ∪ preempt wait |
| slice length | `sched_slice.dur` of non-idle slices |
| switches | non-idle slice starts; voluntary = ended in a sleep state, involuntary = preempted |
| migrations | consecutive slices of one thread on different CPUs; "wakeup" if a marker came between them, else "preempt" (moved while runnable) |
| runqueue length | per CPU: 1 if running a non-idle task + threads runnable-but-not-running assigned to it (woken → target CPU; preempted → the CPU it was on) until they run; time-weighted over (cpu × time) |
| work-conservation violation | time during which some observed CPU has runqueue length > 1 while another is idle |
| per-CPU vectors | observed / busy / idle / switches / wakeups targeted / wakeups ran / runnable wait / rq avg, index = CPU id |
| tail contributors | threads (by name) with waits above each distribution's p99 bucket: count exact at bucket granularity, summed ns approximated by bucket upper edges |

**Honesty notes built into the output.** Per-CPU runqueue lengths are approximate: the trace has no migrate event, so a runnable thread the balancer moves before it runs is still counted on its previous CPU (the node total is exact; a `sched_migrate_task` recorder would make per-CPU exact — a possible follow-up). Per-CPU accounting starts at the CPU's first event in the window; CPUs with no events are listed as `unobserved_cpus`, never assumed idle or busy. Waits that do not end inside the window are counted (`wakeup_censored`, `preempt_censored`), not extrapolated. `spurious_wakeups` counts markers for threads already running or runnable; `first_runs` counts slices with no preceding marker (threads running at window start). `missed_sched_events` is the recorder's own drop counter when the trace carries it. Percentiles come from a log-linear histogram (16 sub-buckets per octave, upper bucket edge reported, never above the exact max) — within ~6% of exact; count/sum/min/max/avg are exact.

<details>
<summary>Output shape (JSON)</summary>

`meta` (window, ncpu, observed_cpus, slices, runnable_markers, wakeup_new, threads_seen, wakeup_censored, preempt_censored, spurious_wakeups, first_runs, missed_sched_events, aggregate_ms) · `wakeup_latency` / `wakeup_latency_same_cpu` / `wakeup_latency_cross_cpu` / `preempt_wait` / `run_delay` / `slice_len` (each: count, sum_ns, avg_ns, min_ns, p50_ns, p90_ns, p99_ns, max_ns, hist_log2[32] — bucket i covers [2^(8+i), 2^(9+i)) ns, bucket 0 also holds smaller values, bucket 31 also holds larger) · `switches` (switches, switches_per_s, voluntary, involuntary, migrations, migrations_per_s, wakeup_migrations, preempt_migrations) · `runqueue` (observed_cpu_ns, avg, p50, p90, p99, max, hist[len] = observed ns per length) · `per_cpu` (the vectors + unobserved_cpus) · `imbalance` (busy_max_share, busy_cv, wakeups_max_share, wakeups_cv, work_conservation_violation_ns, _frac) · `wakeup_tail_top` / `preempt_tail_top` ([{comm, count, sum_ns}]).
</details>

<details>
<summary>Implementation and checks</summary>

- `src/analyze/sched_aggregate.rs`: a DuckDB `UNION ALL … ORDER BY ts, kind` stream of slice ends, slice starts, runnable markers and `wakeup_new`; a per-thread state machine (unknown / running / waiting{wakeup|preempt} / sleeping) and per-CPU accumulators; runqueue time-weighting and the work-conservation integral are updated at every queue change; tail attribution fills per-thread-name histograms in the same pass and reads each name's count above the overall p99 bucket at the end. Memory is O(threads + cpus + histogram buckets), independent of event count.
- CLI `sched aggregate --database … [--trace-id] [--start-time/--end-time s] [--top-k N] [--format json]`; MCP tool `sched_aggregate` with the same parameters; the MCP tool-set test lists it.
- Tests: 9 new (in-memory trace databases with known answers: same-CPU wakeup; cross-CPU wakeup + preemption + migration; all-on-CPU-0 window tripping the imbalance metrics and the work-conservation integral; censored wait + first run; tail attribution; multi-trace guard; histogram bucket edges and percentiles; query shape). `cargo fmt --check`, `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` (497 passed) clean.
- Reviewer pass per the repo's `rust-bpf-code-reviewer` definition: no critical issues; verdict posted as a comment.
- No BPF, map, or trace-schema change; `Cargo.toml` untouched. Timing on a synthetic 192-CPU, 10 s trace with 38.4 M `sched_slice` rows (half of them idle) + 19.2 M runnable markers — about 10 k task switches and 10 k wakeups per second per CPU, heavier than a typical host: release binary 15.6 s with tail attribution, 13.4 s with `--top-k 0` (the debug binary takes 251 s). The hot loop reads rows through `duckdb::Row::get` (~4 M events/s); Arrow batch iteration would cut that several-fold if per-capture cost ever matters.
</details>

**Open choices (say the word and it changes):** (1) tail grain by thread name (now) vs by process; (2) add a `sched_migrate_task` event to the recorder in a follow-up so per-CPU runqueue and migration attribution become exact; (3) Arrow-batch row iteration in the hot loop if the per-capture cost on large hosts needs to come down.

### DIFF
--- a/skills/systing-analyze/SKILL.md
+++ b/skills/systing-analyze/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: systing-analyze
-description: Analyze a systing trace database (.duckdb). Use when the user asks about a systing trace — flamegraphs, scheduling latency, CPU hotspots, network behavior, memory/allocation activity, off-CPU time, TPU op/metric data, or any question about what's in a trace.duckdb file. Orchestrates the systing-analyze MCP tools (trace_info, query, flamegraph, sched_stats, cpu_stats, network_*).
+description: Analyze a systing trace database (.duckdb). Use when the user asks about a systing trace — flamegraphs, scheduling latency, CPU hotspots, network behavior, memory/allocation activity, off-CPU time, TPU op/metric data, or any question about what's in a trace.duckdb file. Orchestrates the systing-analyze MCP tools (trace_info, query, flamegraph, sched_stats, cpu_stats, sched_aggregate, network_*).
 ---
 
 # Analyzing systing traces
@@ -23,6 +23,7 @@ Systing stores traces in **DuckDB**. The `systing-analyze` MCP server exposes st
 | Why is the process blocked / off-CPU? | `flamegraph` | `stack_type="uninterruptible-sleep"` (D) or `"interruptible-sleep"` (S), or `"all-sleep"` |
 | Is the scheduler oversubscribed? Latency? | `sched_stats` | no filter = whole-trace ranking; `pid` = per-thread breakdown; `tid` = single thread with end-state distribution |
 | Which CPUs are busy / idle? | `cpu_stats` | per-CPU utilization, idle%, IRQ/softIRQ time, runqueue depth p50/p90/p99 |
+| How does the scheduler behave overall? Compare two schedulers/hosts? | `sched_aggregate` | one summary per capture: wakeup latency (same/cross-CPU), preempt wait, slice length, switch + migration rates, time-weighted runqueue length, per-CPU load vectors + imbalance metrics, log2 histograms, tail threads; percentiles within ~6%, per-CPU runqueue approximate |
 | What's the network doing? | `network_connections` | per-connection bytes, retransmit % |
 | Interface-level network? | `network_interfaces` | per-interface, per-protocol breakdown |
 | Both sides of a connection (multi-node)? | `network_socket_pairs` | matched socket pairs, within or across traces |
@@ -30,7 +31,7 @@ Systing stores traces in **DuckDB**. The `systing-analyze` MCP server exposes st
 | What ran on the TPU? Duty cycle? HBM? | `query` | no dedicated tool — see TPU schema below |
 | Anything else | `query` | raw SQL; see schema below |
 
-There is also a CLI with the same analyses: `systing-analyze query|stacktrace flamegraph|sched stats|sched cpu-stats|network connections|network interfaces|network socket-pairs` (and `systing-analyze mcp` to start this server).
+There is also a CLI with the same analyses: `systing-analyze query|stacktrace flamegraph|sched stats|sched cpu-stats|sched aggregate|network connections|network interfaces|network socket-pairs` (and `systing-analyze mcp` to start this server).
 
 ## Key schema for `query`
 
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -10,6 +10,7 @@ mod network_connections;
 mod network_interfaces;
 mod network_socket_pairs;
 mod query;
+mod sched_aggregate;
 mod sched_stats;
 
 pub use cpu_stats::{CpuStatsParams, CpuStatsResult, CpuStatsSummary, PerCpuStats};
@@ -25,6 +26,10 @@ pub use network_interfaces::{
 };
 pub use network_socket_pairs::{
     NetworkSocketPairsParams, NetworkSocketPairsResult, SocketPair, SocketSide,
+};
+pub use sched_aggregate::{
+    Dist, Imbalance, PerCpu, RqDist, SchedAggregate, SchedAggregateMeta, SchedAggregateParams,
+    SwitchStats, TailContributor, HIST_BASE_LOG2, HIST_OCTAVES,
 };
 pub use sched_stats::{
     EndStateCount, ProcessSchedStats, SchedStatsParams, SchedStatsResult, SchedSummary,
--- a/src/analyze/sched_aggregate.rs
+++ b/src/analyze/sched_aggregate.rs
@@ -1,0 +1,1448 @@
+//! Per-capture scheduler aggregates (`systing-analyze sched aggregate`).
+//!
+//! One pass over the scheduler event stream of a trace (`sched_slice` starts
+//! and ends, `thread_state` runnable markers, `wakeup_new`) produces a compact
+//! summary meant to be emitted once per capture and compared across hosts or
+//! schedulers: wakeup latency, preempt wait, on-CPU slice length, context
+//! switch and migration rates, a time-weighted runqueue-length distribution,
+//! per-CPU load vectors with imbalance metrics, log2 histograms beside every
+//! percentile, and the threads that contribute most to the latency tail.
+//!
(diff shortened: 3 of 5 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this branch (5 files: `src/analyze/sched_aggregate.rs` (new), `src/analyze/mod.rs`, `src/mcp.rs`, `src/bin/systing_analyze.rs`, `skills/systing-analyze/SKILL.md`), all five passes. Three findings raised during the review were fixed before the commit (listed below); what follows is the verdict on the code as pushed.

### Summary
A new analysis, `sched aggregate`, that turns a trace's scheduler event stream into one per-capture summary for comparing schedulers or hosts: wakeup latency, preempt wait, run delay, slice length, switch/migration rates with splits, a time-weighted runqueue-length distribution, per-CPU load vectors with imbalance metrics, log2 histograms beside every percentile, and the thread names above each latency tail's p99. It is a single ordered pass over a DuckDB `UNION ALL … ORDER BY ts, kind` stream with a per-thread state machine and per-CPU accumulators; memory is O(threads + cpus + histogram buckets), independent of event count. Every definition is written in the module docs in the same words as the PR body, and every number that could mislead carries its caveat in the output itself (censored counts, unobserved CPUs, spurious wakeups, first runs, the recorder's missed-event counter). CLI subcommand, MCP tool, and the skill doc are wired; the MCP tool-set test lists the tool.

### Critical Issues
None.
- Schema & versioning: no trace table/schema definition touched → no `SCHEMA_VERSION` bump, no `[bump minor]`; `Cargo.toml`/`Cargo.lock` untouched (diff stat: five files, none of them manifests).
- BPF: no BPF code, no map, no recorder change → the map/context-pairing (NMI-writer) check does not apply to this diff. The analysis reads tables the existing `sched` recorder already writes.
- Recorder consistency: the analysis joins on `utid` (and `thread.tid` only to find the idle thread), never on raw tid/pid. Python-version and ELF criteria: not applicable. No `unsafe`.

1. Per-CPU "running" state was a bool; an overlapping or zero-length record for one thread could clear another thread's CPU. Now `CpuState.running: Option<utid>` and a release only applies when the CPU still runs that thread.
2. Tail attribution used a second full pass over the event stream (it doubled the runtime: 26.8 s vs 13.4 s on the synthetic 38 M-slice trace). Replaced by per-thread-name log-linear histograms filled in the single pass; the p99 bucket is read off the overall histogram at the end and each name's count above it is summed (exact at bucket granularity; `sum_ns` approximated by bucket upper edges, stated in the doc). Runtime with attribution: 15.6 s on the same trace.
3. `LogHist::bucket_upper(0)` returned 256 ns although bucket 0 also holds [256, 272) ns; now the general formula applies (272 ns) and the edge test pins every bucket's lower and upper value mapping back to it.

- The hot loop pulls rows through `duckdb::Row::get` per column (~4 M events/s in release). If the per-tick cost on large hosts matters, fetching Arrow record batches and iterating the columns would cut it several-fold without changing the algorithm.
- Per-CPU runqueue length is approximate because the trace has no migrate event; adding `sched_migrate_task` to the recorder (its own PR) would make per-CPU lengths and the wakeup/preempt migration split exact.
- `sched aggregate` now sits beside `sched cpu-stats`, whose runqueue percentiles are event-weighted and exclude preempted threads (its own docs say so); a one-line cross-reference in the `cpu-stats` help text would steer readers to the time-weighted numbers here. Left for a doc follow-up.

### Positive Notes
- The work-conservation-violation integral and the node-wide over/idle counters are updated incrementally at every queue change (one CPU per event), so the "one CPU queues while another idles" signal costs nothing extra and is exact for the observed CPUs.
- Window edges are handled by counting rather than guessing: per-CPU accounting starts at the CPU's first event, CPUs with no events are listed as unobserved, waits that do not end inside the window are censored, and a narrower window drops edge-crossing slices and says so in the params docs.
- Tests build in-memory trace databases with hand-computed answers for each statistic (same-CPU and cross-CPU wakeups, a preemption and a migration, an all-on-CPU-0 window that trips the imbalance metrics, censored waits, tail attribution, the multi-trace guard) and pin the histogram bucket edges — the module is verified by its outputs, not by reading its SQL.

Gates on the pushed tree: `cargo fmt --all -- --check` clean; `cargo build` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 497 passed / 0 failed (9 new tests). Release-binary timing on a synthetic 192-CPU, 10 s trace with 38.4 M slices + 19.2 M runnable markers (~10 k switches/s/CPU): 15.6 s with tail attribution, 13.4 s with `--top-k 0` (debug build: 251 s).

PR #201, opened 2026-08-20
analyze: sched aggregate — the runnable marker's CPU is the previous CPU, not the placement

josefbacik:
Follow-up to #200: `sched aggregate` assumed a runnable marker's `cpu` was the CPU the woken task was placed on. It is the CPU the task **last ran on** — the recorder stamps it at `sched_waking`, which the kernel fires before `select_task_rq` picks a CPU (`wakeup_new` markers do carry the placement; plain wakeups do not). Wakeup latency was right; the per-CPU queue metrics built on that CPU were not. This PR re-states those metrics as what the trace can support, renames the affected fields so the old meaning cannot be read by accident, and fixes two smaller things a cold review found. Nothing to decide beyond accepting the new semantics; the open question at the end is about a recorder follow-up.

**What changes (JSON field names included, since a consumer is being built against them):**
| before | now | meaning |
|---|---|---|
| `wakeup_latency_same_cpu` / `_cross_cpu` | `wakeup_latency_prev_cpu` / `wakeup_latency_migrated` | the woken thread ran on the CPU it last ran on / somewhere else |
| `per_cpu.wakeups_targeted` | `per_cpu.wakeups_prev_cpu` | markers by the thread's previous CPU |
| `imbalance.wakeups_max_share` / `_cv` over targeted | over `per_cpu.wakeups_ran` | where wakeups actually landed (exact: the CPU of the first slice after the marker) |
| per-CPU runqueue = running + woken waiters (on the marker CPU) + preempted waiters | running + **preempted** waiters only; woken threads are unplaced until they run | per-CPU lengths are now exact for what they claim; the unplaced count is reported node-wide as `runqueue.unplaced_avg` (time-weighted) |
| work-conservation violation counted woken waiters on the marker CPU | counts preempted waiters only | "a preempted thread waits behind a running one while another CPU idles" — the balancer-failure shape, not the placement shape |
| `per_cpu.runnable_wait_ns` by marker CPU | by the CPU the thread then ran on (wakeups) / was preempted on | |
| trace-end closing slices counted as preemptions (+1 involuntary, +1 `preempt_censored` per busy CPU) | `meta.running_at_end` | the recorder closes still-running threads' slices with no end state at trace end |
| marker at the same ns as the slice it precedes → `first_run` + `spurious_wakeup` | ordered before the slice start → 0 ns wait | |

Docs tightened in the same pass: percentiles are at most 6.25% above exact (upper bucket edge), `busy_max_share` reads 1/observed_cpus when even, `threads_seen` excludes the idle thread, `wakeup_new` markers are part of wakeup latency.

<details>
<summary>Checks</summary>

- `cargo fmt --all -- --check`, `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`: clean; 499 passed / 0 failed (the three semantic tests rewritten for the new meaning, two new: running-at-end, marker-at-slice-start).
- Reviewer pass per the repo's `rust-bpf-code-reviewer` definition: no critical issues; verdict posted as a comment.
- No BPF, map, or trace-schema change; `Cargo.toml` untouched.
- Kernel reference for the premise: `kernel/sched/core.c` `try_to_wake_up()` — `trace_sched_waking(p)` fires before `select_task_rq()`/`set_task_cpu()`; `trace_sched_wakeup` (post-placement) is not attached by the recorder; `sched_wakeup_new` fires after placement.
</details>

**Open question (not done here):** exact per-CPU placement for woken threads needs a recorder event — `sched_migrate_task` (fires on `set_task_cpu` at wakeup and on balancer moves; cheaper, also makes migrations exact) or `sched_wakeup` (post-placement `task_cpu`). Say which, if either, and it becomes its own recorder PR with the event cost measured.


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this branch (3 files: `src/analyze/sched_aggregate.rs`, `src/bin/systing_analyze.rs`, `src/mcp.rs`), all five passes. One finding raised during the review was fixed before the commit; what follows is the verdict on the code as pushed.

### Summary
A semantics correction to `sched aggregate`: the runnable marker's `cpu` is the woken thread's previous CPU (recorded at `sched_waking`, before placement), so the metrics that assumed it was the placement are re-stated — the same/cross split becomes prev-CPU/migrated, wakeup imbalance is computed over actual placement (`wakeups_ran`), woken threads are unplaced until they run (per-CPU runqueue = running + preempted waiters; a node-wide `unplaced_avg` keeps the total), and work-conservation violation counts preempted waiters only. Two smaller fixes ride along: trace-end closing slices are `running_at_end` rather than preemptions, and markers sort before slice starts at equal timestamps. Field renames make the old meaning unreadable by accident; the module docs and the MCP description say exactly what each number now is.

### Critical Issues
None.
- Schema & versioning: no trace table/schema change (this is the text/JSON of an analysis) → no `SCHEMA_VERSION` bump, no `[bump minor]`; `Cargo.toml`/`Cargo.lock` untouched.
- BPF: no BPF code, no map → the map/context-pairing check does not apply.
- Recorder consistency, Python, ELF: not applicable. No `unsafe`.

1. The first version of `on_runnable` touched the marker's CPU (`touch_cpu`) to record `wakeups_prev_cpu`, which would have registered a CPU as observed — and idle — on the strength of a marker alone. Now the counter is bumped without observing the CPU, and the assembly copies `wakeups_prev_cpu` for unobserved CPUs too (a count is a fact about the CPU; a time integral is not).

- `runnable_wait_ns` now mixes two attributions (wakeup waits by run CPU, preempt waits by preempting CPU); the field doc says so. If a consumer wants them apart, two vectors are a small follow-up.
- Exact placement needs a recorder event; `sched_migrate_task` is the cheaper of the two options (fires only on CPU changes, covers balancer moves too). Named in the PR body as the open question.

### Positive Notes
- The invariant "per-CPU runqueue + unplaced = all runnable threads" is kept explicit: the per-CPU integral and the node-wide unplaced integral are both time-weighted through the same `advance_global` step, so the two numbers are comparable.
- The rewritten tests pin the new meaning with hand-computed answers (woken thread unplaced while waiting; preempted waiter behind a running thread = the only violation shape; running-at-end not a preemption; marker at slice start = 0 ns wait).

Gates on the pushed tree: `cargo fmt --all -- --check` clean; `cargo build` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 499 passed / 0 failed.

PR #202, opened 2026-08-20
Fedora packaging prep: upgrade indicatif

michel-slm:
0.18 got its bump because the console upgrade was considered semver breaking:

https://github.com/console-rs/indicatif/releases#release-0.18.0

That change is this

https://github.com/console-rs/console/releases#release-0.16.0

> The 0.15.12 release was yanked after it turned out to be semver-incompatible with existing usage by several of the most popular dependent crates, because it introduced a std feature -- and those crates used default-features = false but relied on the std-guarded features.

Given the tests all pass we should be good to go

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -920,15 +920,14 @@ dependencies = [
 
 [[package]]
 name = "console"
-version = "0.15.11"
+version = "0.16.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


josefbacik:
@michel-slm just need you to fix the clippy things and then it should be good.

josefbacik (commented):
The Clippy failure isn't from this change — it's `clippy::chunks_exact_to_as_chunks` at `src/stack_recorder.rs:654`, already fixed on main in fec6b1b; a rebase onto main clears it.

PR #203, opened 2026-08-20
sched: record sched_migrate_task so wakeup placement and migrations are exact [bump minor]

josefbacik:
**Stacks on #201 — merge #201 first.** This PR is based on #201's branch, so the diff shown here is (B) only: the recorder event and the analysis changes on top of #201's semantics. GitHub retargets it to `main` when #201 merges; if #201 changes in review, this rebases after.

Records the kernel's `sched_migrate_task` tracepoint as a new `sched_migrate` table and uses it in `sched aggregate`, so the placement of woken threads — and every move of a runnable thread — is exact. This is the recorder follow-up #201 left open, answered with `sched_migrate_task` (it fires on every CPU change, so balancer moves become visible too, and it costs one event per change rather than one per wakeup). Schema 15 → 16, so the title carries `[bump minor]`. Nothing to decide beyond accepting the table; one optional follow-up is named at the end.

**Why.** The runnable marker comes from `sched_waking`, which fires before the kernel picks a CPU, so its `cpu` is where the thread last ran. Until the thread runs, the trace cannot say which CPU it was queued on, and a move of a runnable thread is invisible — `sched aggregate` had to keep woken threads unplaced (#201). `sched_migrate_task` fires from `set_task_cpu()` at the moment the CPU changes: right after `sched_waking` when the scheduler picked another CPU, and on load-balancer / NUMA / affinity moves of runnable tasks. A woken thread with no migrate event before its next slice stayed on its previous CPU.

**What changes**
- BPF: `tp_btf/sched_migrate_task` → `systing_sched_migrate`, in the sched recorder's program set (loaded only with `sched`), `trace_task()`-gated; one `task_event` of the new type `SCHED_MIGRATE` (appended to the enum, existing values unchanged) per event with `cpu` = the CPU the task leaves and `target_cpu` = `dest_cpu`. Writes only the task ringbuf (tp_btf context), `handle_missed_event(MISSED_SCHED_EVENT)` on a full ring like its siblings; no new map. The migrated task rides in `next` (sighted like a wakee for process discovery; `prev` stays zeroed and is not read).
- Table `sched_migrate (trace_id, ts, utid, orig_cpu, dest_cpu)`: the `wakeup_new` wiring mirrored through model, arrow schema, parquet writer, collector, paths, DuckDB DDL/import/export, stream allowlist and util import. `SCHEMA_VERSION` 16 + `SCHEMA_CHANGES.md` entry. The Perfetto converter does not carry the rows yet.
- `sched aggregate` (`meta.placement_exact` = the table exists): a woken thread is queued on its previous CPU from its marker on and moved by the events; no event before its first run = the scheduler kept it there. Per-CPU runqueue lengths, the work-conservation violation and `per_cpu.runnable_wait_ns` (now the time queued per CPU, split at migrations) include woken waiters, and `runqueue.unplaced_avg` reads 0. New fields, all additive: `per_cpu.wakeups_placed` (`Option`: the scheduler's placement per CPU; `wakeups_ran` stays where the thread ran), `switches.migrate_events`, `migrate_events_per_s`, `migrate_at_wakeup`, `migrate_while_runnable`, `migrate_other`, `meta.placement_exact`, `meta.migrate_mismatch` (events whose `orig_cpu` disagreed with the tracked queue — a dropped event in between). Existing fields keep their meaning on both trace shapes (the slice-based `switches.migrations` is unchanged), and a CPU named only by a marker or a migrate is still never counted as observed — its queue length enters the distributions with its first slice.

**Cost.** One ring record per CPU change. Migrations are a subset of wakeups (those placed elsewhere) plus balancer moves, so expect a fraction of the wakeup event rate; the recorder's missed-event counter covers drops. Captures without the sched recorder pay nothing.

<details>
<summary>Checks</summary>

- `cargo fmt --all -- --check`, `cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib`: clean; 505 passed / 0 failed — 6 new `sched aggregate` tests with hand-computed answers (legacy shape unchanged; a migrate at wakeup places the thread exactly, with the runqueue histogram and violation derived by hand; a runnable move re-queues a preempted waiter with split queue time; no migrate before the first run = previous CPU; `wakeup_new` placement; a migrate for an unseen thread counts as other and observes no CPU) and the `sched.rs` stream round-trip extended with a migrate event.
- The e2e validation suite now asserts `sched_migrate` has rows on multi-CPU hosts and that every row's `utid` exists in `thread` (zero orphans) — the first-contact check for the recorder side, which this sandbox cannot run (no BPF); CI's Test Suite does.
- Legacy-trace identity check: the branch's release binary run on a trace without the table (a 4-CPU / 10 ms synthetic DB, 972 slices) produces output byte-identical to #201's apart from the additive fields (`placement_exact: false`, `wakeups_placed: null`, `migrate_*: 0`) and the wall-clock `aggregate_ms`. The same trace with a derived `sched_migrate` table (77 events) runs the exact path: runnable time is conserved between the two shapes (legacy `runqueue.avg` 0.45053 + `unplaced_avg` 0.83435 / 3 observed CPUs = exact `runqueue.avg` 0.72865) and `per_cpu.runnable_wait_ns` sums to the same total, redistributed by queue.
- Reviewer pass per the repo's `rust-bpf-code-reviewer` definition: verdict posted as a comment. Map/context pairing: the new program writes the task ringbuf only; the `perf_event` sampler writes the stack/memory rings and per-CPU arrays, and `src/bpf/` has no `LRU_HASH`. Hazard-precedent grep (`git log --grep=migrate`): no prior recorder of this event.
- Kernel references: `kernel/sched/core.c` `set_task_cpu()` → `trace_sched_migrate_task(p, new_cpu)` before `__set_task_cpu()` (so `task_cpu(p)` at the tracepoint is the CPU it leaves); `try_to_wake_up()` → `trace_sched_waking` → `select_task_rq` → `set_task_cpu` when the pick differs; `wake_up_new_task()` uses `__set_task_cpu` directly, which is why `wakeup_new` still carries its own placement.
</details>

**Optional follow-up:** the Perfetto converter could emit these rows as `sched_migrate_task` ftrace events (~40 lines mirroring `wakeup_new`) so the UI shows migrations. Say so if wanted.


josefbacik:
**rust-bpf-code-reviewer verdict** — in-session execution of the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition over the full diff of this branch (17 files: `src/bpf/systing_system.bpf.c`, `src/systing_core.rs`, `src/sched.rs`, `src/trace/{models,schema}.rs`, `src/record/collector.rs`, `src/parquet/writer.rs`, `src/parquet_paths.rs`, `src/duckdb.rs`, `src/stream.rs`, `src/bin/systing_util.rs`, `src/analyze/sched_aggregate.rs`, `src/bin/systing_analyze.rs`, `src/mcp.rs`, `skills/systing-analyze/SKILL.md`, `SCHEMA_CHANGES.md`, `tests/trace_validation.rs`), all five passes. Three findings raised during the review were fixed before the commit (listed below); what follows is the verdict on the code as pushed.

### Summary
Records `sched_migrate_task` as a new `sched_migrate` table and teaches `sched aggregate` exact placement from it. The BPF side is one tp_btf program shaped like `sched_waking`'s (reserve the task ring, fill, submit; `trace_task()` gate; part of the sched recorder's set). The storage side mirrors `wakeup_new` through every layer. The analysis keeps every existing field's meaning and adds the placement/migrate fields beside them, switching on the table's presence (`meta.placement_exact`); the state machine gains one event kind and a per-CPU queue membership for woken threads, with the queue time charged by segment.

### Critical Issues
None.
- Schema & versioning: new table → `SCHEMA_VERSION` 15 → 16, `SCHEMA_CHANGES.md` entry present, `[bump minor]` in the PR title; `Cargo.toml`/`Cargo.lock` untouched.
- Recorder consistency: the table attributes by `utid` (through the shared `UtidGenerator`), never raw tid/pid; the e2e validation suite asserts zero utid orphans against `thread` and (on multi-CPU hosts) that the table has rows.
- BPF: the new program writes only the task ringbuf (tp_btf context); the `perf_event` sampler writes the stack/memory rings and per-CPU arrays; no `LRU_HASH` under `src/bpf/`; no new map. The enum value is appended, so existing `event_type` values (and every skeleton consumer) are unchanged.
- ELF / Python criteria: not applicable. No `unsafe`.

1. `on_runnable` duplicated the marker bookkeeping (`advance_global` + `wakeups_prev_cpu`) in both the exact and legacy arms; hoisted, and a new task's placement is now counted at its `wakeup_new` marker (its `target_cpu` is the chosen CPU) instead of at its first run, so a balancer move before it runs is not mistaken for the placement.
2. `on_migrate` carried a legacy-mode conditional (`queued = Preempt || placement_exact`) that could never be false: migrate rows exist only when the table does, which is exactly when `placement_exact` is true. Removed; the function documents that it is only reached with exact placement.
3. A marker or a migrate naming a CPU with no slices in the window must not register that CPU as observed (the #201 review point); the queue membership now goes through `queue_change`, which does the interval bookkeeping only for CPUs already observed and carries the count otherwise — pinned by a test (`migrate_of_an_unseen_thread_counts_as_other_and_never_observes_a_cpu`).

- The migrate row rides the merged stream's `dur` column as `dest_cpu` (CAST to BIGINT) to keep one row shape for the five event kinds; a dedicated column would be clearer at the cost of a wider stream. Left as is and documented at the query and the `handle` arm.
- The Perfetto converter could emit the rows as `sched_migrate_task` ftrace events (mirrors the `wakeup_new` block); a follow-up if the UI view is wanted.
- `per_cpu.wakeups_placed` is `Option<Vec<u64>>` (null on pre-16 traces) while the `switches.migrate_*` counters are plain zeros there; consumers that flatten the JSON need the one `Option`. Kept: "not measurable" and "measured zero" should not read the same for a per-CPU vector.

### Positive Notes
- The exact path keeps the existing invariants explicit and testable: runnable time is conserved between the legacy and exact shapes (legacy `runqueue.avg` + `unplaced_avg`/observed CPUs = exact `runqueue.avg` on the same synthetic trace, 0.45053 + 0.83435/3 = 0.72865), and `per_cpu.runnable_wait_ns` sums to the same total in both, redistributed by queue.
- Tests pin hand-computed answers for each new behaviour (placement at wakeup with the runqueue histogram and violation derived by hand; a preempted waiter re-queued with its wait split across two CPUs; no-migrate = previous CPU; new-task placement; unseen-thread/unseen-CPU handling) and the legacy shape's output is byte-identical apart from the additive fields.
- Kernel ordering is encoded rather than assumed: at one timestamp the stream handles slice end, then markers, then migrates, then slice start, matching `sched_waking` → `set_task_cpu` → enqueue → `sched_switch`.

Gates on the pushed tree: `cargo fmt --all -- --check` clean; `cargo build` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 505 passed / 0 failed.

PR #204, opened 2026-08-20
analyze: bound stacktrace flamegraph memory — --top N as a SQL LIMIT, streamed output, --max-depth D

josefbacik:
**What changes.** `systing-analyze stacktrace flamegraph` gets `--top N` (a `LIMIT` inside the fold query) and `--max-depth D` (the depth cut inside the `GROUP BY` key), and it streams folded lines to stdout as DuckDB produces them instead of building every unique stack in memory three times before printing the first line. The stderr footer keeps `# Unique stacks:` as the exact total (counted before the cut) and adds `# Matched samples:` (the sample total of everything that passed the filters), `# Output stacks:` (how many were emitted) and `# Max depth:` when set; `# Output samples:` stays the sum over the printed rows. The MCP `flamegraph` tool keeps its shape, gains `max_depth`, and its metadata gains `matched_samples`.

**Why.** Without a bound, the process's memory grows with the trace's unique-stack count with no ceiling: each unique stack was held as a `Vec<String>` of frames, as a folded line, and again in the joined folded text. On whole-system traces from large hosts that is gigabytes on top of what DuckDB uses for the fold, and a consumer that only wants the top few thousand stacks pays for all of them. 6e55afe and 12c99e3 bounded DuckDB's side of the fold (memory limit, spill); this bounds the Rust side. Measured on synthetic traces (table below): the Rust-side excess is about 1 GiB per 300K unique stacks and about 2.7 GiB per 1M; with `--top 4096` it is gone and the fold returns in about a second instead of tens of seconds.

**What to decide.** (1) Footer semantics: `# Unique stacks:` stays the exact total, and `# Matched samples:` / `# Output stacks:` are new, so under `--top` a consumer can keep trace-wide denominators (matched samples, unique stacks) while cross-checking the printed rows against `# Output samples:`. The alternative is to make `# Unique stacks:` the emitted count; I kept the total so a truncated output still says how much it left out. (2) `--max-depth` keeps the ROOT-most D frames (frames are stored root→leaf; the prefix is what an icicle view aggregates on). Say so if you want the leaf side instead. (3) The CLI default stays unbounded (no change for existing users); the structured/MCP path keeps its default of 500, now applied as a `LIMIT` as well.

<details>
<summary>Evidence — synthetic traces, release build, peak RSS = ru_maxrss of the child</summary>

Synthetic DBs (only the tables the fold reads): 100K distinct frames; stacks of depth 20–60 (mean 40) with 5 shared root frames drawn from 100 root shapes and pseudo-random deeper frames; samples half uniform over all stacks (the long tail) and half over the first 1,000 (the head). Times are wall clock from spawn to exit with stdout drained by the measurement script; "first line" is when the first folded line arrived.

| trace | binary | flags | wall | first line | peak RSS | stdout lines |
|---|---|---|---|---|---|---|
| 300K unique stacks / 3M samples | v1.13.5 | — | 10.2 s | 8.9 s | 1918 MiB | 297,906 |
| 300K / 3M | this PR | — | 8.2 s | 1.3 s | 992 MiB | 297,906 |
| 300K / 3M | this PR | `--top 4096` | 0.52 s | 0.38 s | 997 MiB | 4,096 |
| 1M unique stacks / 10M samples | v1.13.5 | — | 35.0 s | 30.7 s | 6198 MiB | 993,238 |
| 1M / 10M | this PR | — | 26.2 s | 3.8 s | 3971 MiB | 993,238 |
| 1M / 10M | this PR | `--top 4096` | 1.11 s | 0.70 s | 3806 MiB | 4,096 |
| 1M / 10M | this PR | `--top 4096 --max-depth 64` | 1.09 s | 0.72 s | 3929 MiB | 4,096 |

The "this PR" peak is DuckDB's own working set for the fold: the test host has no cgroup limit, so DuckDB ran with `memory_limit=51200MiB` and used what it liked; under a cgroup limit it stays inside the fraction 6e55afe gives it and spills. The Rust side's share is now flat in the number of unique stacks.

Equivalence checks on the 300K trace: the set of folded lines is identical between v1.13.5 and this PR (sorted diff empty); `--top 4096` equals the first 4,096 lines of the unbounded stream; counts are non-increasing down the output.
</details>

<details>
<summary>Design notes</summary>

- The fold is a `WITH folded AS MATERIALIZED (...)` CTE so it can be both counted (`(SELECT COUNT(*) FROM folded)` = unique stacks, `(SELECT SUM(count) FROM folded)` = matched samples) and cut (`ORDER BY ... LIMIT N` = a top-N heap) from one aggregation. `COUNT(*) OVER ()` gives the same number but measured 3.4× the query time on the 1M trace (2.7 s vs 0.8 s); a separate count query would re-run the aggregation. The two scalar subqueries over the materialized CTE cost nothing measurable.
- The id→string conversion (`array_to_string` over `frame_ids`) sits in the SELECT outside the limited subquery, so it runs only for surviving rows. In the same SELECT as the LIMIT it runs for every folded row (measured 2.4 s vs 0.8 s).
- `--max-depth D` is `list_slice(frame_ids, 1, D)` as the GROUP BY key, so stacks that only differ deeper than D merge in the database and a top-N over the merged stacks is exact. A read-side truncation after a top-N is not: the heaviest merged stack can be made of many light full-depth stacks (`fold_top_n_over_merged_stacks_is_exact` pins the case).
- The sort is `count DESC, trace_id, frame_ids` (a total order), so equal-count stacks come out in the same order every run and `--top` picks the same set. The old query ordered by count only.
- The CLI prints the request-describing footer lines before the stream and the count lines after it; consumers key on the `# ...:` prefixes. `# Matched samples:` / `# Unique stacks:` describe everything that passed the filters; `# Output samples:` / `# Output stacks:` describe what was printed; the pairs differ only under `--top`.
- Still loaded in full: the frame-name table (one entry per unique frame in the trace, not per unique stack). Resolving names inside DuckDB for the surviving rows would remove it too; left as a follow-up unless it shows up as the next growth term.
</details>

**Tests.** Eight fold tests on an in-memory trace DB (order and tie-break, top-N keeps both totals, max-depth merges in the database and makes top-N exact, min-count applies before the totals, empty window, structured result equals the stream, zero bounds rejected), three query-shape tests, and one CLI case in the analyze e2e suite (`--top 1` and `--max-depth 1` against a recorded trace, footer pairs checked). Local: fmt clean, clippy `-D warnings` clean, `cargo test --lib` 508/0.

**Hazard precedent (repo history):** d913a2f introduced the frame-id interning the fold reads and the Rust-side name resolution this keeps; 6e55afe / 12c99e3 bounded DuckDB's memory and spill for the analyze path. None of them bounded the per-stack materialization in Rust; this change does. No BPF, no schema, no recorder, no hot capture path — analysis-side only.

### DIFF
--- a/skills/systing-analyze/SKILL.md
+++ b/skills/systing-analyze/SKILL.md
@@ -180,12 +180,13 @@ Every table has a `trace_id` column. Always include it in joins, and filter on i
 
 ## Using `flamegraph`
 
-Returns `stacks` (array of `{frames, count}`), `metadata` (total_samples, unique_stacks, time_range, stack_type), and `folded` — folded-stack text (semicolon-separated frames root→leaf, space, count) for inferno / brendangregg's FlameGraph. Parameters:
+Returns `stacks` (array of `{frames, count}`), `metadata` (total_samples, matched_samples, unique_stacks, time_range, stack_type), and `folded` — folded-stack text (semicolon-separated frames root→leaf, space, count) for inferno / brendangregg's FlameGraph. `total_samples` is the whole trace; `matched_samples` and `unique_stacks` count what passed the filters BEFORE the `top_n` cut, so you can tell how much was cut. Parameters:
 - `stack_type` — `"cpu"` (default), `"interruptible-sleep"`, `"uninterruptible-sleep"`, `"all-sleep"`, `"all"`
 - `pid` / `tid` — restrict to a process or thread
 - `start_time` / `end_time` — seconds offset from trace start
 - `min_count` — minimum sample count per stack (default 1)
-- `top_n` — keep the top N stacks (default 500)
+- `top_n` — keep the top N stacks (default 500); the cut happens inside the database, so large traces stay cheap
+- `max_depth` — truncate every stack to its D root-most frames before merging (stacks that only differ deeper than D count as one); default full depth
 - `trace_id` — for multi-trace DBs
 
 ## Common investigation patterns
--- a/src/analyze/flamegraph.rs
+++ b/src/analyze/flamegraph.rs
@@ -64,7 +64,14 @@ pub struct FlamegraphParams {
     pub end_time: Option<f64>,
     pub trace_id: Option<String>,
     pub min_count: u64,
-    pub top_n: usize,
+    /// Keep only the `top_n` stacks with the most samples (`None` = every
+    /// stack). Applied as a `LIMIT` inside the fold query, so DuckDB keeps a
+    /// top-N heap and the caller never sees the long tail.
+    pub top_n: Option<usize>,
+    /// Cut every stack down to its `max_depth` frames nearest the root
+    /// (frames are stored root→leaf) before stacks are merged, so two stacks
+    /// that only differ below that depth count as one. `None` = full depth.
+    pub max_depth: Option<usize>,
 }
 
 impl Default for FlamegraphParams {
@@ -77,7 +84,8 @@ impl Default for FlamegraphParams {
             end_time: None,
             trace_id: None,
             min_count: 1,
-            top_n: 500,
+            top_n: Some(500),
+            max_depth: None,
         }
     }
 }
@@ -90,9 +98,16 @@ pub struct StackEntry {
 }
 
 /// Metadata about a flamegraph result.
+///
+/// `total_samples` counts every stack sample in the trace (or the selected
+/// trace); `matched_samples` and `unique_stacks` count what passed the
+/// filters (stack type, pid/tid, time window, `min_count`, and `max_depth`
+/// merging) BEFORE any `top_n` cut, so a truncated result can still say
+/// what it was cut from.
 #[derive(Debug, Serialize)]
 pub struct FlamegraphMetadata {
     pub total_samples: u64,
+    pub matched_samples: u64,
     pub unique_stacks: u64,
     pub time_range_seconds: (f64, f64),
     pub stack_type: String,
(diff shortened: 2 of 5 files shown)


josefbacik:
**rust-bpf-code-reviewer verdict** (in-session execution of the repo's reviewer agent definition, by the author, over the full diff at `303dd5513d815676918fffbb438af8dd15c535d0`; mechanics: the five review passes below were applied to the diff as a whole, not to a sample)

### Summary
A focused change to `systing-analyze stacktrace flamegraph`: the fold query gains an in-database `LIMIT` (`--top N`) and an in-database depth cut (`--max-depth D`), and the CLI streams folded lines instead of materializing every unique stack three times. The structure is simple (one streaming core, one collecting wrapper) and the SQL is explained where it is built. No BPF code, no schema, no recorder, no ELF parsing is touched, so passes 3 and 4 are empty by inspection.

### Critical Issues
None. `Cargo.toml`/`Cargo.lock` version untouched; no `SCHEMA_VERSION` change needed (no table or column changes); no `unsafe`.

### Improvements (all applied before commit)
- CLI validation: `--top 0` / `--max-depth 0` were only rejected inside the library with a library-flavoured message; they now fail at argument parsing (`clap` range `1..`) so the user sees the flag name, and the library keeps its own guard for MCP/library callers.
- `FlamegraphQuery` carried a `Clone` derive nothing used; dropped.
- The exact total was first written as `COUNT(*) OVER ()`; measured at 3.4× the query time on a 1M-stack synthetic, so it became a materialized CTE counted once (same number, no extra pass). The reasoning lives in the comment above the SQL.
- A consumer reading the footer under `--top` needs the matched-sample total as well as the unique-stack total, or its denominators shrink to the printed rows; `matched_samples` comes from the same CTE (`SUM(count)`, cast to BIGINT so duckdb-rs reads it as i64; `COALESCE` for the empty case) and is printed as `# Matched samples:`. No measurable cost.

### Suggestions
- The frame-name table is still loaded in full (one entry per unique frame in the trace). It is bounded by unique frames, not unique stacks, so it is not the growth term this change removes; resolving names inside DuckDB for the surviving rows only would remove it too, at the cost of a second query shape. Left as a follow-up, stated in the PR body.
- `FlamegraphMetadata` could carry the emitted count so the structured result did not need `stacks.len()`; not added, because the JSON shape the MCP tool documents would grow for no consumer.

### Positive Notes
- The three-layer SQL comment says why each layer exists (materialize once to count and cut; LIMIT inside so DuckDB keeps a heap; id→string conversion outside so it only runs on survivors) and the shape tests pin each of those facts.
- The sort's total order (`count DESC, trace_id, frame_ids`) makes `--top` deterministic across runs; the in-memory fold tests pin the tie-break with a concrete case.
- Streaming keeps exactly one folded line in flight and reuses its buffer; the footer moves after the stream and the comment states that consumers key on the `# ...:` prefixes.

josefbacik:
Rebased onto current `main` (v1.13.7, which carries the clippy fix from #205) so the Clippy job can pass; no content change — the rebased commit `9a3b597d3d2f295cad48da3ac3aabbcd86839e3d` has the same patch-id as the reviewed `303dd5513d815676918fffbb438af8dd15c535d0`, and the verdict above applies to it unchanged. Local gates on the rebased tree: fmt clean, clippy `-D warnings` clean, `cargo test --lib` 510/0.

PR #207, opened 2026-08-21
Start the CPU stack sampler after every other BPF link is attached

josefbacik:
The CPU stack sampler now attaches after every other BPF link, so a capture no longer records systing's own attach phase. Before this change the perf_event sampler was attached first, ahead of `skel.attach()`, and it sampled the tracepoint/raw_tp/fentry enables — which go through `text_poke_bp` and, on a many-core host, stall every CPU in IPI batches. Those samples look like real workload: kworkers and other tasks spinning on `text_mutex`, plus the IPI handlers themselves.

**What the reviewer decides:** that the sampler should start once the whole attach sequence is done (tracepoints, raw_tp/fentry, USDT/uprobe probes, memory-alloc uprobes), and still before a traced command is signalled to exec.

**Why it matters.** On a 224-CPU host in the middle of an IPI storm (job-start TLB flushes plus expedited RCU), the attach phase ran for about 9 s of a 10 s capture and read as about 6 cores of lock contention. Downstream analysis flagged that capture as node lock contention when nothing on the node was contending. Normally the attach phase costs about 0.01 core per capture and is invisible; under an IPI storm it balloons by two to three orders of magnitude, which is exactly when a false contention reading does the most damage.

**What changes.**
- `setup_perf_events` is split: `setup_perf_counter_events` keeps the `--perf-counter` events and the `perf_counters` map population where they were (before `skel.attach()`, so the map is populated before anything reads it); the new `attach_clock_sampler` opens the clock sampling events (hardware cpu-cycles, falling back to the software cpu-clock inside a VM — the same fallback as before) and attaches `systing_perf_event_clock` to them.
- `attach_clock_sampler` is called last in the attach sequence, after `skel.attach()`, the USDT/uprobe probes and the memory-alloc uprobes, and before the traced child is signalled to exec. `recorder.clock_sampling` is recorded right after it, as before; it is only read when the trace's sysinfo row is written.
- No BPF change, no schema change, no new flag.

**What does not change.** The capture window: `--duration` starts counting after the whole attach sequence either way, so a 10 s capture still samples for 10 s. The sampling event, period and VM fallback are the same code, moved. The sched/irq/network recorders attach exactly as before.

**Validation.** `cargo fmt -- --check` and `cargo clippy --all-targets` clean on the branch; the integration suite runs in CI (the attach sequence is exercised by every trace test). The in-repo `rust-bpf-code-reviewer` pass is posted as a comment below: no critical issues; its one improvement (name the counter-only function for what it does) is applied.

### DIFF
--- a/src/systing_core.rs
+++ b/src/systing_core.rs
@@ -2511,21 +2511,28 @@ fn cycles_sample_period(target_freq_hz: u64) -> u64 {
     (max_hz / target_freq_hz).max(1)
 }
 
-fn setup_perf_events(
+/// Open the per-CPU clock sampling events (hardware cpu-cycles, falling back
+/// to the software cpu-clock inside a VM) and attach `systing_perf_event_clock`
+/// to them.
+///
+/// This runs LAST in the attach sequence — after `skel.attach()`, the
+/// USDT/uprobe probes and the memory-alloc uprobes — so the stack sampler
+/// never observes the tracer's own attach phase. Enabling a tracepoint,
+/// raw_tp or fentry link goes through `text_poke_bp`, whose IPI batches stall
+/// every CPU on a many-core host; a sampler that is already live records that
+/// stall (kworkers spinning on `text_mutex`, the IPI handlers themselves) as
+/// if it were the workload. On a 224-CPU host mid IPI storm the attach phase
+/// ran ~9 s and read as ~6 cores of lock contention in a 10 s capture.
+/// The capture window is unchanged: `--duration` starts counting after the
+/// whole attach sequence either way.
+fn attach_clock_sampler(
     skel: &mut SystingSystemSkel,
     opts: &Config,
-    counters: &PerfCounters,
-    perf_counter_names: &[String],
     num_cpus: u32,
-) -> Result<(
-    Vec<libbpf_rs::Link>,
-    Vec<PerfOpenEvents>,
-    ClockSamplingConfig,
-)> {
+) -> Result<(Vec<libbpf_rs::Link>, ClockSamplingConfig)> {
     use crate::perf;
 
     let mut perf_links = Vec::new();
-    let mut event_files_vec = Vec::new();
 
     // Clamp to avoid divide-by-zero for library callers that bypass the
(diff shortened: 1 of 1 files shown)


josefbacik:
**rust-bpf-code-reviewer pass** — run in-session over the full diff (1 file, `src/systing_core.rs`) from the repo's agent definition (`rust-bpf-code-reviewer.md`), all five passes; mechanics disclosed: the author executed the reviewer definition over the diff before commit, fixed what it raised, and is posting the verdict here.

### Summary
A two-function split plus a reorder: the clock sampler's open+attach moves out of `setup_perf_events` into `attach_clock_sampler`, which now runs after every other link; the counter-event setup keeps its place. Small, readable, does one thing; no BPF, schema or version change.

### Critical Issues
None. No schema touched (no `SCHEMA_CHANGES.md` / `SCHEMA_VERSION` / `[bump minor]` needed); `Cargo.toml` version untouched; no `unsafe`; no ELF parsing; no recorder or Python-version change; no BPF map or program change (map/context pairing unchanged — the perf_event program still writes the same maps it did).

### Improvements
- (applied) `setup_perf_events` after the split only handled `--perf-counter` events; renamed to `setup_perf_counter_events` so the name says what it does. Its one call site updated.

### Suggestions
- (applied) The `attach_clock_sampler` doc comment said the reorder "costs nothing"; replaced with the fact it stands on — the capture window is unchanged because `--duration` starts after the whole attach sequence either way.
- The moved block is byte-identical code (VM fallback, `eprintln!`, attach loop); no idiom changes were made to it on purpose, so the diff stays a move plus a call-site reorder.

### Positive Notes
- The ordering invariant is stated at both ends: in the `attach_clock_sampler` doc (why last) and at the call site (what is now in place when it runs); the run-command path keeps `child.signal_exec()` after the sampler attach, so the exec is still captured.
- `recorder.clock_sampling.set(...)` moved with the value it records and keeps its existing comment; the value is only read at sysinfo-row write time, so nothing observes the later set.
- The perf counter map is still populated before `skel.attach()`, so the context-switch readers never see an empty map.

PR #209, opened 2026-08-21
Match --cgroup targets by ancestry, not a start-time snapshot of leaf ids

josefbacik:
## What changes

`--cgroup <path>` traces a task when its cgroup is the target **or any cgroup below it**, cgroups created after the trace started included — the kernel decides (`task_under_cgroup_hierarchy()`, the predicate behind `bpf_task_under_cgroup()`), instead of userspace enumerating the target's descendants at start or BPF walking a task's cgroup ancestors by hand.

Three patchsets: the first matched by walking `cgrp->ancestors[]` with probe reads and comparing ids against a hash map; the review asked for the kfunc instead, and the second dropped the walk, the CO-RE flavor for pre-6.1 cgroups, the depth bound and the lockdown special case for the kernel's own test — and refused to start on a kernel without the kfunc. The third (the last commit) keeps the pre-existing behaviour as the **fallback** on such kernels instead of refusing: the start-time snapshot, with its known loss of cgroups created after the trace started.

## Why

A filtered trace of a pod whose container runtime keeps creating cgroups under the pod (a privileged container running a nested OCI runtime, with `runc:[2:INIT]` and short-lived helpers in its process mix) came back with **zero rows in every table** three times out of three over two weeks, while systing exited Ok after its full duration and an unfiltered trace on the same host attributed the same pod normally. The `--cgroup` path was correct (a wrong path is a hard `Failed to access cgroup path` error); the filter matched nothing because the tasks lived in leaves that did not exist when the start-time id snapshot was taken.

## How

Two kernel entry points into the same predicate are used, because no single one is callable from every program that shares `trace_task()`:

**Current task, every program type — `bpf_current_task_under_cgroup(map, idx)`.** The helper (since 4.8) tests `current` against a slot of a `BPF_MAP_TYPE_CGROUP_ARRAY`: a `READ_ONCE` of the slot and an `ancestors[]` compare, lock-free and NMI-safe. Userspace opens each `--cgroup` directory and puts the fd into slot `i` of the new `cgroup_targets` map (the kernel holds the cgroup reference for the map's lifetime). `trace_task()` loops over the targets with it; all of its callers outside the sched tracepoints pass `current`, including the `perf_event` sampler, the kprobe/uprobe/usdt programs, the raw tracepoints and the classic tracepoints.

**Any task, `tp_btf` programs — `bpf_task_under_cgroup(task, cgrp)`.** The sched tracepoints test tasks that are not current (the switched-in task, a wakee, a migrating task, a forked child's parent), so they call the kfunc through the new `trace_task_btf()`. The kfunc needs a trusted `struct cgroup *`; those are the references the new `SEC("iter/cgroup") systing_cgroup_target_add` program parks in slot `i` of the new `cgroup_target_refs` map (`struct cgroup __kptr *` plus the cgroup id). Userspace runs that program once per target before attaching anything: a cgroup iterator on the target's directory fd with `BPF_CGROUP_ITER_SELF_ONLY`, driven by one `read()` (process context), which takes the reference with `bpf_cgroup_from_id()` and records the id; userspace reads the slot back and refuses to start if the id is not the target's. `task_in_cgroup_filter()` loads the kptr under `bpf_rcu_read_lock()` — `struct cgroup` is one of the verifier's RCU-protected kptr types, so the load is accepted as trusted for this `KF_RCU` kfunc — and holds no lock and takes no reference per event.

**Why not the kfunc everywhere** (read at the tags in the kernel tree): at v6.6 the generic kfunc set that holds `bpf_task_under_cgroup` is registered for TRACING, SCHED_CLS and STRUCT_OPS programs only (`kernel/bpf/helpers.c` `kfunc_init()`, `kernel/bpf/btf.c` `bpf_prog_type_to_kfunc_hook()`); v6.12 adds `tracepoint` and `perf_event` to the tracing hook but gives `kprobe` its own hook without the generic set and `raw_tracepoint` none. systing's `trace_task()` is shared by all of those. **Why not `bpf_cgroup_from_id()` per event:** at v6.6 `cgroup_get_from_id()` → `kernfs_find_and_get_node_by_id()` takes a plain `spin_lock(&kernfs_idr_lock)` (`fs/kernfs/dir.c`), not callable from `sched_wakeup` (hardirq context) or from the NMI sampler; at v6.12 that path is RCU, but the program-type limits above still hold. Resolving the pointers once at start in process context sidesteps both. **Why `bpf_cgroup_from_id()` rather than `bpf_cgroup_acquire(ctx->cgroup)` in the iterator:** at v6.6 the cgroup iterator's `ctx->cgroup` is `PTR_TO_BTF_ID_OR_NULL` without `PTR_TRUSTED` (it gained that in v6.7), so the acquire would not verify there; by-id works on both.

**Which kernels take which path.** The mode is decided once at start from the capability, not the version: if the running kernel's vmlinux BTF exports `bpf_task_under_cgroup` (`kernel_has_task_under_cgroup_kfunc()`, the same probe pattern as `kernel_has_rss_stat_btf()`; the kfunc is v6.5+ — `bpf_rcu_read_lock` 6.2, the cgroup iterator 6.1, referenced kptrs 6.0), `--cgroup` runs in **kernel mode** as described above. Otherwise it runs in **legacy mode**: the mechanism this PR started from — userspace enumerates each target and every descendant cgroup that exists at start (`collect_descendant_cgroup_ids()`) into the `cgroups` hash map, sized to that count, and every program matches the task's own cgroup id against the set exactly; cgroups created under a target after the trace started are not traced. That is the old behaviour, loss included, rather than no `--cgroup` at all. A rodata flag (`tool_config.cgroup_match_kernel`) routes `trace_task()` / `trace_task_btf()` to one path or the other; rodata is a verifier constant, so the branch not taken is dead code at load, the kfunc call behind `bpf_ksym_exists()` is never reached on a kernel without it, and the iterator program is only autoloaded in kernel mode. The mode in use is printed once at start. `SYSTING_CGROUP_FILTER_LEGACY=1` forces legacy mode on any kernel, so the fallback can be exercised — and is tested, below — where no pre-6.5 kernel is at hand. Not taken: a mixed mode (the helper for `current` on every kernel, the snapshot only for the non-current tasks in the sched tracepoints) would lose less on old kernels, but a task would then match in the sampler and not in the sched events on the same host; the fallback is the old behaviour whole.

**Cost.** Paid only when `--cgroup` is set, per target (normally one): one helper call (a pointer read and an array compare) in the non-`tp_btf` programs, one map lookup plus the kfunc's `ancestors[]` compare in the sched tracepoints. No probe reads, so nothing changes under lockdown confidentiality mode.

**Map/context pairing.** All three filter maps are written only from process context — `cgroup_targets` by userspace, `cgroup_target_refs` by the iterator program run by userspace before attach, the legacy `cgroups` set by userspace at start — and only read by the tracing programs; no BPF program that runs in NMI or IRQ context writes any of them, nothing is LRU, and the references are dropped with the map on exit. The maps of the mode not in use stay at one entry.

## Tests

- `src/cgroup.rs`: `cgroup_id()` returns the target's own inode only, errors on a missing path and on a non-directory; `collect_descendant_cgroup_ids()` (legacy mode) keeps its four tests — target first, descendants deduplicated, regular files and symlinks skipped, missing path / non-directory are errors.
- `src/systing_core.rs`: the pure mode decision (`cgroup_match_mode`: kernel mode iff the kfunc exists and nothing forces the fallback) and the program selection (the iterator program is required only in kernel mode).
- `tests/trace_validation.rs::test_e2e_cgroup_filter_follows_cgroups_created_after_start` (`#[ignore]`, root + BPF; unchanged apart from wording): creates `<own cgroup>/systing-cgf-<pid>/` as the `--cgroup` target and a sibling `…-outside/leaf/` with a busy task in it, starts a 10 s trace, and **3 s in** creates `target/mid/late/` and moves a second busy task there. It asserts the late task has a process row whose `cgroup_id` is the new leaf's inode, that it has stack samples or sched slices, and that the outside task has no row.
- `tests/trace_validation.rs::test_e2e_cgroup_filter_legacy_snapshot_fallback` (`#[ignore]`, root + BPF): the same fixture under `SYSTING_CGROUP_FILTER_LEGACY=1` (set for the test's duration by a drop guard), plus a leaf that exists under the target before the trace starts with a busy task in it. It asserts the inverse pair: the early task is traced in its own leaf with stack samples or sched slices (the snapshot covers the subtree as it was), the late task has **no** row (the legacy mode's known loss, and the proof the fallback was in effect — kernel mode would have traced it), and the outside task has no row.

## Verification

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 521 + 11 + 28 passed, 0 failed (the four `collect_descendant_cgroup_ids` tests are back; the mode-decision and program-selection tests are new).
- `clang -target bpf` compile of `systing_system.bpf.c` with the build's include set: clean apart from the pre-existing `read_tcp_seq_from_skb` qualifier warning.
- VM run of both e2e tests (root + BPF) on a virtme-ng guest, kernel `6.12.0 #1 SMP PREEMPT_DYNAMIC` (BTF-enabled build; 4 vCPU / 8 GiB, software-emulated CPU): **`2 passed; 0 failed; 39 filtered out`** — `test_e2e_cgroup_filter_follows_cgroups_created_after_start` (kernel mode, as in the second patchset) and `test_e2e_cgroup_filter_legacy_snapshot_fallback` (the knob set): `early workload pid 303: traced, cgroup_id 102 (leaf), 10042 stack samples + sched slices` / `late workload pid 306: not traced (the legacy mode's known loss)` / `outside workload pid 304: not traced (correct)`. A first run of this patchset failed the legacy test on its discriminator: the snapshot was taken after the skeleton was opened, ~30 s into set-up on the emulated guest, so the cgroup the test creates 3 s in was already inside it; the snapshot now happens first thing in `systing()` (see How), and the rerun is the pass above. To run them yourself: `sudo ./scripts/run-integration-tests.sh trace_validation test_e2e_cgroup_filter`.
- Not run: a 6.6 kernel. On one, `--cgroup` takes legacy mode through the BTF probe; the forced-knob run above is the same code path with the probe's answer overridden.
- Rebased onto v1.13.13 (the merges of #211, #210 and #208) after the branch went conflicting: patchsets 1 and 2 re-applied unchanged; patchset 3's one conflict was both #210 and this branch appending tests to the same module tail (both kept), plus an `#[allow(clippy::too_many_arguments)]` on `configure_bpf_skeleton`, which now takes #210's ring plan beside this branch's `--cgroup` mode. Gates and both e2e tests re-run on the rebased bytes: `2 passed; 0 failed` on the same guest (`early workload pid 297: traced … 10027 stack samples + sched slices` / `late workload pid 300: not traced` / `outside workload pid 298: not traced`).

No schema, recorder, `Cargo.toml` version or dependency change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,6 +253,11 @@ the process going off the CPU or going on the CPU in addition to the process
 being traced, so you will miss events for the unwanted process leaving the CPU.
 Perfetto handles this appropriately, but it looks odd.
 
+`--cgroup <path>` traces every task whose cgroup is that cgroup or any cgroup
+below it, including cgroups created after the trace started (membership is
+decided by the kernel's `bpf_task_under_cgroup()`); it requires **Linux kernel
+6.5 or newer** and fails with a clear error on older kernels.
+
 `--trace-event` - This will add an `instant` track event for each event that
 this tool captures.  The format is "<trace type>:<optional
 info>:<class>:<name>".  This is most easily obtained by running
--- a/README.md
+++ b/README.md
@@ -254,9 +254,13 @@ being traced, so you will miss events for the unwanted process leaving the CPU.
 Perfetto handles this appropriately, but it looks odd.
 
 `--cgroup <path>` traces every task whose cgroup is that cgroup or any cgroup
-below it, including cgroups created after the trace started (membership is
-decided by the kernel's `bpf_task_under_cgroup()`); it requires **Linux kernel
-6.5 or newer** and fails with a clear error on older kernels.
+below it, including cgroups created after the trace started — membership is
+decided by the kernel's `bpf_task_under_cgroup()`, which needs **Linux kernel
+6.5 or newer**. On older kernels (no `bpf_task_under_cgroup` in the kernel's
+BTF) systing falls back to matching a snapshot of the target's cgroups taken
+when the trace starts, so cgroups created under the target afterwards are not
+traced; the mode in use is printed at start. Setting
+`SYSTING_CGROUP_FILTER_LEGACY=1` forces the fallback on any kernel.
 
 `--trace-event` - This will add an `instant` track event for each event that
 this tool captures.  The format is "<trace type>:<optional
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -103,6 +103,12 @@ const volatile struct {
 				 * cgroup_targets / cgroup_target_refs maps;
 				 * rodata so the filter loops are
 				 * verifier-bounded. */
+	u32 cgroup_match_kernel; /* 1: the kernel decides --cgroup membership
+				  * (cgroup_targets / cgroup_target_refs);
+				  * 0: legacy exact match of the task's own
+				  * cgroup id against the cgroups map (see the
+				  * map comments). Rodata, so the branch not
+				  * taken is dead code to the verifier. */
 	u32 no_stack_traces;
 	u32 no_cpu_stack_traces;
 	u32 no_sleep_stack_traces;
@@ -642,6 +648,23 @@ struct {
 	__type(value, struct cgroup_target_ref);
 	__uint(max_entries, 1);
 } cgroup_target_refs SEC(".maps");
+
+/*
+ * Legacy --cgroup matching, used when the running kernel's BTF does not
+ * export bpf_task_under_cgroup (before 6.5) or when SYSTING_CGROUP_FILTER_LEGACY
+ * forces it: userspace loads the cgroup ids of each target and of every
+ * descendant that exists when the trace starts, sized to that count, and the
+ * tracing programs match the task's own cgroup id against the set exactly.
+ * Cgroups created under a target after the trace started are not in the set
+ * and their tasks are not traced - the limitation the kernel-side matching
+ * above removes. Written by userspace only, lookup-only from BPF.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, u64);
+	__type(value, u8);
+	__uint(max_entries, 1);
+} cgroups SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
@@ -1716,6 +1739,18 @@ static bool should_filter_systing(struct task_struct *task)
 	return false;
 }
 
+/*
+ * Legacy --cgroup matching (cgroup_match_kernel == 0): the task's own cgroup
+ * id looked up in the start-time set userspace loaded into the cgroups map.
+ * Any program type, any kernel; misses cgroups created after the trace
+ * started (see the cgroups map comment).
+ */
+static bool task_in_legacy_cgroup_set(struct task_struct *task)
+{
+	u64 cgid = task_cg_id(task);
+	return bpf_map_lookup_elem(&cgroups, &cgid) != NULL;
+}
+
 /* Every filter except --cgroup; see trace_task() and trace_task_btf(). */
 static bool trace_task_common(struct task_struct *task)
 {
@@ -1743,8 +1778,11 @@ static bool trace_task(struct task_struct *task)
 {
 	if (!trace_task_common(task))
 		return false;
-	if (tool_config.filter_cgroup && !current_in_cgroup_filter())
-		return false;
+	if (tool_config.filter_cgroup) {
+		if (tool_config.cgroup_match_kernel)
+			return current_in_cgroup_filter();
+		return task_in_legacy_cgroup_set(task);
+	}
 	return true;
 }
 
@@ -1760,8 +1798,11 @@ static bool trace_task_btf(struct task_struct *task)
 {
 	if (!trace_task_common(task))
 		return false;
-	if (tool_config.filter_cgroup && !task_in_cgroup_filter(task))
-		return false;
+	if (tool_config.filter_cgroup) {
+		if (tool_config.cgroup_match_kernel)
+			return task_in_cgroup_filter(task);
+		return task_in_legacy_cgroup_set(task);
+	}
 	return true;
 }
 
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -580,6 +580,16 @@ struct {
 	__uint(max_entries, 10240);
 } perf_counters SEC(".maps");
 
+/*
+ * The --cgroup filter set: cgroup v2 ids (dfl_cgrp->kn->id, the directory
+ * inode) of each --cgroup target. Written by userspace before tracing starts,
+ * read-only from BPF (lookups only, from both the perf_event sampler and the
+ * tp_btf/fentry programs, so it must stay a plain preallocated HASH — never
+ * LRU). A task matches when its own cgroup OR any ancestor is in the set (see
+ * task_in_cgroup_filter()); userspace only enumerates descendants into this
+ * map under kernel lockdown, where the ancestor walk's probe reads are
+ * unavailable. Sized by userspace to the ids it loads.
+ */
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__type(key, u64);
@@ -1454,6 +1464,98 @@ static u64 task_cg_id(struct task_struct *task)
 }
 
 /*
+ * CO-RE flavor type for kernels < 6.1, where struct cgroup carried its
+ * ancestors' cgroup ids directly (u64 ancestor_ids[], indexed by level) instead
+ * of the struct cgroup *ancestors[] pointer array that replaced it in 6.1. The
+ * ___pre61 suffix is ignored by CO-RE type matching, so this matches the
+ * kernel's actual struct cgroup at runtime.
+ */
+struct cgroup___pre61 {
+	u64 ancestor_ids[0];
+};
+
+/*
+ * Deepest cgroup v2 nesting level the --cgroup ancestry walk inspects. Real
+ * hierarchies are shallow (a Kubernetes container leaf sits at level 4-6; a
+ * nested runtime inside it adds a few more); the bound keeps the walk a
+ * verifier-bounded loop. A task deeper than this is still matched on its own
+ * cgroup id and on its first CGROUP_FILTER_MAX_DEPTH ancestors.
+ */
+#define CGROUP_FILTER_MAX_DEPTH 16
+
+/*
+ * --cgroup filter: is the task's own cgroup, or any ancestor of it, one of the
+ * --cgroup targets?
+ *
+ * Matching by ancestry rather than by exact id keeps the filter correct for
+ * cgroups created AFTER the trace started - a nested container runtime making
+ * cgroups inside a pod, a systemd unit spawning transient scopes. The task's
+ * leaf cgroup may be brand new, but its ancestor chain still reaches the
+ * target. Userspace therefore loads only the targets' own ids into the
+ * `cgroups` map, not a start-time snapshot of their descendants (which by
+ * construction could never contain a cgroup that did not exist yet).
+ *
+ * The walk probe-reads cgrp->ancestors[i] (cgrp->ancestor_ids[i] before 6.1)
+ * from a CO-RE-relocated array base: ancestors[] is a flexible array member,
+ * and the verifier does not accept a direct BTF load of its pointer elements
+ * past the struct's declared size, so each element is read with
+ * bpf_probe_read_kernel at an explicit offset. Probe reads are restricted under
+ * kernel lockdown confidentiality mode; there the walk is skipped and
+ * userspace enumerates each target's descendants at start instead (the
+ * pre-ancestry behaviour, with its known start-time-snapshot limitation).
+ *
+ * Cost, paid only when --cgroup is set: one hash lookup, plus at most
+ * level x (up to 3 probe reads + 1 hash lookup) when the task is not directly
+ * in a target. Lookups only - this map is never written from BPF, so it is
+ * safe from both the perf_event (NMI) sampler and the tp_btf/fentry programs.
+ */
+static bool task_in_cgroup_filter(struct task_struct *task)
+{
+	struct cgroup *cgrp = task->cgroups->dfl_cgrp;
+	u64 cgid = cgrp->kn->id;
+
+	/* The task sits directly in a target (or, under lockdown, in one of the
+	 * pre-enumerated descendants). */
+	if (bpf_map_lookup_elem(&cgroups, &cgid))
+		return true;
+	if (tool_config.confidentiality_mode)
+		return false;
+
+	/* ancestors[level] is the cgroup itself, checked above; walk the proper
+	 * ancestors, root first. */
+	int level = cgrp->level;
+	if (level > CGROUP_FILTER_MAX_DEPTH)
+		level = CGROUP_FILTER_MAX_DEPTH;
+
+	if (bpf_core_field_exists(cgrp->ancestors)) {
+		/* >= 6.1: struct cgroup *ancestors[], one pointer per level. */
+		struct cgroup **ancestors =
+			(void *)cgrp + bpf_core_field_offset(cgrp->ancestors);
+		for (int i = 0; i < CGROUP_FILTER_MAX_DEPTH && i < level; i++) {
+			struct cgroup *anc;
+			if (bpf_probe_read_kernel(&anc, sizeof(anc), &ancestors[i]))
+				break;
+			u64 id = BPF_CORE_READ(anc, kn, id);
+			if (id && bpf_map_lookup_elem(&cgroups, &id))
+				return true;
+		}
+	} else {
+		/* < 6.1: u64 ancestor_ids[], the cgroup id itself per level. */
+		struct cgroup___pre61 *old = (void *)cgrp;
+		u64 *ancestor_ids =
+			(void *)cgrp + bpf_core_field_offset(old->ancestor_ids);
+		for (int i = 0; i < CGROUP_FILTER_MAX_DEPTH && i < level; i++) {
+			u64 id;
+			if (bpf_probe_read_kernel(&id, sizeof(id), &ancestor_ids[i]))
+				break;
+			if (id && bpf_map_lookup_elem(&cgroups, &id))
+				return true;
+		}
+	}
+	return false;
+}
+
+/*
  * CO-RE flavor type for kernels < 6.18 that have icsk_timeout as a separate field.
  * The ___pre618 suffix is ignored by CO-RE type matching, so this matches the
  * kernel's actual struct inet_connection_sock at runtime.
@@ -1562,11 +1664,8 @@ static bool trace_task(struct task_struct *task)
 		if (bpf_map_lookup_elem(&pids, &pid) == NULL)
 			return false;
 	}
-	if (tool_config.filter_cgroup) {
-		u64 cgid = task_cg_id(task);
-		if (bpf_map_lookup_elem(&cgroups, &cgid) == NULL)
-			return false;
-	}
+	if (tool_config.filter_cgroup && !task_in_cgroup_filter(task))
+		return false;
 	return true;
 }
 
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -99,6 +99,10 @@ _Static_assert(sizeof(struct systing_stack_build_id) ==
 const volatile struct {
 	u32 filter_pid;
 	u32 filter_cgroup;
+	u32 num_cgroup_targets; /* --cgroup arguments = filled slots of the
+				 * cgroup_targets / cgroup_target_refs maps;
+				 * rodata so the filter loops are
+				 * verifier-bounded. */
 	u32 no_stack_traces;
 	u32 no_cpu_stack_traces;
 	u32 no_sleep_stack_traces;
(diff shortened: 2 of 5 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/fc12-cgroup-ancestry`

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the full diff (4 files, +491/−30) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
A targeted correctness fix: the `--cgroup` filter moves from "exact match against a start-time snapshot of leaf ids" to "own id or any ancestor id", which is the semantics the flag always implied. The BPF walk is bounded, lock-free, lockdown-aware and CO-RE-portable; userspace shrinks to one id per target except under lockdown. The e2e test reproduces the exact failure (a leaf created after trace start, two levels below the target) with a negative control. No critical issues.

### Critical Issues
None. No schema change (`src/duckdb.rs`, `src/trace/schema.rs` untouched; `SCHEMA_VERSION` unchanged), no recorder change, no `Cargo.toml` version or dependency change, no ELF parsing, no Python version work.

### Improvements
1. **BPF — `task_in_cgroup_filter()` re-derives what `task_cg_id()` computes.** The exact-lookup key is `cgrp->kn->id` computed inline because the walk also needs `cgrp`. Acceptable (one deref saved vs. calling both), but if a third reader of `dfl_cgrp` appears, factor a `task_cgroup(task)` accessor and make `task_cg_id()` use it. Not changed now (YAGNI).
2. **Userspace — `confidentiality_mode` is read back from rodata rather than re-detected.** Correct (avoids a second `/sys/kernel/security/lockdown` read and a duplicate stderr message) and it is the value the BPF side will actually run with; the `unwrap_or(false)` arm is unreachable in practice because the rodata `expect` earlier in the same function already fired if rodata is not mmap'ed. Fine as written; noted so the fallback is understood as belt-and-braces, not a silent mode switch.
3. **Test — `CgroupWorkload::spawn_in` panics inside the helper thread on failure.** The join re-raises the original payload via `resume_unwind`, so the message survives; the channel hand-off of created dirs means cleanup still runs. Slightly more machinery than the other e2e tests carry, justified by the cgroup dirs needing removal on every exit path.

### Suggestions
- `CGROUP_FILTER_MAX_DEPTH = 16` is documented as "deeper than any real hierarchy"; the clamp `level = min(level, 16)` means a task at depth > 16 is matched on its own id and its first 16 ancestors (root-first), which is the right degradation (the shallow targets are the ones people pass).
- The map-definition comment says "never LRU"; upstream readers without the team's history may read it as a non sequitur — kept, because the sentence before it states the reason (lookups from NMI context).
- Comments carry no source-line references (checked); they explain why (verifier flexible-array rule, lockdown, kernel-version flavor), which the definition allows under "non-obvious verifier requirements or kernel version considerations".

### Positive Notes
- The exact-lookup-first order keeps the common case (a task directly in a leaf target, and every lockdown case) at exactly the old cost.
- `cgroup_id()` keeps the "typo'd --cgroup is a loud error" property that `collect_descendant_cgroup_ids()` had, with a unit test pinning `InvalidInput` on a non-directory.
- The e2e test's `cgroup_id == leaf inode` assertion proves the match came from an ancestor, not merely that something was traced.

- `cargo fmt --all -- --check`: pass.
- `cargo clippy --all-targets -- -D warnings`: pass.
- `cargo test --lib --bins`: 519 + 11 + 28 passed, 0 failed (the three new `cgroup_id` tests included).
- `clang -target bpf` dry compile of `systing_system.bpf.c`: clean except the pre-existing `read_tcp_seq_from_skb` const-qualifier warning (not in this diff).
- Map/context pairing check (grep): `LRU_HASH` under `src/bpf/` → 0 hits; the `cgroups` map is written only by userspace (`skel.maps.cgroups.update`) and only looked up from BPF — unchanged by this diff.

josefbacik:
Just use the bpf_task_under_cgroup kfunc.

josefbacik:
Verifier-load evidence for the merge gate named in the PR body: the new end-to-end test ran on a real kernel at this PR's head.

- Head: `432064efb63d93f5ac7e8bd1e55c87c8a08b141d` (this PR's head at the time of the run)
- Host: `scripts/run-integration-tests.sh`-equivalent VM rig — virtme-ng guest on kernel `6.12.0 #1 SMP PREEMPT_DYNAMIC` (BTF-enabled build, from the bzImage header; software-emulated CPU, 4 vCPU / 8 GiB); the BPF object loaded and attached, so the verifier accepted the ancestry-matching change
- Target: `trace_validation`, filter `test_e2e_cgroup_filter_follows_cgroups_created_after_start --nocapture`

Test output (guest):

```
Recording trace (10s, --cgroup /sys/fs/cgroup/systing-cgf-254, late cgroup at +3s)...
Initialized streaming parquet output to /tmp/.tmphkL9yD
Detected virtualized environment, automatically switching to software events (--sw-event)
Generating Parquet trace...
Flushing scheduler trace records from streaming...
Flushed scheduler trace records in 0.55s
Writing clock snapshot...
Writing process and thread data...
Writing network interface metadata...
Wrote metadata records in 0.11s
Flushing memory trace records...
Flushed memory trace records in 0.01s
Flushing stack samples and symbolizing stacks...
Note: symbolization start: anon RSS 8 MiB (289 unique stacks)
Note: symbolization exited-process pass done: anon RSS 8 MiB (289 unique stacks)
Note: symbolization done: anon RSS 17 MiB (289 unique stacks)
Flushed and symbolized stack samples in 4.13s
Flushing network trace records...
Flushed network trace records in 0.00s
Flushing perf counter records from streaming...
Flushed perf counter records in 0.06s
Flushing sysinfo records from streaming...
stop-phase: symbolize and write parquet done in 5149ms
Successfully wrote Parquet trace files
Converting to Perfetto format: /tmp/.tmphkL9yD/trace.pb...
Successfully wrote /tmp/.tmphkL9yD/trace.pb in 2.08s
ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 39 filtered out; finished in 61.15s

  late workload pid 258: traced, cgroup_id 74 (leaf), 10065 stack samples + sched slices
  outside workload pid 256: not traced (correct)
```

Exit code 0. The workload whose cgroup was created 3 s after the trace started is traced (10,065 stack samples plus scheduler slices attributed to it), and the workload outside the filtered subtree is not — the behaviour the change exists to produce. Not exercised here: the live cgroup layout on the production TPU pool named in the PR body (that check would confirm the field case, it is not required for the fix's correctness).

josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/fc12-cgroup-ancestry`, second patchset (`bpf_task_under_cgroup`)

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the full diff against `main` (5 files, +734/−140; the delta against the first patchset is +413/−280) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
The `--cgroup` filter now asks the kernel whether a task is in or below a target — `bpf_current_task_under_cgroup()` for `current` from every program type, `bpf_task_under_cgroup()` for non-current tasks from the `tp_btf` sched programs — with the target references taken once at start by a cgroup-iterator program. The hand-rolled ancestor walk, its pre-6.1 CO-RE flavor, the depth bound, the lockdown branch and the userspace descendant enumeration are gone. The e2e test is unchanged in what it asserts. No critical issues; one correctness property (the kfunc never reaches a program type that cannot call it) rests on a call-site split that the review checked by hand and the 6.12 guest exercises at load.

### Critical Issues
None. No schema change (`src/duckdb.rs`, `src/trace/schema.rs` untouched; `SCHEMA_VERSION` unchanged), no recorder change, no `Cargo.toml` version or dependency change, no ELF parsing, no Python version work. `libbpf_sys` is reached only through the `libbpf_rs::libbpf_sys` re-export the crate already depends on.

### Improvements
1. **BPF — the `trace_task()` / `trace_task_btf()` split is enforced by convention, not by the compiler.** A future `trace_task_btf()` call from a kprobe/tracepoint/perf_event/raw_tracepoint program would fail at load ("calling kernel function bpf_task_under_cgroup is not allowed"), loudly and on every kernel, so the failure mode is a refused load rather than a wrong filter; the function comments name the rule. Checked by hand: the seven `trace_task_btf()` sites are the `handle_sched_*` bodies (wakeup_new, switch, waking, migrate, process_exit, process_fork) and `systing_sched_process_exec`, each reached only from a `SEC("tp_btf/…")` wrapper; every other `trace_task()` caller passes `bpf_get_current_task_btf()`.
2. **BPF — `task_in_cgroup_filter()` returns `false` without the kfunc.** Dead on every kernel userspace lets `--cgroup` start on (the BTF probe refuses below 6.5), kept so the `tp_btf` programs load on older kernels for traces without `--cgroup`. Not a fallback and documented as such.
3. **Userspace — the iterator handoff is verified per slot.** `install_cgroup_targets()` reads `cgroup_target_refs[i].id` back after each run and bails on a mismatch, so a silently-empty filter (the bug class this PR exists for) cannot come back through an iterator that ran but stored nothing. The kptr itself reads as zero from userspace by design; the `id` is the witness.
4. **Userspace — `fs::File::open()` on the cgroup directory** is the fd the `CGROUP_ARRAY` update needs (the kernel takes its own reference); the `File` lives in the `CgroupTarget` vector for the run, which is longer than needed but harmless.

### Suggestions
- `MAX_CGROUP_TARGETS = 64` is mirrored in Rust (`const MAX_CGROUP_TARGETS: usize`) and enforced with a clear error; the two constants are not tied at build time. A `static_assert`-style check is not available across the C/Rust boundary here; the name match and the comments will do.
- The iterator program fills "the first free slot" rather than an index passed in; userspace runs the iterators in order, so slot `i` gets target `i`, and the id readback pins it. If a future caller ever ran them concurrently the readback would catch the mismatch.
- `bpf_cgroup_from_id()` inside the iterator rather than `bpf_cgroup_acquire(ctx->cgroup)`: chosen for 6.6, where the iterator's cgroup argument is `PTR_TO_BTF_ID_OR_NULL` without `PTR_TRUSTED` (the trusted flag arrived in 6.7), so the acquire would not verify there. The comment says so.

### Positive Notes
- The `perf_event` sampler's path is a single helper call that reads one pointer and compares `ancestors[]` — strictly cheaper than the first patchset's hash lookup plus probe-read walk, and nothing it calls takes a lock.
- Both new maps are written only from process context (userspace and the userspace-driven iterator) and read by every tracing program — the pairing the NMI sampler needs, with no `LRU_HASH` anywhere under `src/bpf/` (grep: 0 hits).
- The refusal on kernels without the kfunc is one probe against vmlinux BTF, the same shape as `kernel_has_rss_stat_btf()`, and the message names the floor.

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 515 + 11 + 28 passed, 0 failed (the four removed descendant-enumeration tests account for the drop from 519; the three `cgroup_id` tests still run).
- `clang -target bpf` dry compile of `systing_system.bpf.c`: clean except the pre-existing `read_tcp_seq_from_skb` const-qualifier warning (not in this diff); the object carries the `iter/cgroup` section and `struct cgroup_target_ref` with the `kptr` type tag on `cgrp`.
- Map/context pairing check: `LRU_HASH` under `src/bpf/` → 0 hits; `cgroup_targets` is written only by userspace (`skel.maps.cgroup_targets.update`) and `cgroup_target_refs` only by `systing_cgroup_target_add`, which userspace runs before attach; both are lookup-only from every tracing program.
- VM e2e run on the 6.12 guest: PASS — `test_e2e_cgroup_filter_follows_cgroups_created_after_start` on the virtme-ng 6.12.0 guest (4 vCPU / 8 GiB): `1 passed; 0 failed; 39 filtered out`, `late workload pid 255: traced, cgroup_id 74 (leaf), 10001 stack samples + sched slices`, `outside workload pid 253: not traced (correct)` (the run on the committed bytes); the verifier loaded every program, the iterator included. The first attempt failed to load the iterator program ("infinite loop detected": the counter's address passed to the lookup spilled it untracked); fixed with a per-iteration `u32 key = i` in both loops before this verdict.

josefbacik:
> Just use the bpf_task_under_cgroup kfunc.

Done in f5524d8: the kfunc from the `tp_btf` sched programs, the `bpf_current_task_under_cgroup()` helper (CGROUP_ARRAY of the target fds) for `current` everywhere else — the only form the kprobe/raw_tracepoint programs may call on 6.6 and 6.12, and the perf_event sampler on 6.6 — and a cgroup-iterator program that takes the target references once at start. Kernel floor 6.5+ for `--cgroup`. Details in the updated description.

josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/fc12-cgroup-ancestry`, third patchset (the legacy fallback)

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the third commit's diff (5 files, +581/−69 against the second patchset; the cumulative PR diff against `main` is the same five files) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
The second patchset refused `--cgroup` on kernels without `bpf_task_under_cgroup`; this one keeps the pre-existing behaviour there instead — the start-time snapshot of each target's cgroups matched exactly, with its known loss of cgroups created after the trace started. One rodata flag selects the mode; both paths keep their maps process-context-written and tracing-read. The fallback is exercised on the 6.12 guest through an environment knob with a second e2e test whose discriminator is the known loss itself. No critical issues.

### Critical Issues
None. No schema change (`src/duckdb.rs`, `src/trace/schema.rs` untouched; `SCHEMA_VERSION` unchanged), no recorder change, no `Cargo.toml` version or dependency change, no ELF parsing, no Python version work. The BPF object adds one hash map (`cgroups`, the pre-#209 map restored at one entry) and one rodata byte.

### Improvements
1. **BPF — the mode switch is a rodata constant, so the unused branch is dead code at load.** `trace_task()` and `trace_task_btf()` test `tool_config.cgroup_match_kernel` before either matcher; on a kernel without the kfunc the legacy branch is the only live one and the `bpf_ksym_exists()` guard on the kfunc call is never reached. Checked: `systing_cgroup_target_add` is put in the required set only in kernel mode (`get_required_bpf_programs`, unit-tested), so no iterator program is loaded where it could not verify.
2. **Userspace — the snapshot is the first thing `systing()` does.** `resolve_cgroup_filter()` moved from after the skeleton open to the top of `systing()`: in legacy mode the snapshot is what "at start" means, and everything after it (symbolizer, output sink, skeleton open and load — tens of seconds on an emulated guest) would otherwise widen the window in which a cgroup created "after start" still makes the set. The first rig run of this patchset failed exactly there (the late cgroup, created 3 s in, was inside a snapshot taken ~30 s in) and passed after the move.
3. **Userspace — the mode is printed once at start, with its reason.** "membership decided by the kernel" vs "this kernel's BTF does not export bpf_task_under_cgroup" / "`SYSTING_CGROUP_FILTER_LEGACY` is set", so a trace's log says which semantics it ran with — the operator-facing half of "the known loss".
4. **Test — `LegacyCgroupFilterEnv` clears the knob on drop, panic path included.** The two e2e tests share a process; the integration runner is `--test-threads=1`, and the guard means the kernel-mode test can never inherit the forced fallback if the legacy test fails midway.

### Suggestions
- The mixed mode (the helper for `current` everywhere, the snapshot only for the sched tracepoints' non-current tasks) would lose less on old kernels; it was rejected because a task would then match in the sampler and not in the sched events on the same host. Documented in the body; no code.
- `cgroup_match_mode()` is a two-boolean function kept pure for its unit test; if a third input ever appears (a per-recorder need), it is the one place to extend.
- The README sentence now names both modes and the knob; USAGE docs elsewhere do not mention `--cgroup` and were left alone.

### Positive Notes
- The legacy matcher is the pre-#209 code verbatim (`task_cg_id()` → exact lookup), and `collect_descendant_cgroup_ids()` returns with its four tests — the fallback is the old behaviour whole, not a third design.
- Map/context pairing holds for all three filter maps: `cgroup_targets` (userspace), `cgroup_target_refs` (the iterator program run by userspace before attach), `cgroups` (userspace at start) — written from process context only, lookup-only in every tracing program, no `LRU_HASH` under `src/bpf/` (grep: 0 hits).
- The legacy e2e test proves both halves with one fixture: a leaf present at start IS traced (the mode works), a leaf created mid-trace is NOT (the loss is real and the mode was in effect).

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 521 + 11 + 28 passed, 0 failed (the four restored `collect_descendant_cgroup_ids` tests and the two new mode/program-selection tests by name).
- `clang -target bpf` dry compile of `systing_system.bpf.c`: clean except the pre-existing `read_tcp_seq_from_skb` const-qualifier warning (not in this diff).
- Map/context pairing check: `LRU_HASH` under `src/bpf/` → 0 hits; the three filter maps as above.
- VM e2e run on the 6.12 guest: PASS — both tests on the virtme-ng 6.12.0 guest (4 vCPU / 8 GiB): `2 passed; 0 failed; 39 filtered out`; legacy run: `early workload pid 303: traced, cgroup_id 102 (leaf), 10042 stack samples + sched slices`, `late workload pid 306: not traced (the legacy mode's known loss)`, `outside workload pid 304: not traced (correct)`; kernel-mode run passed on its own assertions (late leaf traced in its own cgroup, sibling excluded). The first attempt failed the legacy test on its discriminator because the snapshot was taken after the skeleton open; moving it to the top of `systing()` (Improvement 2) is the fix, and the rerun above is on the committed bytes.

josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/fc12-cgroup-ancestry`, rebased onto v1.13.13

Mechanics, disclosed: the branch's three commits were rebased onto `main` at `a28ac40a` (v1.13.13 = the merges of #211, #210 and #208) because the PR had gone CONFLICTING. The repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the conflict-resolution delta — `git range-diff 0c257af5..620453a3 a28ac40aed..HEAD` — and over the combined tree's BPF map/program set (the third patchset's verdict stands for the patchset's own content). Author and reviewer are the same session.

### What the rebase changed
- Patchsets 1 and 2 re-applied as byte-identical patches (`range-diff`: `=`).
- Patchset 3 re-applied with context-only differences, plus one real conflict: #210 and this patchset both appended tests at the tail of `systing_core.rs`'s test module. Resolution: both sets kept — #210's four `ring_plan_*` tests, then `test_cgroup_match_mode_prefers_kernel_unless_forced` and `test_cgroup_filter_iterator_program_only_in_kernel_mode`. No line of either side was dropped or edited; a duplicated `#[test]` attribute from the splice was caught on read-back and removed before the commit.
- Commit messages: unchanged (read back over all three: one `Co-authored-by` each, no injected trailer).

### Combined-tree check
- `configure_bpf_skeleton()` now takes both #210's `ring_plan: RingbufPlan` and this branch's `cgroup_kernel_mode: bool`; `systing()` resolves the `--cgroup` filter first, then plans the rings — independent inputs, no ordering dependency.
- Map/context pairing on the combined set: #210's inner rings are created by userspace after load and inserted into the outers before attach (`create_ring_shards`); this branch's `cgroup_targets` / `cgroup_target_refs` / `cgroups` are filled by userspace (and the userspace-driven iterator program) after load and before attach (`install_cgroup_targets`, the legacy id loop). Every filter-map and ring-map writer runs in process context; the tracing programs only look up. `LRU_HASH` under `src/bpf/` → 0 hits.
- #208's `systing_page_fault_sw` (a `perf_event` program) calls `trace_task()` with `bpf_get_current_task_btf()` — the helper path, which that program type may call; the `trace_task()` / `trace_task_btf()` split from the second patchset is unchanged.

### Critical Issues
None.

- On the rebased bytes (HEAD `e4c0034`): `cargo fmt --all -- --check` pass; `cargo clippy --all-targets -- -D warnings` pass (after the `#[allow]` above); `cargo test --lib --bins` 529 + 11 + 28 passed, 0 failed (529 = the previous 521 plus #210's eight `ring_plan_*` tests); clang dry compile of the combined BPF object clean except the pre-existing warning; VM rig on the 6.12 guest: `2 passed; 0 failed; 39 filtered out` — `early workload pid 297: traced, cgroup_id 102 (leaf), 10027 stack samples + sched slices`, `late workload pid 300: not traced (the legacy mode's known loss)`, `outside workload pid 298: not traced (correct)`; the kernel-mode test passed on its own assertions.

PR #210, opened 2026-08-22
Shard ring buffers per CPU at load time instead of a fixed eight per family

josefbacik:
Ring buffers are now sharded per CPU at load time — `min(num_cpus, 64)` rings per family instead of a compile-time eight — with the family's byte budget held constant, so a 192-CPU host runs 64 × 4 MiB rings where it ran 8 × 32 MiB. Why: with eight rings, 24 CPUs share one ring and one consumer there, and `__bpf_ringbuf_reserve` takes the ring's spinlock before it checks for space, so a full ring is also a lock its producers contend on — and a node-wide 60 s sched+irq capture on that class dropped ~85 % of its events (the rings-full data loss is the case for this change). The decisions for review: the constant-budget sizing rule (per-CPU share unchanged, single-CPU burst absorption smaller, eight times the consumers), the 64-ring ceiling, and `--ringbuf-shards N` as the only knob (`8` restores the old layout exactly).

<details>
<summary>Mechanism</summary>

- **BPF**: the eight `ringbuf_<family>_node{0..7}` static inners and their `.values` initializers are gone; each outer `ARRAY_OF_MAPS` keeps only its inner-ring *template* (`RINGBUF_TEMPLATE_BYTES`, 64 KiB — libbpf creates one transient map from it for the outer's `inner_map_fd`; the kernel's inner-map compatibility check for ringbufs, `bpf_map_meta_equal`, compares type/key/value/flags, not `max_entries`) and `NR_RINGBUFS_MAX = 64` slots. The eight reserve sites shard by the rodata `tool_config.nr_ringbufs`.
- **Userspace**: `plan_ringbufs()` picks count and size; `configure_bpf_skeleton` sizes each outer to `nr_rings` and sets the rodata; after load, `create_ring_shards()` creates the inner rings (`MapHandle::create`) and inserts them into each outer with **one** `BPF_MAP_UPDATE_BATCH` per family. The batch matters: the kernel runs an RCU grace period after map-of-maps updates (`maybe_wait_bpf_programs`), once per `BPF_MAP_UPDATE_ELEM` but only once per batch call, so 64 per-slot updates per family would have been seconds of startup on a busy many-core host (the cost the existing `autocreate(false)` comment measured at ~100 ms per grace period). Kernels without batch support for this map type fall back to per-slot updates. Families no enabled recorder writes stay `autocreate=false`. The handles own the ring fds for the life of the trace (libbpf's `ring_buffer` registers the fd with epoll without duplicating it) and are released explicitly after the poller threads are joined.
- **Consumers**: one poller thread per ring as before; the sched family keeps one SPSC channel and one `sched_rec_{i}` consumer per ring, so consumer parallelism scales with the ring count. `SessionRecorder::event_recorders` is sized to the ceiling (idle shards are empty collections).

</details>

<details>
<summary>Ring size at a constant family budget</summary>

Up to eight rings each ring is `--ringbuf-size-mib` rounded up to a power of two (what libbpf did to the static rings: 25 → 32 MiB), so hosts with at most eight CPUs get exactly the footprint they had. Above eight the family keeps the budget eight such rings implied and splits it, rounded **down** to a power of two, never below 1 MiB:

| host | `--ringbuf-size-mib` | before | after |
|---|---|---|---|
| 4 CPUs | 25 | 4 × 32 MiB reachable (+4 page-sized) | 4 × 32 MiB |
| 48 CPUs | 25 | 8 × 32 MiB = 256 MiB | 48 × 4 MiB = 192 MiB |
| 192 CPUs | 25 | 8 × 32 MiB = 256 MiB | 64 × 4 MiB = 256 MiB |
| 192 CPUs | default (50) | 8 × 64 MiB = 512 MiB | 64 × 8 MiB = 512 MiB |
| 384 CPUs | 25 | 8 × 32 MiB (48 CPUs/ring) | 64 × 4 MiB (6 CPUs/ring) |

The per-CPU share is unchanged; what a single CPU's burst can borrow from its neighbours shrinks (a 4 MiB ring shared by 3 CPUs instead of a 32 MiB ring shared by 24), while the number of consumers draining the family rises eight-fold. The wakeup threshold follows the per-ring size (half a ring).

</details>

<details>
<summary>Flagged choices</summary>

- **Count gated by host size, not by recorder set.** A node-wide sched+irq capture on a 192-CPU host now records up to ~7× more events than its previously lossy capture; that is the point, but the parquet write at stop and the held table grow with it on that host class. Deployments requesting node-wide sched on such hosts should size the request (duration, recorder set, scope) for a capture that does not drop — or pass `--ringbuf-shards 8` until they do.
- **Ceiling 64.** Above 64 CPUs several CPUs share a ring again (6 at 384). The ceiling bounds thread count (one poller per ring per family) and map count; raising it is a one-line change if a host class needs it.
- **Constant budget rather than constant per-ring size.** Holding the per-ring size would have multiplied a family's memory by `num_cpus / 8` (3.2 GiB per family at 64 × 50 MiB), on exactly the hosts where the tracer's memory ceiling already bites.

</details>

<details>
<summary>Evidence for the problem</summary>

192-vCPU hosts, node-wide 60 s `--only-recorder sched --only-recorder irq` captures: ~85 % of scheduler events dropped (`Missed sched/IRQ events: 345.7 M` and `339.4 M` on two captures, i.e. hundreds of millions of events lost per minute; 15 of 56 on-demand captures on that host class in one week ended with their rings full). Ring-buffer reserve spin appears in the host's lock profile during those on-demand windows on this host class; which programs' producers are spinning has not been read at artifact grade, so this change makes no claim about it — an earlier sentence attributing that spin to the tracer's own producers (also in the commit message) is withdrawn, and the case rests on the overflow. Kernel facts, read at v6.12: `kernel/bpf/ringbuf.c` `__bpf_ringbuf_reserve` — `raw_spin_lock_irqsave` precedes the `new_prod_pos - cons_pos > mask` space check; `kernel/bpf/map_in_map.c:83-92` `bpf_map_meta_equal` compares type, key/value size, flags and the BTF record, not `max_entries`; `kernel/bpf/arraymap.c:1417` `array_of_maps_map_ops.map_update_batch = generic_map_update_batch`; `kernel/bpf/syscall.c:143-155` `maybe_wait_bpf_programs` → `synchronize_rcu()` for map-of-maps, called per `BPF_MAP_UPDATE_ELEM` (`:1628`, `:1682`) and once after a `BPF_MAP_UPDATE_BATCH` (`:5167`).

</details>

<details>
<summary>Verification</summary>

- `cargo fmt`, `cargo clippy --all-targets -D warnings`, unit tests (`ring_plan_*`: small hosts keep the old footprint; 192/384 CPUs split a constant budget; non-power-of-two counts never exceed it; override and ceiling).
- e2e (`trace_validation`: the validation suite — sched, stacks, exits, migrations; cpu-stacks-only, i.e. the sched family left uncreated; tracepoint instants, the probe family) on a v6.12 guest at 4 vCPU (4 rings × 64 MiB, the ≤ 8 path) and at 16 vCPU (16 rings × 32 MiB, the > 8 path); every missed-event counter 0 — output in the first comment.
- Not exercised: a 192-CPU host (64 rings) end to end, and the batch-update fallback on a kernel without `BPF_MAP_UPDATE_BATCH` for `ARRAY_OF_MAPS` (pre-5.6). The first production read on the 192-vCPU class is the `Missed sched/IRQ events` / `Missed stack events` counters before and after.

</details>

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -65,7 +65,17 @@
  */
 #define SKIP_STACK_DEPTH 3
 #define PERF_EVENT_SKIP_STACK_DEPTH 0
-#define NR_RINGBUFS 8
+/*
+ * Upper bound on rings per family. The live count is
+ * tool_config.nr_ringbufs (min(num_cpus, NR_RINGBUFS_MAX), or the
+ * --ringbuf-shards override), set by userspace before load; the outer
+ * ARRAY_OF_MAPS below are sized to it and populated from userspace after
+ * load. A compile-time 8 was the ceiling before: on a 192-CPU host that
+ * put 24 CPUs behind one ring, and __bpf_ringbuf_reserve takes the ring's
+ * spinlock before it checks for space, so a full ring is also a lock every
+ * producer on those CPUs spins on.
+ */
+#define NR_RINGBUFS_MAX 64
 
 /* Mirror of the kernel's struct bpf_stack_build_id (filled by bpf_get_stack
  * under BPF_F_USER_BUILD_ID), with the offset/ip union flattened to one u64
@@ -134,6 +144,13 @@ const volatile struct {
 				* gate is belt-and-braces (and lets the verifier
 				* prune the bodies if the programs ever get loaded
 				* for another reason). */
+	u32 nr_ringbufs;       /* Rings per family: events land in ring
+				* `cpu % nr_ringbufs`. Userspace sizes each outer
+				* ARRAY_OF_MAPS to this before load and creates +
+				* inserts the inner rings after load (one batch
+				* update per family). Rodata, so the modulo is by
+				* a constant the verifier can see; 0 is never
+				* loaded (userspace pins it >= 1). */
 } tool_config = {};
 
 enum event_type {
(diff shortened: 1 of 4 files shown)


josefbacik:
End-to-end runs of this change on a real kernel (virtme-ng guests on a `6.12.0` BTF kernel, software-emulated CPUs), `trace_validation` tests `test_e2e_only_cpu_stacks_emits_samples`, `test_e2e_tracepoint_instant`, `test_e2e_validation_suite` (the rig runs the `#[ignore]`d e2e tests, `--ignored --test-threads=1`). The runs were on the tree first committed as `8f2afb6`; the PR head `79f50a1` differs from it only in the poller thread names and one test comment (no BPF or data-path change; fmt, clippy and the unit tests were re-run on the head).

Guest 1 — 4 vCPU / 8 GiB (the ≤ 8-ring path, one ring per CPU at the legacy size):

```
Ring buffers: 4 rings per family x 64 MiB (4 possible CPUs)
test test_e2e_tracepoint_instant ... Tracing for 3 seconds
test test_e2e_validation_suite ... Tracing for 4 seconds
Missed sched/IRQ events: 0
Missed stack events: 0
Missed probe events: 0
Missed cache events: 0
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 36 filtered out; finished in 367.17s
```

Guest 2 — 16 vCPU / 8 GiB (the > 8-ring path: 16 rings created from userspace, batch-inserted, each polled; the default 50 MiB → 64 MiB legacy ring split eight ways across sixteen rings = 32 MiB each, 512 MiB per family as before):

```
Ring buffers: 16 rings per family x 32 MiB (16 possible CPUs)
test test_e2e_tracepoint_instant ... Tracing for 3 seconds
test test_e2e_validation_suite ... Tracing for 4 seconds
Missed sched/IRQ events: 0
Missed stack events: 0
Missed probe events: 0
Missed cache events: 0
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 36 filtered out; finished in 429.41s
    found 426 exit states (14 EXIT_DEAD, 412 EXIT_ZOMBIE)
    found 14 sched_migrate rows over 16 CPUs, 0 orphans
```

The cpu-stacks-only test is the case where the sched family is not created at all (`autocreate=false` on its outer map, no inner rings, no consumers); the validation suite records sched, stacks, process exits and migrations across every ring and checks the parquet, DuckDB and Perfetto views against each other; the tracepoint test exercises the probe family. Exit code 0 on both guests. Not exercised here: a host with more than 64 CPUs (the ceiling) and the per-slot fallback for kernels without batch updates on `ARRAY_OF_MAPS`.

josefbacik:
Review pass (the repository's `rust-bpf-code-reviewer` definition, executed by the author over the full diff; the PR head `79f50a1` includes the one fix the pass produced) — mechanics disclosed: structure → Rust → BPF → schema/versioning → integration, on `git show HEAD` for all four files, with the kernel claims checked against the v6.12 source cited in the PR body.

### Summary
The change replaces a compile-time eight-ring layout with a load-time plan and userspace-created inner rings, inserted in one batch per family. The Rust side is a pure planning function with tests, one creation function, and a table-driven poller setup; the BPF side shrinks (the 64 static inner maps and their initializers go away) and gains one rodata field. No schema, versioning, recorder-identity, Python-version or ELF-parsing surface is touched.

### Critical Issues
None found.
- Schema / `SCHEMA_VERSION` / `SCHEMA_CHANGES.md`: not touched (no table or DuckDB change).
- `Cargo.toml` / `Cargo.lock` version: not touched.
- Recorder identity (`utid`): no recorder or table added.
- NMI safety: the rings are written by exactly the programs that wrote them before, through the same `bpf_map_lookup_elem(&<family>, &idx)` + `bpf_ringbuf_reserve` path; `bpf_ringbuf_reserve` is NMI-safe and no new map is written from NMI context. The only BPF-side behavioural change is the divisor of the shard index (`tool_config.nr_ringbufs`, rodata, constant to the verifier) and the outer maps' slot count.
- `unsafe`: none added; the one pre-existing `sysconf` block keeps its `// SAFETY:` comment.

### Improvements
1. **Applied.** Poller thread names for the non-sched families were `"<outer>_<i>"` (`stack_ringbufs_17`), which the kernel truncates to 15 bytes — every poller of a family would read `stack_ringbufs_` in `ps` and in systing's own captures. They now use the family's short inner-ring prefix plus the index (`rb_stk_17`), the pattern the sched family already used (`events_<i>` / `sched_rec_<i>`). The prefix rides in `RingShards` beside the handles.
2. Not applied, noted: `plan_ringbufs` converts the per-ring byte count with `u32::try_from(..).expect(..)`; a `--ringbuf-size-mib` of 4096 or more would panic with that message rather than a usage error. The previous code overflowed the same multiplication silently in release builds, so this is no worse; a bound on the flag is a separate, small follow-up if anyone wants it.

### Suggestions
- The poller setup matches on the outer map's name as a string; the family table is the single source and the `other =>` arm bails, so an enum would be tidier but is not needed.
- `create_ring_shards` reports the per-slot fallback once per run with the kernel's error text; keeping that note on stderr (rather than silent) is deliberate — it is the signal that a host is on a pre-batch kernel.

### Positive Notes
- One `BPF_MAP_UPDATE_BATCH` per family with a per-slot fallback, and the comment says exactly why (the grace period per map-of-maps update) — the cost model that the old `autocreate(false)` comment measured is carried forward rather than rediscovered.
- The plan is a pure function with tests that pin the two invariants reviewers will ask about (small hosts byte-identical; non-power-of-two counts never exceed the budget).
- Handle lifetime is explicit: `RingShards` lives in `ThreadHandles` and is dropped after the poller join, with the reason (epoll registers the fd without duplicating it) stated once, where it matters.
- The BPF object got smaller and simpler: eight template types, no per-slot declarations, one rodata field.

josefbacik:
Body edited: the spin-attribution sentence is withdrawn; the case rests on the ring overflow

PR #211, opened 2026-08-24
pystacks: discover the runtime from the exe and libpython only, reading lazily

josefbacik:
## Why

Python process discovery read every candidate binary in full (`read_to_end`) and, when the executable was not the runtime (the `--enable-shared` launcher case: pyenv, conda, most distro builds), went on to read every mapping whose path contained "python", in address order, until one carried `_PyRuntime`. Extension modules dlopen'ed after start-up map below libpython and all live under a python-named path, so on an ML image the scan read libtpu, jaxlib and torch in full before reaching libpython: gigabytes per run, and cold on every run, because `ELF_CACHE` is process-local and the continuous lane spawns a fresh process per capture. On hosts whose page cache is already starved by the workload, that read stalled the host for tens of seconds at every capture start — the job's own page faults competing with ours.

## What

1. `runtime_candidates`: the executable, then each distinct `libpython*` shared object in mapping order — nothing else is parsed. A CPython runtime lives in the executable or in a libpython object; `site-packages/*.so` never carries `_PyRuntime`. The cheap `looks_like_python` gate is unchanged, so embedders (uwsgi/gunicorn) are still found through their libpython mapping.
2. `parse_elf_py_info` reads lazily through `object`'s `ReadCache` instead of slurping the file: the ELF headers, the symbol tables and — via `data_range` — the 8-byte pointer and at most 32 bytes of string behind `_PySys_ImplCacheTag`, never a whole section. The `(dev, ino, size)` cache and its I/O-failure rule (a transient error is never cached as a rejection) are unchanged; an explicit readability probe replaces the implicit one `read_to_end` gave.

Flagless. `ElfPyInfo` (runtime address, version, `is_dynamic`) is identical for every accepted binary, so the frame walker downstream sees no change.

## Measured

`test_discovery_read_volume_real_python` (per-thread `rchar` around `check_python_process` on a live interpreter with stdlib extension modules loaded, cold cache): discovery asks for **5.2 MB** where the previous order read **40.5 MB across 20 files** (CPython 3.13, pyenv shared build). On an ML image the previous figure was the extension modules below libpython; the new one is bounded by libpython's tables whatever else is mapped. Warm-cache parse of a libpython: 0.04–0.07 s, the same order as before (symbol names are now read in 256-byte chunks instead of from one in-memory buffer).

## Trade-off to know

Discovery no longer parses a mapped object just because its path contains "python". A runtime embedded in an object that is neither the executable nor `libpython*`-named would no longer be discovered; I know of no such layout (pyenv, conda, Debian/RHEL, bazel hermetic toolchains and the static case all fit), and a report of one is a one-line addition to `runtime_candidates`.

## Tests

- `test_runtime_candidates_exe_then_libpython_only` — an ML-image-shaped map list: extension modules and lib-dynload objects excluded, exe first, one entry per libpython.
- `test_runtime_candidates_embedder_without_python_exe` — uwsgi-style embedder; case-insensitive, debug/versioned spellings.
- `test_elf_py_info_real_interpreters` — every pyenv interpreter present on the host (eight shared builds, 3.8–3.14): libpython recognised through the lazy reader with the version from the cache tag, launcher exe classified.
- `test_discovery_read_volume_real_python` — the live-process discovery and the byte bound above (asserts under a quarter of the previous order's bytes). Both real-interpreter tests skip with a note where no pyenv is installed.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and the full `cargo test` are green on the branch.

### DIFF
--- a/src/pystacks/discovery.rs
+++ b/src/pystacks/discovery.rs
@@ -5,11 +5,11 @@
 use super::offsets;
 use super::process::{self, MemoryMapping};
 use super::types::{BpfLibBinaryId, PyPidData, BPF_LIB_DEFAULT_FIELD_OFFSET};
+use object::read::{ReadCache, ReadRef};
 use object::{Object, ObjectSymbol};
 use std::collections::HashMap;
 use std::fs;
-use std::io::Read;
-use std::os::unix::fs::MetadataExt;
+use std::os::unix::fs::{FileExt, MetadataExt};
 use std::path::Path;
 use std::sync::{Arc, LazyLock, Mutex};
 use std::time::Instant;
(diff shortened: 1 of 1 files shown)


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of f79fa58; verdict as it produced it.

### Summary
Replaces the full-file read and the generic "python"-pathed scan in pystacks discovery with an exe-then-libpython candidate list and a lazy `object::ReadCache` parse. Uses the `object` crate throughout (no hand-rolled ELF), keeps `ElfPyInfo` semantics identical, and backs the change with candidate-order tests, real-interpreter parses and a measured read-volume bound.

### Critical Issues
None. No schema, version, recorder or Python-version-support surface is touched; `Cargo.toml` is unchanged.

### Improvements
- `read_impl_cache_tag` reads the cache-tag pointer as 8 little-endian bytes (`u64::from_le_bytes`), as the previous code did. Both supported targets (x86-64, aarch64 Linux) are LE 64-bit; a one-line comment naming the assumption would help the next reader. Not changed in this PR.

### Suggestions
- `elf_py_info` probes readability both before and after the parse; the pre-parse probe only saves a parse attempt on an unreadable path (EISDIR). Fine as is; it could collapse to the post-parse probe alone.
- `test_discovery_read_volume_real_python` kills its interpreter before asserting, but a panic in the map or exe reads ahead of that point would leave the child to its 120 s sleep. A drop guard would close that; acceptable for a unit test.

### Positive Notes
- The candidate list is a pure function over the map list, so the exclusion that matters (extension modules mapped below libpython) is unit-tested directly.
- `data_range` for the cache-tag bytes avoids pulling whole `.data`/`.rodata` sections through the lazy reader.
- The per-thread `rchar` instrument turns the I/O claim into a regression test rather than a number in the PR body.

PR #212, opened 2026-08-24
Keep eight rings on kernels that wait per insert, and raise the fd limit

josefbacik:
## What changes

Two follow-ups to the per-CPU ring-buffer sharding (#210), both about the hosts it will meet.

1. **Kernels before 6.8 keep eight rings by default.** `plan_ringbufs()` now takes the running kernel's `major.minor` and caps the *automatic* shard count at the legacy eight below 6.8; an explicit `--ringbuf-shards N` is honoured on any kernel. An unreadable version counts as old.
2. **The file-descriptor soft limit is raised at start.** systing raises `RLIMIT_NOFILE`'s soft limit to the hard limit, prints both and a projection of what the run needs, and warns when the projection exceeds the hard limit.

Plus: `--ringbuf-size-mib` ≥ 4096 is a clear error rather than a panic; `--ringbuf-shards` is documented in `docs/USAGE.adoc` and the skill's flag table; the `create_ring_shards` comment that placed map-of-maps batch support "pre-5.6" now says 5.19.

## Why

**The shard cap.** The sharding inserts a family's inner rings with one `BPF_MAP_UPDATE_BATCH`, on the premise that the kernel waits one RCU grace period per batch. That premise holds from **6.8**: `37ba5b59d6ad` ("bpf: Call maybe_wait_bpf_programs() only once for generic_map_update_batch()") moved the wait out of the per-element path. Before it — v6.6 `kernel/bpf/syscall.c`, `bpf_map_update_value()` ending in `maybe_wait_bpf_programs()` → `synchronize_rcu()` for `ARRAY_OF_MAPS` / `HASH_OF_MAPS`, called per element by `generic_map_update_batch()`; unchanged through 6.6.144, and not backported — every inserted ring costs its own grace period. 64 rings × the families a run creates = that many grace periods before the first event, on a busy many-core host tens of milliseconds each, and paid again on every capture start. Eight rings on those kernels is the old layout exactly (`8` restores it byte for byte), so nothing is lost there that was not lost before; the per-CPU layout applies where the batch is cheap.

**The fd limit.** Each inner ring holds three descriptors (its map handle, the consumer's `ring_buffer` epoll, the poller thread's epoll); each CPU holds the clock sampler's perf event and BPF link, plus the software page-fault event and link per CPU when the memory recorder runs on a non-x86 host. 64 rings × 8 families × 3 = 1,536 before the per-CPU descriptors, so a sharded layout on a large host exceeds the 1024 soft limit a unit typically inherits, and systing never raised it — Rust's standard library does not either. A process may always raise its soft limit up to its hard limit without privilege.

## How

- `parse_kernel_release()` reads `major.minor` from the release string (tolerating `-custom.2` / `+` suffixes); `kernel_batches_map_of_maps_updates()` compares against `MAP_OF_MAPS_BATCH_WAIT_ONCE_KERNEL = (6, 8)`; `plan_ringbufs()` takes the result as a fourth argument and chooses `num_cpus.clamp(1, 8)` instead of `num_cpus.clamp(1, 64)` when it is false and no override is given. The startup line reads `Ring buffers: N rings per family x M MiB (C possible CPUs; kernel X.Y; one per CPU | legacy count, kernel inserts rings one grace period at a time | --ringbuf-shards)`.
- `raise_fd_soft_limit()` (`getrlimit`/`setrlimit` through `libc`, already a dependency): raises soft → hard when they differ, reports a failed raise as a warning, never fails the run. `projected_fd_need()` = `rings × families × 3 + CPUs × (2, or 4 with the memory recorder off x86) + 256`, printed as `File descriptors: soft limit raised 1024 -> 1048576 (hard 1048576); this run needs about 1984`, with a warning when the projection exceeds the hard limit.
- `SYSTING_DIAG_RINGBUFS` now also prints `[diag] create_ring_shards (N rings x F families) took …`, the number to read when a capture's set-up looks slow.

No BPF change; no schema, recorder, `Cargo.toml` version or dependency change.

## Tests

- `ring_plan_keeps_the_legacy_count_where_inserts_wait_per_ring`: 192 CPUs at 6.6 / 6.7 / 5.19 / unreadable → 8 × 32 MiB; small hosts unaffected; the override honoured at 6.6; 6.8 / 6.12 / 6.18 / 7.1 → 64.
- `kernel_release_parses_real_release_strings`: `6.12.0`, `6.6.97-generic`, `6.18.44-custom.2`, `7.1.7`, `6.12`, and the rejects (`6`, `garbage`, empty); the comparison at the boundary.
- `fd_projection_counts_rings_cpus_and_a_fixed_allowance`: the arithmetic on both sides of the memory-recorder rule; the old layout on a small host under 1024.
- The existing `ring_plan_*` tests pass a batching kernel and are otherwise unchanged.

## Verification

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 527 + 11 + 28 passed, 0 failed — the three new tests and the four existing `ring_plan_*` tests by name.
- VM run of `test_e2e_validation_suite` (the trace-validation e2e suite, root + BPF) on a virtme-ng guest, kernel `6.12.0` (BTF-enabled build; 4 vCPU / 8 GiB, software-emulated CPU): **`1 passed; 0 failed; 38 filtered out`**. The guest's startup lines from this branch: `Ring buffers: 4 rings per family x 64 MiB (4 possible CPUs; kernel 6.12; one per CPU)` and `File descriptors: soft limit raised 1024 -> 4096 (hard 4096); this run needs about 288` — the raise and the projection are live, and a 6.12 kernel takes the per-CPU rule as before. To run it yourself: `sudo ./scripts/run-integration-tests.sh trace_validation test_e2e_validation_suite`.
- The kernel facts behind the cap were read at the tags in a linux tree: v6.6 `kernel/bpf/syscall.c` — `maybe_wait_bpf_programs()` at :142 (`synchronize_rcu()` for `ARRAY_OF_MAPS` / `HASH_OF_MAPS`), called at the end of `bpf_map_update_value()` (:205), which `generic_map_update_batch()` calls per element; unchanged at v6.6.144 (:144 / :207); `37ba5b59d6ad` is in v6.8-rc1 and in no 6.6.y tag; `9263dddc7b6f` ("bpf: Extend batch operations for map-in-map bpf-maps") is v5.19.

## Not run

A 6.6 kernel: the per-element wait is read from the source at the tags named above, not timed; the cap's effect on such a host is exactly the pre-sharding layout, so the risk of the cap being wrong is the sharding's benefit forgone there, not a regression. The `[diag]` line exists so the first run on such a host can put a number on it.

### DIFF
--- a/docs/USAGE.adoc
+++ b/docs/USAGE.adoc
@@ -241,7 +241,7 @@ Sets the target CPU stack-sampling rate in samples per second per CPU.
 The default is 1000.
 Sampling runs in fixed-period mode: with `--sw-event` (cpu-clock) the rate is exact; with the cpu-cycles hardware event it is the rate at the CPU's maximum frequency, so a throttled CPU samples proportionally slower.
 The chosen period is recorded in the trace's `sysinfo.sample_period` column.
-Higher values give finer-grained profiles at the cost of more overhead and larger traces; very high rates may be throttled by the kernel's perf interrupt-rate limit and can increase ringbuf drops (raise `--ringbuf-size-mib` if needed).
+Higher values give finer-grained profiles at the cost of more overhead and larger traces; very high rates may be throttled by the kernel's perf interrupt-rate limit and can increase ringbuf drops (raise `--ringbuf-size-mib` if needed). Ring buffers are sharded per CPU, up to 64 rings per family at a constant family byte budget; `--ringbuf-shards N` sets the count explicitly (`8` is the fixed layout of older releases), and kernels before 6.8 stay at 8 by default because inserting each ring costs them an RCU grace period at start.
 
 === *--sw-event*
 
--- a/skills/systing-trace/SKILL.md
+++ b/skills/systing-trace/SKILL.md
@@ -135,6 +135,7 @@ sudo systing --only-recorder sched --only-recorder network -d 10 --output trace.
 | `--marker-threshold <N>` | Stop after N marker instant events — **requires `--continuous`** |
 | `--marker-duration-threshold <MS>` | Stop when a marker range exceeds MS ms — **requires `--continuous`** |
 | `--ringbuf-size-mib <N>` | Increase BPF ring buffer size if events are lost |
+| `--ringbuf-shards <N>` | Rings per ring-buffer family: `0` (default) = one per CPU up to 64 at a constant family byte budget, except on kernels before 6.8, which keep 8; `8` restores the fixed eight-ring layout exactly |
 | `--resolve-addresses` | Resolve network IPs to hostnames via DNS (off by default) |
 
 ### TPU
--- a/src/systing_core.rs
+++ b/src/systing_core.rs
@@ -499,6 +499,45 @@ const RINGBUF_FAMILIES: &[(&str, &str, bool)] = &[
     ("memory_ringbufs", "rb_mem_", false),
 ];
 
+/// The first kernel whose BPF_MAP_UPDATE_BATCH on a map-of-maps waits ONE RCU
+/// grace period per batch call rather than one per element. Before it
+/// (through every 6.6.y release) `generic_map_update_batch` reaches
+/// `maybe_wait_bpf_programs` -> `synchronize_rcu()` for each inserted inner
+/// ring, so inserting 64 rings into a family costs 64 grace periods at start,
+/// on a busy many-core host tens of milliseconds each, paid again on every
+/// capture. The auto shard count stays at the legacy eight below this
+/// version; an explicit `--ringbuf-shards` is always honoured.
+const MAP_OF_MAPS_BATCH_WAIT_ONCE_KERNEL: (u32, u32) = (6, 8);
+
+/// The running kernel's `major.minor`, from the release string (`6.12.0`,
+/// `6.18.44-custom.2`, `6.6.97-generic`); `None` when it cannot be read or parsed.
+pub(crate) fn kernel_release_major_minor() -> Option<(u32, u32)> {
+    parse_kernel_release(&sysinfo::System::kernel_version()?)
+}
+
+/// Parse `major.minor` out of a kernel release string, tolerating whatever
+/// follows the minor (`.0`, `-custom.2`, `+`). Anything short of two numbers is
+/// `None`.
+pub(crate) fn parse_kernel_release(release: &str) -> Option<(u32, u32)> {
+    let mut parts = release.split('.');
+    let major = parts.next()?.trim().parse::<u32>().ok()?;
+    let minor_part = parts.next()?;
+    let minor_digits: String = minor_part
+        .chars()
+        .take_while(|c| c.is_ascii_digit())
+        .collect();
+    let minor = minor_digits.parse::<u32>().ok()?;
+    Some((major, minor))
+}
+
+/// Does this kernel insert a batch of inner rings with one RCU grace period
+/// (6.8+)? An unreadable version counts as "no": the cost of being wrong that
+/// way is eight rings instead of sixty-four, the cost of the other way is a
+/// multi-second set-up on every capture.
+pub(crate) fn kernel_batches_map_of_maps_updates(kernel: Option<(u32, u32)>) -> bool {
+    kernel.is_some_and(|v| v >= MAP_OF_MAPS_BATCH_WAIT_ONCE_KERNEL)
+}
+
 /// How many rings each family gets and how big each is.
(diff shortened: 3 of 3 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/shards-kernel-cap-nofile`

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the branch's diff against `main` (v1.13.13; `src/systing_core.rs`, `docs/USAGE.adoc`, `skills/systing-trace/SKILL.md`) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
Two guards for the per-CPU ring sharding on the hosts it will meet: the automatic shard count stays at the legacy eight on kernels whose batch insert waits an RCU grace period per element (before 6.8), and the soft `RLIMIT_NOFILE` is raised to the hard limit at start with a printed projection of the run's descriptor need. Userspace only; no BPF, schema, recorder or dependency change. No critical issues.

### Critical Issues
None. `src/bpf/` untouched; `SCHEMA_VERSION` unchanged; `Cargo.toml` unchanged (`libc` was already a dependency); no ELF parsing, no Python version work.

### Improvements
1. **The kernel rule is a pure function with the version as an input.** `plan_ringbufs(num_cpus, size, override, kernel: Option<(u32,u32)>)` — the decision is unit-tested on both sides of 6.8, with `None`, and with an override; the one caller reads the version once (`kernel_release_major_minor()`, the same `sysinfo` source `is_old_kernel()` already uses) and prints it in the startup line, so a host's log says which rule applied.
2. **Failing closed on an unreadable version is the cheap direction.** `None` → the legacy eight: the cost of being wrong that way is the sharding's benefit forgone, the cost of the other way is a multi-second set-up on every capture. Stated in the doc comment.
3. **The limit raise never fails the run.** `raise_fd_soft_limit()` returns `None` only when `getrlimit` fails and reports a failed `setrlimit` as a warning; the caller prints the projection either way so an EMFILE that follows has a named cause. `rlim_t` is widened with `u64::from`, no truncating casts.
4. **The projection names its terms.** `projected_fd_need()` = rings × families × 3 + CPUs × (2 | 4 off x86 with the memory recorder) + 256, with each term explained where it is defined; the caller feeds it the families the run will actually create (`get_required_ringbuf_families` + the always-on stack family) and the same `is_recorder_enabled("memory")` predicate the fault leg uses.
5. **`--ringbuf-size-mib ≥ 4096` is refused before planning** (a `bail!` in `systing()`), so `plan_ringbufs`'s `expect("ring size fits in u32")` is unreachable rather than a reachable panic.

### Suggestions
- The `[diag] create_ring_shards … took …` line is behind `SYSTING_DIAG_RINGBUFS` like the other ring diagnostics; a future change could print it unconditionally when it exceeds a second, which is the case an operator wants to see.
- `MAP_OF_MAPS_BATCH_WAIT_ONCE_KERNEL` is a `(major, minor)` pair compared with tuple ordering; if a patch-level floor is ever needed the parser already discards the patch number and would need a third field.

### Positive Notes
- The shard cap changes nothing on kernels ≥ 6.8 (the existing `ring_plan_*` tests pass a batching kernel and are otherwise byte-identical) and restores the pre-sharding layout exactly below it.
- The doc comment on `create_ring_shards` now states the kernel facts correctly (per-slot wait before 6.8; map-of-maps batch support from 5.19) with the commit ids, so the next reader does not have to re-derive them.

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 527 + 11 + 28 passed, 0 failed.
- BPF: no change under `src/bpf/`; no map/context pairing to re-check (the ring inserts and the rlimit call are process-context userspace as before).
- VM run of `test_e2e_validation_suite` on the virtme-ng 6.12.0 guest (4 vCPU / 8 GiB): `1 passed; 0 failed; 38 filtered out`; guest startup lines `Ring buffers: 4 rings per family x 64 MiB (4 possible CPUs; kernel 6.12; one per CPU)` / `File descriptors: soft limit raised 1024 -> 4096 (hard 4096); this run needs about 288`.
- Not run: a kernel before 6.8 (the cap is unit-tested; its effect there is the pre-sharding layout exactly).

PR #213, opened 2026-08-24
analyze: bound the MCP `query` tool's result inside DuckDB — LIMIT at the cap, count(*) for the total, a fixed out-of-memory message

josefbacik:
## What changes

The MCP `query` tool's row cap now lives inside DuckDB. The statement runs as `SELECT * FROM (<sql>) LIMIT 10001`, so DuckDB never produces — and the binding never materializes — more than one row past `MAX_QUERY_ROWS`. When the result is truncated, `total_row_count` comes from a separate `SELECT count(*) FROM (<sql>)`. A DuckDB out-of-memory failure is reported as one fixed message that names the bound instead of echoing the engine's allocation text.

## Why

`AnalyzeDb::query` capped what it *kept* (10,000 rows) but not what it *fetched*. The statement ran as written, the binding materialized the whole result — as plain allocations outside DuckDB's `memory_limit` accounting — and the loop then walked every row only to report the total. An unbounded `SELECT` over a table with tens of millions of rows grew the process by gigabytes while DuckDB itself stayed inside its 4 GiB bound; in a memory-limited container that is an OOM kill of the whole MCP server, not a query error the caller can react to.

## What the reviewer decides

1. **The wrap-and-fall-back shape.** The wrapped statement is prepared first; if preparing it fails (a `SHOW`, `PRAGMA`, `SET`, `COPY`, `ATTACH` or multi-statement shape, or a SQL error), the raw statement runs exactly as before, including the old drain-and-count for the total, so existing callers and error texts are unchanged and a SQL error is reported against the user's own statement. A failure while *executing* the wrapped statement is reported directly — the raw statement is never re-run unbounded.
2. **The extra execution for the total.** `count(*)` over the same query re-runs its pipeline once with no row materialization: memory stays under DuckDB's bound (spillable operators), time is the query's own cost once more. Only truncated results pay it.
3. **The fixed out-of-memory wording**: `query exceeded the DuckDB memory bound (N MiB): narrow the time window, add LIMIT, or aggregate`, with N the cgroup-derived limit recorded at open (omitted when there is none). Every other error passes through unchanged.

The CLI display path (`query_strings`) is untouched. No schema, version, or BPF change.

## Tests

`analyze::tests::test_query_*`: the cap with an exact total on a 25,000-row table; below-cap and at-cap results not truncated; trailing terminator, trailing `--` comment, CTE and duplicate column names through the wrapper; `PRAGMA` and `SHOW` through the fallback; a SQL error naming the user's column without the wrapper; the out-of-memory classifier on both of DuckDB's message shapes; and an end-to-end sort under a 1 MB `memory_limit` with a 1 MB `max_temp_directory_size` that reports the fixed message.

### DIFF
--- a/src/analyze/flamegraph.rs
+++ b/src/analyze/flamegraph.rs
@@ -950,6 +950,7 @@ mod tests {
         AnalyzeDb {
             conn,
             path: std::path::PathBuf::new(),
+            mem_limit_mib: None,
         }
     }
 
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -246,6 +246,22 @@ fn extract_column_names(rows: &duckdb::Rows<'_>) -> (usize, Vec<String>) {
 pub struct AnalyzeDb {
     conn: Connection,
     path: PathBuf,
+    /// DuckDB's configured `memory_limit` in MiB, when one was derived from
+    /// the cgroup at open; named in the error a query gets when it exceeds it.
+    mem_limit_mib: Option<u64>,
+}
+
+/// Text DuckDB puts at the front of every `OutOfMemoryException` message,
+/// whether the buffer pool hit `memory_limit` or a spill hit
+/// `max_temp_directory_size` ("failed to offload data block"). Matched as a
+/// substring because the binding may wrap the engine's message.
+const DUCKDB_OOM_MARKER: &str = "Out of Memory Error";
+
+/// Strip trailing whitespace and statement terminators so the text can be
+/// embedded as a subquery.
+fn trim_statement(sql: &str) -> &str {
+    sql.trim()
+        .trim_end_matches(|c: char| c == ';' || c.is_whitespace())
 }
 
 impl AnalyzeDb {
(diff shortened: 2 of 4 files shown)


josefbacik:

Review mechanics: the repository's `rust-bpf-code-reviewer` agent definition was executed in-session over the full diff of this branch against `main` (four files: `src/analyze/mod.rs`, `src/mcp.rs`, and a one-line test-initializer addition each in `src/analyze/flamegraph.rs` and `src/analyze/sched_aggregate.rs`), all five passes.

### Summary
A contained change to `AnalyzeDb::query`: the row cap moves into the statement DuckDB executes, the truncated total moves to a separate `count(*)`, and DuckDB's out-of-memory class is mapped to one fixed message. No BPF, schema, recorder, Python-version or ELF-parsing code is touched; `Cargo.toml`'s version is untouched.

### Critical Issues
None.

### Improvements
- (Taken before commit) The wrapped and fallback arms each repeated "execute, extract columns, read rows up to the cap"; folded into one `run_capped(&mut Statement, drain)` helper so the cap logic exists once.

### Suggestions
- `trim_statement` strips trailing `;`/whitespace textually; a statement that ends inside a string literal containing `;` falls back to the raw path (graceful, and covered by the fallback contract). Noted, not changed.
- The fallback arm discards the wrapped-prepare error on purpose (the raw statement's own error is the one reported); the comment says so.

### Positive Notes
- The execution-failure path never re-runs the raw statement unbounded, which is the property the change exists for; the doc comment on `query` states it.
- Out-of-memory classification keys on DuckDB's own exception marker, matched as a substring so a binding-side wrapper cannot hide it, and every other error keeps the engine's text.
- Tests pin the cap, the total, the wrapper's shape handling, the fallback shapes, the error-text contract and the out-of-memory mapping end to end.

PR #214, opened 2026-08-25
Make the real-interpreter discovery tests count, and read the cache tag without the section

josefbacik:
## What changes

Follow-up to #211 (lazy pystacks discovery), in `src/pystacks/discovery.rs` only:

1. **The real-interpreter tests cannot pass vacuously any more.** `test_elf_py_info_real_interpreters` and `test_discovery_read_volume_real_python` find runtimes from two sources — pyenv `--enable-shared` builds under `$PYENV_ROOT/versions`, and the system interpreter at `/usr/bin/python3`, asked where its runtime lives (libpython for a shared build, the executable itself for a static one). With no runtime at all they fail under `CI` (set by GitHub Actions) and skip with a note elsewhere.
2. **The cache-tag read has a discriminating test.** Wherever a runtime carries `_PySys_ImplCacheTag`, `read_impl_cache_tag` must return exactly `cpython-3XX`; `(major, minor)` alone — the previous assertion — is also satisfied by the filename fallback in `detect_python_version`, so it could not tell the two apart.
3. **`read_impl_cache_tag` reads its bytes directly.** `Section::data_range` in object 0.36 pulls the whole containing section through the reader before slicing, which on a lazily-read libpython is the MBs of `.rodata`/`.data` the lazy path was written to avoid — and the opposite of what the function's own comment said. It now translates the virtual address to a file offset through the allocated section that holds it and reads the few bytes it needs with the `ReadRef` directly; only allocated sections with an address are candidates, so the lookup does not depend on section order.
4. Comments: the readability probe in `elf_py_info` is described as what it is (it catches a file that is gone or unreadable, not an error in the middle of an otherwise-readable one).

No behaviour change for discovery's results: the same symbol is found, the same bytes are interpreted; only how many bytes are read to get there.

## Why

#211's cold verification (a differential probe over twelve real binaries) found the fix right and the test rig weak: CI provisions no pyenv, so the green run proved nothing about the lazy reader on a real runtime, and the tag read — the most intricate hunk — had no test that could fail on it. Both are cheap to close, and the section-sized read is the one place the "reads only what it needs" claim was not yet true.

## How

- `real_runtimes()` (test helper) builds the runtime list: every pyenv version directory with `lib/libpython{M}.{m}.so.1.0` + `bin/python{M}.{m}`, plus `/usr/bin/python3`'s answer to a six-line probe (`version_info`, `Py_ENABLE_SHARED`, `LIBDIR`, `INSTSONAME`, `sys.executable`), de-duplicated by runtime path. `skip_or_fail_without_runtime()` is the one place the CI rule lives.
- The real-interpreter test asserts verdict (`is_dynamic == shared`), version and `_PyRuntime` presence for each runtime; opens it through `ReadCache` and, if `_PySys_ImplCacheTag` is present with a nonzero address in `.symtab` or `.dynsym`, asserts the tag string; prints `N runtimes verified, K of them through the cache tag`. The shared-build exe classification stays reported, not asserted (a launcher has no `_PyRuntime`; a static-linking build legally does).
- The read-volume test takes the first shared runtime (a static one if that is all there is) and keeps its 4× assertion against the previous scan order's byte count.
- `read_impl_cache_tag(elf, reader)`: `bytes_at(va)` = first section with `SHF_ALLOC`, a nonzero address, `va` inside it and a file range (`.bss`-style sections have none), then `reader.read_bytes_at(file_off + (va - s_addr), len)`; `detect_python_version` passes the reader through from `parse_elf_py_info`.

## Tests

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 537 + 11 + 28 passed, 0 failed.
- The two real-interpreter tests, run alone with output on a developer host with eight pyenv shared builds (3.8–3.14) and a static distro `python3.12`: `real-interpreter check: 9 runtimes verified, 7 of them through the cache tag` (the 3.13.14 and the stripped distro build carry no `_PySys_ImplCacheTag`; the rest do, and every tag read returned the expected `cpython-3XX`); the read-volume test: `discovery read 5171101 bytes; the previous scan order would have read 40457304 bytes across 20 files` (that run's interpreter is a 3.13 build without the tag symbol, so the figure is the symbol tables and unchanged by item 3). The same test with `CI=true` passes where runtimes exist; with none it fails by construction (the `assert!` is the only path to `return` under `CI`).
- Bytes item 3 saves on a tag-bearing build, from the section headers of pyenv's `libpython3.12.so.1.0` (31 MB): the old `data_range` reads were `.data` (1,040,320 B, where the symbol's pointer lives) and `.rodata` (1,027,501 B, where the string lives) in full — about 2.07 MB per binary — where the new path reads 64 bytes; the symbol tables (~0.86 MB) are read either way.

## Not changed

A synthetic-ELF unit test for the tag read (deterministic on any machine) would need a fixture binary in the repo; the real-runtime check covers the symbol-bearing builds a developer machine has (pyenv 3.8 exports it; 3.9–3.12 pyenv builds carry it as a local `.symtab` entry; 3.13+ and distro-stripped builds do not carry it, and those take the filename fallback as before).

### DIFF
--- a/src/pystacks/discovery.rs
+++ b/src/pystacks/discovery.rs
@@ -28,10 +28,10 @@ struct ElfPyInfo {
 
 /// Per-binary ELF parse results keyed by (st_dev, st_ino, st_size). `None`
 /// records a binary that is not a Python runtime — rejections cost a full
-/// ELF parse too, so they're worth remembering. I/O failures are never
-/// cached: a transient read error (ESTALE, a /proc/pid/root path vanishing
-/// when the process exits mid-check) must not blacklist an interpreter for
-/// the rest of the session.
+/// ELF parse too, so they're worth remembering. A file that fails a
+/// readability probe (see `elf_py_info`) is not cached: a transient read
+/// error (ESTALE, a /proc/pid/root path vanishing when the process exits
+/// mid-check) must not blacklist an interpreter for the rest of the session.
 type ElfCache = HashMap<(u64, u64, u64), Option<Arc<ElfPyInfo>>>;
 
 static ELF_CACHE: LazyLock<Mutex<ElfCache>> = LazyLock::new(Default::default);
@@ -67,6 +67,10 @@ fn elf_py_info(file_path: &str) -> Option<Arc<ElfPyInfo>> {
     // A rejection is cached only if the file is still readable: an I/O
     // error mid-parse (the file vanished under us) is indistinguishable
     // from "not a runtime" inside the parser, and must not be remembered.
+    // The probe reads the first 64 bytes, so it stands in for "the file is
+    // gone or unreadable", not for an error in the middle of a file that
+    // is otherwise fine — that narrower case would be cached as a
+    // rejection.
     if parsed.is_none() && !file_readable(&file) {
         return None;
     }
@@ -100,7 +104,7 @@ fn parse_elf_py_info(file_path: &str, file: &fs::File) -> Option<Arc<ElfPyInfo>>
     let reader = ReadCache::new(file);
     let result = object::File::parse(&reader).ok().and_then(|elf| {
         let py_runtime_addr = find_symbol_address(&elf, "_PyRuntime")?;
-        let version = detect_python_version(&elf, file_path)?;
+        let version = detect_python_version(&elf, &reader, file_path)?;
         Some(Arc::new(ElfPyInfo {
             py_runtime_addr,
             version,
@@ -370,10 +374,11 @@ fn find_symbol_address<'data, R: ReadRef<'data>>(
 /// so the pattern match is unaffected by /proc/pid/root resolution).
 fn detect_python_version<'data, R: ReadRef<'data>>(
     elf: &object::File<'data, R>,
+    reader: R,
     module_path: &str,
 ) -> Option<(i32, i32, i32)> {
     // Try to find version from _PySys_ImplCacheTag symbol value
-    if let Some(version_str) = read_impl_cache_tag(elf) {
+    if let Some(version_str) = read_impl_cache_tag(elf, reader) {
         if let Some(ver) = parse_cpython_version(&version_str) {
             return Some(ver);
         }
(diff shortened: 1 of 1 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/pystacks-discovery-tests`

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the branch's diff against `main` (v1.13.16; `src/pystacks/discovery.rs` only) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
A test-rig follow-up to #211 with one small production change: the real-interpreter tests find runtimes in two places and fail under CI when there are none, the cache-tag read gets a direct assertion, and `read_impl_cache_tag` reads its bytes through the `ReadRef` from the section's file range instead of pulling the whole section. No BPF, schema, recorder or dependency change. No critical issues.

### Critical Issues
None.

### Improvements
1. **The targeted read keeps the function's contract and fixes its comment.** `bytes_at` now requires `SHF_ALLOC`, a nonzero section address, a file range (so `.bss`-style sections are skipped) and `va` inside the section, then reads `min(32, bytes left in the section's file range)` from the file offset; the pointer-follow and direct-string branches are unchanged. The `ReadRef` is the same `&ReadCache` the parser runs on, passed down through `detect_python_version`.
2. **The CI rule lives in one helper.** `skip_or_fail_without_runtime` is the only place a skip can happen; it panics when `CI` is set, so the green-by-skip path cannot come back silently.
3. **The runtime locator asks the interpreter rather than guessing.** `/usr/bin/python3` reports `Py_ENABLE_SHARED`, `LIBDIR`, `INSTSONAME` and `sys.executable`; a shared build's runtime is `LIBDIR/INSTSONAME`, a static one's is the canonicalised executable; entries are de-duplicated by runtime path against the pyenv list.
4. **The tag assertion is gated on the symbol's presence, and counted.** A runtime without `_PySys_ImplCacheTag` (3.13+, stripped distro builds) takes the filename fallback and is not asserted on the tag; the test prints how many runtimes were checked through the tag so a vacuous count is visible in the log.

### Suggestions
- A fixture ELF in the repo would make the tag read testable on any machine; left out on purpose (binary in the tree, and the real-runtime check covers it on developer hosts).
- `real_runtimes()` could also consult `$PATH`'s `python3`; left at `/usr/bin/python3` so the result does not depend on a pyenv shim on the path.

### Positive Notes
- The previous two tests' bodies are preserved where they were right (the launcher-vs-runtime classification stays reported, not asserted; the read-volume 4× bar is unchanged).
- `object::elf::SHF_ALLOC` is compared against `SectionFlags::Elf { sh_flags }` with a `u64::from`, so the match is exact on both 32- and 64-bit ELF.

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass. `cargo test --lib --bins`: 537 + 11 + 28 passed, 0 failed.
- The two real-interpreter tests alone, with output, on a host with eight pyenv shared builds and a static distro python3.12: `9 runtimes verified, 7 of them through the cache tag`; `discovery read 5171101 bytes; the previous scan order would have read 40457304 bytes across 20 files`; `CI=true` run: pass (runtimes present).
- BPF: no change under `src/bpf/`.

PR #215, opened 2026-08-25
Name the cause when a --cgroup target cannot be resolved from the tracer's cgroup namespace

josefbacik:
## What changes

Follow-up to #209 (`--cgroup` matching decided by the kernel):

1. **A `--cgroup` target the kernel cannot resolve from the tracer's cgroup namespace is refused with the cause named.** In kernel mode the iterator program parks a reference to each target with `bpf_cgroup_from_id()`, and the kernel resolves that id within the calling process's cgroup namespace. From a container with a private cgroup namespace and the host's cgroup filesystem mounted (the `kubectl debug` / privileged-pod pattern), a host path is visible — the fd steps have no namespace check — but the id is outside the namespace, the iterator gets nothing, and systing failed on a bare map check (`cgroup_target_refs[0] holds cgroup id 0 after the iterator run, expected N`). It now says: the kernel could not resolve this cgroup from the tracer's cgroup namespace; run systing in the host cgroup namespace, or set `SYSTING_CGROUP_FILTER_LEGACY` to match a start-time snapshot instead.
2. **A new e2e test reproduces that case** (`test_e2e_cgroup_filter_outside_cgroupns_names_the_cause`): a child process joins a leaf cgroup, `unshare -C`s there so that leaf is its namespace root, and runs the `systing` binary with `--cgroup <sibling of the leaf>`. Kernel mode must exit non-zero with the cause and the knob named; the same command with the knob set must trace.
3. **`install_cgroup_targets` reads the id from a layout, not from bytes 8..16.** A `#[repr(C)]` mirror of the BPF side's `struct cgroup_target_ref { struct cgroup __kptr *cgrp; u64 id; }` supplies the offset (`offset_of!`), and the value's size is checked against the mirror, so a change on the BPF side fails loudly.
4. **The BTF probe has a unit test.** `kernel_has_task_under_cgroup_kfunc()` now calls `btf_has_func(&btf, name)`, tested against the running kernel's BTF with a function every kernel exports, an invented name, and a report of whether the kfunc is present.
5. Comments and docs: the stale BPF comment ("userspace refuses `--cgroup` on kernels without the kfunc" — it falls back) is corrected; `trace_task()`'s comment says its `task` argument feeds only the pid checks in kernel mode (the `--cgroup` decision is on `current` through the helper) and why there is no runtime guard; the README says any non-empty value of `SYSTING_CGROUP_FILTER_LEGACY` forces the fallback and carries the cgroup-namespace note.

## Why not an automatic fallback

The mode is a rodata constant (`tool_config.cgroup_match_kernel`) fixed before the skeleton loads, and the iterator runs after load; falling back automatically would mean re-opening and re-loading the skeleton in legacy mode for that run — a second BPF load and a second code path to test — for a case the host-level unit never hits (it runs in the root cgroup namespace). The named cause plus the knob is the honest cure for the human debug-container case. If the automatic fallback is wanted anyway, it is a contained change at the same bail site.

## Kernel facts the refusal rests on

`cgroup_get_from_id()` (`kernel/cgroup/cgroup.c`, 6.6 through 6.18) looks the id up in the cgroup kernfs and then checks that the cgroup descends from the calling process's cgroup namespace root (`cgroup_is_descendant(cgrp, root_cgrp)`), returning `-ENOENT` otherwise; `bpf_cgroup_from_id()` is a thin kfunc over it. The fd-based steps (`cgroup_targets`, the iterator attach by fd) have no namespace check, which is why the failure surfaced only at the id step.

## Tests

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass (re-run on the committed bytes). `cargo test --lib --bins`: 539 + 11 + 28 passed, 0 failed — `cgroup_target_ref_id_reads_the_id_beside_the_kptr` and `btf_func_probe_sees_real_functions_and_not_invented_ones` by name.
- `clang -target bpf` compile of `systing_system.bpf.c` with the build's include set: clean apart from the pre-existing `read_tcp_seq_from_skb` qualifier warning (the BPF change is two comments).
- VM run of the three `--cgroup` e2e tests (root + BPF) on a virtme-ng guest, kernel `6.12.0` (BTF-enabled build; 4 vCPU / 8 GiB, software-emulated CPU): **`3 passed; 0 failed; 39 filtered out`** — the kernel-mode and forced-legacy tests as before, and the new `test_e2e_cgroup_filter_outside_cgroupns_names_the_cause`: `Running systing --cgroup /sys/fs/cgroup/systing-cgn-252/outside from a private cgroup namespace rooted at /sys/fs/cgroup/systing-cgn-252/inside` → `refused with the cause named (correct)` → `forced fallback traced it (correct)`. To run them yourself: `sudo ./scripts/run-integration-tests.sh trace_validation test_e2e_cgroup_filter`.

## Not changed

The mode choice and both matching paths; the BPF programs (one comment); no schema, recorder, `Cargo.toml` version or dependency change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,7 +260,12 @@ decided by the kernel's `bpf_task_under_cgroup()`, which needs **Linux kernel
 BTF) systing falls back to matching a snapshot of the target's cgroups taken
 when the trace starts, so cgroups created under the target afterwards are not
 traced; the mode in use is printed at start. Setting
-`SYSTING_CGROUP_FILTER_LEGACY=1` forces the fallback on any kernel.
+`SYSTING_CGROUP_FILTER_LEGACY` to any non-empty value (`=1` will do) forces the
+fallback on any kernel. The kernel resolves the target within systing's own
+cgroup namespace: from a container with a private cgroup namespace and the
+host's cgroup filesystem mounted, a host path is visible but cannot be
+resolved, and systing says so at start — run it in the host cgroup namespace
+or use the fallback.
 
 `--trace-event` - This will add an `instant` track event for each event that
 this tool captures.  The format is "<trace type>:<optional
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -1561,9 +1561,10 @@ static bool current_in_cgroup_filter(void)
  * this KF_RCU kfunc). The rcu section holds no lock and takes no reference:
  * one map lookup and one ancestors[] compare per target.
  *
- * Kernels without the kfunc (< 6.5) never get here - userspace refuses
- * --cgroup on them - and bpf_ksym_exists() keeps every program that inlines
- * this loadable there for traces without --cgroup.
+ * Kernels without the kfunc (< 6.5) never get here - userspace selects the
+ * legacy snapshot matching on them (tool_config.cgroup_match_kernel == 0,
+ * see task_in_legacy_cgroup_set) - and bpf_ksym_exists() keeps every program
+ * that inlines this loadable there.
  */
 static bool task_in_cgroup_filter(struct task_struct *task)
 {
@@ -1773,6 +1774,12 @@ static bool trace_task_common(struct task_struct *task)
  * the sched tracepoints: the --cgroup test here is on current, through the
  * helper every program type may call). A caller in a tp_btf program that
  * tests a task other than current uses trace_task_btf() instead.
+ *
+ * The task argument feeds the pid/tgid checks only: in kernel mode the
+ * --cgroup decision is bpf_current_task_under_cgroup(), which judges current
+ * whatever task was passed. There is no runtime guard (a helper call per
+ * event on the sampling paths is not worth it), so a new caller must pass
+ * current or use trace_task_btf().
  */
 static bool trace_task(struct task_struct *task)
 {
--- a/src/systing_core.rs
+++ b/src/systing_core.rs
@@ -3133,9 +3133,7 @@ struct CgroupTarget {
 /// matching needs; without it the filter runs in legacy mode.
 fn kernel_has_task_under_cgroup_kfunc() -> bool {
     match libbpf_rs::btf::Btf::from_vmlinux() {
-        Ok(btf) => btf
-            .type_by_name::<libbpf_rs::btf::types::Func<'_>>("bpf_task_under_cgroup")
-            .is_some(),
+        Ok(btf) => btf_has_func(&btf, "bpf_task_under_cgroup"),
         Err(e) => {
             eprintln!(
                 "Warning: failed to open vmlinux BTF ({e}); \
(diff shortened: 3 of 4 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `tracer-oom/cgroup-target-unresolved`

Mechanics, disclosed: the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed in-session by the author (an AI agent session) over the branch's diff against `main` (v1.13.16; `src/systing_core.rs`, `src/bpf/systing_system.bpf.c` (comments only), `tests/trace_validation.rs`, `README.md`) in the five passes it prescribes (structure → Rust → BPF → schema/versioning → integration). Author and reviewer are the same session; this is the pre-push review the repo requires, not an independent second reader.

### Summary
The #209 follow-up: the one failure mode the kernel-side `--cgroup` matching has that the old code did not — a target the kernel cannot resolve from the tracer's cgroup namespace — is refused with the cause and the cure named, and reproduced by an e2e test that runs the binary under `unshare -C`. The id read out of `cgroup_target_refs` is derived from a layout mirror instead of hand-written byte offsets; the BTF probe gains a unit test; three comments and the README are corrected. No BPF program change. No critical issues.

### Critical Issues
None. The only change under `src/bpf/` is two comments; no map, program, rodata or schema change; `SCHEMA_VERSION` and `Cargo.toml` untouched.

### Improvements
1. **The refusal is specific.** `install_cgroup_targets` now distinguishes an EMPTY slot (id 0 — the iterator could not resolve the id in its namespace) from a MISMATCHED one (the old message), and the empty case names the mechanism and both cures. The detection is the value the iterator writes, so it fires exactly when the kernel answered `-ENOENT` inside the iterator, not on any other failure.
2. **Layout-derived read with a size check.** `CgroupTargetRef { cgrp: u64, id: u64 }` mirrors `struct cgroup_target_ref { struct cgroup __kptr *cgrp; u64 id; }` (a kptr is 8 bytes on the BPF target); `cgroup_target_ref_id()` refuses any value whose length is not `size_of::<CgroupTargetRef>()` and reads at `offset_of!(…, id)`. The unit test pins the offset at 8 and the size at 16, so a change on either side is a test failure before it is a mis-read.
3. **The probe is testable without a synthetic BTF.** `btf_has_func(&Btf, name)` is exercised against the running kernel's BTF with positive (`schedule`, `do_exit`) and negative (an invented name) controls; the kfunc's own presence is reported, not asserted, so the test passes on every kernel.
4. **The e2e test runs the binary, not the library.** `unshare -C` changes the calling process's namespace, so the test spawns `sh -c 'echo $$ > <leaf>/cgroup.procs && exec /usr/bin/unshare -C systing --cgroup <sibling> …'` and leaves the test process's namespaces alone; the second half proves the named cure (the knob set → a trace). It skips with a note where cgroup v2, `/usr/bin/unshare` or kernel mode is absent, and only the first half is conditional on kernel mode.

### Suggestions
- An automatic fallback at the bail site (re-open + re-load the skeleton in legacy mode) is possible and contained; not done here for the reasons in the body. If it is wanted, the e2e test's first half becomes "traces, with the legacy mode line", one assertion flipped.
- `trace_task()` still has no runtime guard against a non-`current` caller in kernel mode; the comment now says so and why (one helper call per event on the sampling paths). A debug build assertion would need `bpf_get_current_task_btf()` per call — left out.

### Positive Notes
- The fd-based steps' namespace-agnosticism versus the id step's namespace check is stated where the failure is detected, so the next reader does not have to rediscover it from `cgroup_get_from_id()`.
- The README sentence gives a human in a debug container the two moves in one line.

- `cargo fmt --all -- --check`: pass. `cargo clippy --all-targets -- -D warnings`: pass on the committed bytes. `cargo test --lib --bins`: 539 + 11 + 28 passed, 0 failed (the two new unit tests by name).
- `clang -target bpf` dry compile of `systing_system.bpf.c`: clean except the pre-existing warning. Map/context pairing: no map or program change; `LRU_HASH` under `src/bpf/` → 0 hits.
- VM e2e run on the 6.12 guest: `3 passed; 0 failed; 39 filtered out` — kernel mode, forced legacy, and the cgroup-namespace test (`refused with the cause named (correct)` / `forced fallback traced it (correct)`). The first rig attempt ran only the two older tests because the new one lacked the `#[ignore]` the rig's `--ignored` run selects on; the attribute was added and the second attempt ran all three.

PR #216, opened 2026-08-25
analyze: close two ways around the MCP `query` tool's bounds — one statement per call, and a locked configuration

josefbacik:
## What changes

Two ways around the bounds systing#213 put on the MCP `query` tool are closed. A statement is classified before anything runs: comments and stray `;` are stripped, a second `;`-separated statement is refused, and a statement that can read rows (`SELECT`, `WITH`, `FROM`, `VALUES`, `TABLE`, `PIVOT`, `UNPIVOT`, and `CALL` / `EXECUTE`, which run whatever they name) only ever runs inside the `SELECT * FROM (...) LIMIT 10001` wrapper — never as written. The MCP server also locks DuckDB's configuration right after opening a database, so `SET memory_limit`, `SET threads`, `PRAGMA max_temp_directory_size` and every other setting change are refused by DuckDB itself.

## Why

#213's fallback — run the statement as written when it cannot be wrapped — had two holes. (1) `SELECT ... FROM big; -- note` keeps an internal `;`, the wrapped prepare fails on it, and the raw statement ran unbounded and fully materialized, exactly the pre-#213 shape the tool description says cannot happen. (2) `SET memory_limit = ...` and friends took the same path and ran, so SQL arriving through the tool could lift the memory and spill bounds the server was started with; DuckDB's own out-of-memory text used to suggest doing precisely that, and a model acting on it grows the process past its cgroup.

## What the reviewer decides

1. **Locking the configuration on the MCP path** (`SET lock_configuration = true` after open; irreversible for the connection). It closes the whole settings family rather than a name denylist that `RESET`, `SET GLOBAL` or `PRAGMA memory_limit` would walk around. Cost: nothing on the MCP side can tune DuckDB in-session any more — including `search_path` — which no tool or prompt does today. The CLI keeps its unlocked behaviour.
2. **Refusing, not running, a statement that can read rows but that the wrapper cannot take.** A SQL error in it is still reported against the user's own text (the raw statement is prepared, not executed). `CALL` and `EXECUTE` land here by design — `CALL pragma_table_info('t')` becomes a fixed "rewrite it as a single SELECT" error and `SELECT * FROM pragma_table_info('t')` is the bounded form; `EXECUTE` of a prepared statement would otherwise run unbounded. I found no valid single `SELECT` that DuckDB accepts bare but rejects as a subquery; if one exists it gets the same error instead of an unbounded run.
3. **One statement per call.** Multiple statements were never useful through this tool; they are now an error naming the rule.

Smaller: the out-of-memory marker is matched as a prefix (the binding prints the engine's message bare), so a string literal cannot trigger the rewrite; a truncated result whose separate `count(*)` fails now carries `total_row_count: null` instead of omitting the field; the tool description says all of this. No schema, version, or BPF change.

## Tests

`analyze::query::tests::test_statement_shape_*` (comments, terminators, literals, nesting, multiple/empty, the keyword split); `analyze::tests::test_query_comments_and_terminators_cannot_bypass_the_cap`, `test_query_refuses_row_readers_it_cannot_wrap` (`EXECUTE`, `CALL`), `test_query_cap_boundaries_and_order` (`ORDER BY ... DESC`, exact cap + 1), `test_query_result_json_shape`, `test_lock_configuration_refuses_settings` (nine spellings refused, reads unaffected), the literal case in `test_query_out_of_memory_is_reported_as_the_bound`; `mcp::tests::test_get_or_open_locks_configuration`. `cargo test --lib` 547 passed / 0 failed, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -43,7 +43,7 @@ use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 
-use query::{duckdb_value_to_json, duckdb_value_to_string};
+use query::{duckdb_value_to_json, duckdb_value_to_string, statement_shape, StatementShape};
 
 /// Distinguishes the spill directories of databases opened by one process:
 /// DuckDB names temp files by block id, which is only unique per instance, so
@@ -118,19 +118,36 @@ pub const MAX_QUERY_ROWS: usize = 10_000;
 const MAX_TRACE_INFO_PROCESSES: usize = 25;
 
 /// Result of a SQL query.
-#[derive(Debug, Serialize)]
+///
+/// Serialized as `columns`, `rows`, `row_count`, and — only when the result
+/// was cut at `MAX_QUERY_ROWS` — `truncated: true` plus `total_row_count`,
+/// which is `null` when the separate count could not be computed.
+#[derive(Debug)]
 pub struct QueryResult {
     pub columns: Vec<String>,
     pub rows: Vec<Vec<serde_json::Value>>,
     pub row_count: usize,
-    #[serde(skip_serializing_if = "is_false")]
     pub truncated: bool,
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub total_row_count: Option<usize>,
 }
 
-fn is_false(b: &bool) -> bool {
-    !b
+impl Serialize for QueryResult {
+    fn serialize<S: serde::Serializer>(
+        &self,
+        serializer: S,
+    ) -> std::result::Result<S::Ok, S::Error> {
+        use serde::ser::SerializeStruct;
+        let fields = if self.truncated { 5 } else { 3 };
+        let mut s = serializer.serialize_struct("QueryResult", fields)?;
+        s.serialize_field("columns", &self.columns)?;
+        s.serialize_field("rows", &self.rows)?;
+        s.serialize_field("row_count", &self.row_count)?;
+        if self.truncated {
+            s.serialize_field("truncated", &true)?;
+            s.serialize_field("total_row_count", &self.total_row_count)?;
+        }
+        s.end()
+    }
 }
 
 /// Table information.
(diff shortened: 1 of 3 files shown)


josefbacik:

Review mechanics: the repository's `rust-bpf-code-reviewer` agent definition was executed in-session over the full diff of this branch against `main` (three files: `src/analyze/query.rs`, `src/analyze/mod.rs`, `src/mcp.rs`; +516/−43 at commit `4450370d`), all five passes. No BPF, schema, recorder, Python-version or ELF-parsing code is touched; `Cargo.toml`'s version is untouched.

### Summary
Closes two bypasses of the `query` tool's bounds (systing#213's follow-up): a text classifier strips comments and terminators and refuses multi-statement input before anything runs, a statement that can read rows is never executed raw, and the MCP server locks DuckDB's configuration after open so `SET`/`RESET`/`PRAGMA` cannot lift the memory or spill bounds. Two smaller corrections ride along: the out-of-memory marker is a prefix match, and a truncated result whose count failed serializes `total_row_count: null`.

### Critical Issues
None.

### Improvements
- (Taken before commit, found in this review) The classifier's first draft tested `text.contains(';')` after stripping, which would have refused `SELECT 'a;b' FROM t` as two statements; it now records the byte offsets of separators seen outside quotes and checks only those that fall inside the trimmed statement.
- (Taken before commit) `(\n(SELECT …))` — parentheses separated by whitespace — slipped past the keyword check because only leading `(` were trimmed; parens and whitespace are now trimmed together.
- (Taken before commit) `CALL` and `EXECUTE` can read rows (they run whatever they name) and were not in the must-wrap set; both are in it now, so `EXECUTE q` after a `PREPARE` in the same session is refused instead of running unbounded. Covered by `test_query_refuses_row_readers_it_cannot_wrap`.

### Suggestions
- Dollar-quoted strings (`$$ … $$`) are not tokenized; a `;` inside one reads as a separator and the statement is refused (fails closed). The doc comment says so; no trace-analysis query uses them.
- `lock_configuration` also freezes `search_path` and every other session setting on the MCP connection; nothing in the tools or prompts sets one today, and the PR body flags it as the reviewer's call.
- The raw path is still drain-and-count for the statements left on it (`SHOW`, `PRAGMA`, `DESCRIBE`, `SUMMARIZE`, `EXPLAIN`); their results are bounded by their nature, not by this code.

### Positive Notes
- A statement that can read rows has exactly one execution path (the wrapper); the raw prepare exists only to report the user's own error text, and the code says so at the site.
- The lock uses DuckDB's own irreversible mechanism rather than a name denylist, so `RESET`, `SET GLOBAL`, `PRAGMA memory_limit` and aliases like `max_memory` are all refused by the engine — the test enumerates nine spellings and checks the setting did not move.
- Tests cover the tokenizer in isolation (comments, nesting, literals, terminators, keyword split), the cap end to end on every spelling that used to bypass it, the boundary (`ORDER BY … DESC`, exactly cap + 1), the JSON shape, the lock on both the library and the MCP `get_or_open` path, and the literal false-positive on the OOM marker.

PR #217, opened 2026-08-25
memory recorder: keep the capture when the page-fault leg cannot attach, and record the memory configuration in sysinfo [bump minor]

josefbacik:
memory recorder: keep the capture when the page-fault leg cannot attach, and record the memory configuration in sysinfo [bump minor]

## What this changes

On a non-x86 host, if the memory recorder's page-fault leg (one `PERF_COUNT_SW_PAGE_FAULTS` software event per CPU, #208) fails to open or attach for any reason other than a missing CPU, the capture now continues without it instead of failing outright. The CPU stack sampler and the other memory legs (rss, mmap/munmap/brk) run as before; one line goes to stderr; and the trace's `sysinfo` row says what happened.

Four nullable columns are added to `sysinfo` (schema 17, `[bump minor]` per SCHEMA_CHANGES.md):

- `memory_fault_leg` — `tracepoint` (x86), `perf_sw` (every other arch), or `off:<cause>` (`off:EMFILE`, `off:EPERM`, `off:errno=<n>`, `off:error`).
- `memory_fault_sample_rate`, `memory_map_sample_rate`, `memory_alloc_sample_rate` — the `--memory-*-sample-rate` values the capture ran with.

All four are NULL when the memory recorder was not enabled (`memory_alloc_sample_rate` also when `memory-alloc` was off) and in traces from older systing; a pre-1.14 `sysinfo.parquet` still imports (the import is BY NAME). `systing-analyze trace info` and the MCP `trace_info` tool report them under `system`.

## Why

Before #208 the fault leg was the x86 tracepoint, attached with the rest of the skeleton. The perf-event leg that #208 added for arm64 is opened per CPU *after* the clock sampler, and its open/attach error propagated with `?` (`systing_core.rs`, `attach_clock_sampler`), so on an arm64 host with the memory recorder on, any failure of that open or attach other than a missing CPU — an EPERM from a hardened perf_event_paranoid, an ENOENT/EINVAL from a kernel that rejects the sampled software event, ENOMEM — would have refused the whole capture, CPU profile included, on every tick. That is the wrong trade: a profile without page-fault samples is still a profile, while a host with no profile at all is dark.

One cause is deliberately *not* in that list: descriptor exhaustion. Measured in the VM rig (below), RLIMIT_NOFILE exhaustion never reaches the fault leg — the ring-buffer setup and the clock sampler open first and hit EMFILE before it, and the fault leg's 2×CPU opens fit into descriptor numbers the skeleton setup leaves free. So EMFILE blacks the capture out regardless of this change, and #212 (raise the soft limit to the hard limit at start, print the projected need counting 4 fds per CPU on a non-x86 host with the memory recorder) is the answer for that cause. This change covers what remains, and, more usefully day to day, makes the trace say which fault leg ran and with what rates.

A consumer then needs to tell "no page faults were sampled" from "the leg was not running", which is what `memory_fault_leg` is for. The sample rates ride along because SCHEMA_CHANGES.md (v6) has said since the recorder shipped that they are "a capture-time setting, not carried in the trace, so a consumer that needs absolute figures reads the rate from the recorder's configuration" — now the trace carries them.

## Flagged choices

1. **Degrade, never refuse, for the fault leg only.** The clock sampler keeps its old behaviour: a capture without its CPU stack sampler is not a capture. ENODEV keeps its old meaning in `open_events` (fewer online CPUs than `num_cpus`: break, keep the earlier CPUs' events, no error) — unchanged and not widened.
2. **Partial attach leaves nothing behind.** `attach_page_fault_sw_events` now collects its links locally and returns them only on full success; on an error part-way through the CPUs, the links already attached are dropped (detached) together with the perf fds, so a degraded capture holds no half-attached leg.
3. **The cause is an errno name, not the message.** `off:EMFILE` is meant to be grouped on across a fleet; the full error text goes to stderr. The errno is read from the error chain's raw OS error when present, else from the `(os error N)` suffix the standard library prints — `perf.rs` wraps the perf open error's text into a new `io::Error` (losing the raw errno) and libbpf-rs does not expose its inner `io::Error` through `source()`, so the text is the one place the number survives both paths. Names are given for the errnos these opens produce (EMFILE, ENFILE, EPERM, EACCES, EINVAL, ENOMEM, EBUSY, EOPNOTSUPP); anything else is `errno=<n>`.
4. **`error_code` documented as x86-only.** SCHEMA_CHANGES.md's `memory_fault` line now says the column is the x86 page-fault error code and 0 elsewhere; 0 is an unambiguous sentinel because a user-mode fault on x86 always carries `X86_PF_USER`.

## Not changed (noted for the record)

- `open_events` breaking on the first ENODEV means CPUs after an offline hole get no event on either the clock or the fault leg (pre-existing, shared with the clock path; the skeleton's `cpus` list is `0..num_cpus`). Left as is; say if you want it iterated instead.

## Tests

- Unit: `MemoryFaultLeg::as_sysinfo_value` values; `fault_leg_failure_cause` from a perf.rs-wrapped text (`EMFILE`), from a raw OS error (`EPERM`, `errno=<n>`), and without an errno (`error`); the parquet writer's sysinfo round trip with the four columns; a pre-1.14 `sysinfo.parquet` (without the columns) importing into the v17 schema with NULLs; `trace info` reporting the fields and reading them as None on a pre-v9-style database.
- x86 build: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --lib` 543 passed / 0 failed.
- VM rig (virtme-ng, 6.12 guest, 4 vCPU, x86 with the perf-sw path flipped on in a scratch-only edit, never committed): `memory_recorder` integration tests 7/7 pass (`test_memory_recorder_e2e`, the four `memory_alloc` tests, the two `rss_threshold` tests). Injected fault-leg failure (an invalid perf event config in the same scratch edit): the capture exits 0 with 11,475 CPU stack samples, the other memory tables populated, `memory_fault` empty and `sysinfo.memory_fault_leg = off:ENOENT`. Descriptor sweep (hard RLIMIT_NOFILE lowered step by step from the run's peak of 174 open fds): 171–174 fully normal (`perf_sw`, 27k `memory_fault` rows); 170 fails in the ring-buffer map dup; 158–168 fail in the clock sampler's perf open — the fault leg is never the one that runs out, which is the finding stated under "Why".
- Real arm64: not exercised here; the first fleet read is the next pin wave on the Graviton hosts that already run the memory recorder.
- Rebased onto v1.13.19 after #215 and #216 merged. The one conflict was in `systing_core.rs`'s test module, where #215 and this change both append tests at the end — both kept, #215's first; `analyze/mod.rs` and `mcp.rs` (also touched by #216) merged without conflict, and a range-diff against the pre-rebase commit shows no other delta. The gates above and both VM rig runs were repeated on the rebased commit: fmt / clippy clean, `cargo test --lib` 555 passed / 0 failed (the twelve tests #215 and #216 added included), `memory_recorder` 7/7, injected fault-leg failure → `off:ENOENT` with the capture intact.

Hazard precedent (git log v1.12.0..v1.13.16 for the fd-limit / fault-leg class): #212 (`be22495`, NOFILE raise + projection) and #208 (`85ce761`, the perf-sw leg). This change adds no per-tick or hot-path work; it removes a failure mode and writes one sysinfo row per capture.


josefbacik:
# rust-bpf-code-reviewer verdict — fault-leg degrade + memory sysinfo columns

Mechanics disclosed: the author executed the repository's `.claude/agents/rust-bpf-code-reviewer.md`
definition in-session over the full diff of this branch vs `v1.13.16` (10 files, +481/−24). No separate
agent process ran.

### Summary
The change is small and correctly scoped: one optional leg of the memory recorder stops being fatal to the capture, and the trace's `sysinfo` row gains four nullable columns recording the memory recorder's capture-time configuration. Schema versioning is complete, the parquet → DuckDB import remains name-keyed so older traces still load, and the tests cover the new values, the errno naming and backward compatibility. Ship.

### Critical Issues
None.
- Schema & versioning: `SCHEMA_CHANGES.md` carries a v17 entry; `SCHEMA_VERSION` 16 → 17 in `src/duckdb.rs`; `Cargo.toml` / `Cargo.lock` untouched; the PR title carries `[bump minor]`.
- Recorder consistency: no new per-event table; the new columns sit on the one-row `sysinfo` table (no `utid`/`pid` question).
- No unsafe added; no BPF change; no ELF parsing.

### Improvements
1. `errno_cause_name(Option<i32>)` — after the kind-name fallback was added, the `None` arm is reachable only from its unit test; the function could take `i32` and the caller `if let Some(n) = errno { return errno_cause_name(n) }`. Style, not correctness — left as is for this revision (the tests exercise both arms); a follow-up nit.
2. `MemoryRecorderConfig` derives `PartialEq, Eq` without a comparer — harmless; could drop to `Clone, Debug`.

### Suggestions
- The `(os error N)` text parse in `os_error_number` is justified in the doc comment (perf.rs rewrites the open error into a new `io::Error`, libbpf-rs hides its inner `io::Error`); if perf.rs ever preserves the raw errno (e.g. via `io::Error::new` with a `source`), the raw-errno path takes precedence and the text parse becomes a dead fallback — fine.
- `#[cfg_attr(target_arch = "x86_64", allow(dead_code))]` on the three helpers keeps the x86 build warning-free while tests exercise them on every arch; `#[cfg(any(not(target_arch = "x86_64"), test))]` would be the stricter form.

### Positive Notes
- `attach_page_fault_sw_events` now owns its links until full success, so a failure part-way through the CPUs leaves no half-attached leg — the cleanup is by drop, with no new unwind logic.
- The degrade decision is one `match` at the one call site, with the x86 path untouched (`Tracepoint` is recorded without any new attach work).
- The four new `sysinfo` columns are appended in the same order in `models.rs`, `schema.rs`, the writer's batch builder and the DuckDB DDL, and `test_sysinfo_duckdb_import_without_memory_columns` pins the backward-compatible import.
- `trace info` / MCP `trace_info` expose the fields with the same NULL-substitution pattern the v9 platform columns use, including the pre-v9 emulation test.

Gates on the reviewed bytes: `cargo fmt --check` 0, `cargo clippy --all-targets -- -D warnings` 0, `cargo test --lib` 543 passed / 0 failed (2026-08-25 10:02Z, x86_64 host). VM rig (virtme-ng, 6.12 guest, x86 with the perf-sw path flipped on in a scratch-only edit): `memory_recorder` 7/7 PASS; injected fault-leg failure → capture exit 0, CPU stack samples intact, `memory_fault` empty, `sysinfo.memory_fault_leg = off:ENOENT`.

josefbacik:
# rust-bpf-code-reviewer verdict — addendum for the rebase onto v1.13.19

Scope: the resolution delta only. `git range-diff 4ff80f81..e32ef02a 5ecb24ff..54e32d9f` shows one `!` row, and its non-context lines are four lines of hunk context in `src/systing_core.rs`'s `mod tests`: #215 appended two tests (`cgroup_target_ref_id_reads_the_id_beside_the_kptr`, `btf_func_probe_sees_real_functions_and_not_invented_ones`) at the same end of the module where this change appends its four, so the conflict was the two appends; both were kept in order, #215's first, with the `#[test]` attribute restored on the first of ours. No code outside the test module changed in the resolution; `src/analyze/mod.rs` and `src/mcp.rs` (which #216 also touched, in the `query` / `lock_configuration` region and the `get_or_open` / test region) merged without conflict and carry no delta in the range-diff. Diff stat vs main is unchanged (10 files, +481/−24).

### Critical Issues
None. No schema, Cargo.toml, BPF or recorder change in the resolution; the message keeps exactly one `Co-authored-by` trailer.

### Positive Notes
- Nothing of #215 or #216 was displaced: their hunks are under this change byte-identical (the `=`-class context in the range-diff), and the four memory-lane tests still run by name.

Gates on the rebased bytes (`54e32d9f`): see the Verification bullet in the PR body (fmt / clippy `-D warnings` / `cargo test --lib`, and the VM rig's `memory_recorder` + injected-failure runs).

PR #218, opened 2026-08-25
analyze: lock the configuration of the database the MCP server starts with, and refuse dollar-quoted text before anything reaches prepare()

josefbacik:
## What changes

Two holes left by #216 in the MCP `query` tool's bounds are closed. (1) The configuration lock now applies to the database the server is started with (`mcp -d <path>`), not only to databases opened lazily by path — both open sites go through one `open_locked` helper. (2) Any bare `$` outside a quoted run (dollar-quoted strings, `$n` parameters) is refused before the text reaches `prepare()`.

## Why

Both were found by a cold read of the shipped v1.13.19 build.

1. #216 placed `lock_configuration()` in the lazy-open branch. The command-line database is opened and cached before the first request, so the lazy branch finds it present and skips the lock: `SET memory_limit = '100GB'` through the tool was accepted on exactly the launch shape the lock was written for.

2. The classifier did not know dollar-quoting, so after `$$'$$` it believed the rest of the text was a literal and saw no `;`. That is worse than a mis-count because duckdb-rs's `Connection::prepare()` is not parse-only: it extracts every statement from the text, **executes all but the last** (materialized, outside `memory_limit` accounting) and prepares only the last. So `SELECT $$'$$; SET memory_limit='100GB'; SELECT $$'$$` ran the SET on the raw-prepare path, and `SELECT $$'$$) AS x; <anything>; SELECT * FROM (SELECT $$'$$` ran the middle statement inside the wrapper with no error.

## What the reviewer decides

1. **Refusing every bare `$`** rather than tokenizing dollar-quoting. The tool has no use for `$$…$$` or `$1`, and refusing is the shape that cannot be wrong: the classifier's job is to agree with DuckDB's lexer about where a statement ends, and with `$` gone every remaining possible disagreement (an `E'\''` escape string DuckDB reads as longer) makes the classifier see *more* separators and refuse — never fewer. The cost is that an identifier like `a$b` is refused too; none exist in trace schemas.
2. **Not counting statements with `duckdb_extract_statements`.** That would be the robust form, but duckdb-rs exposes no raw connection handle outside the crate; the doc comments record the `prepare()` fact so the next change to this path knows the constraint.

No schema, version, or BPF change.

## Tests

`mcp::tests::test_initial_db_is_locked` (drives `DbHandle::new(Some(path), …)` through the request channel: `SET memory_limit` refused, spill cap in force, reads work); `analyze::query::tests::test_statement_shape_refuses_bare_dollar` (`$$'$$` in both repro shapes, `$tag$`, `$$a--b$$`, `$1`, `a$b`; a `$` inside a literal or comment stays ordinary text); the two repros plus `$$a--b$$` added to `analyze::tests::test_query_comments_and_terminators_cannot_bypass_the_cap` with `memory_limit` asserted unchanged after them. `cargo test --lib`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` — results in the PR's first comment.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -417,16 +417,22 @@ impl AnalyzeDb {
     ///
     /// The text is classified first (`statement_shape`): comments and
     /// leading/trailing `;` are stripped so they cannot hide a statement
-    /// from the wrapper, and more than one statement is refused. A
+    /// from the wrapper, more than one statement is refused, and so is any
+    /// bare `$` (dollar-quoted strings, `$n` parameters) — the one lexical
+    /// construct under which the classifier and DuckDB could disagree about
+    /// where a statement ends. That agreement matters: the binding's
+    /// `prepare` is not parse-only, it EXECUTES every statement but the last
+    /// of the text it is handed, so nothing may reach `prepare` unless the
+    /// classifier has established that it is exactly one statement. A
     /// statement that reads rows (`SELECT`, `WITH`, `FROM`, `VALUES`, ...)
     /// runs ONLY in wrapped form: if it cannot be wrapped it is refused, and
     /// a SQL error in it is reported against the user's own text by
-    /// preparing — never executing — the raw statement. Statements that
-    /// cannot be embedded as a subquery (`SHOW`, `PRAGMA`, `DESCRIBE`,
-    /// `SUMMARIZE`, `EXPLAIN`, ...) run as written with the cap applied as
-    /// rows are read; their results are small by nature. A failure while
-    /// *executing* the wrapped statement is reported directly — the raw
-    /// statement is never re-run unbounded.
+    /// preparing the raw statement — safe only because it is one statement.
+    /// Statements that cannot be embedded as a subquery (`SHOW`, `PRAGMA`,
+    /// `DESCRIBE`, `SUMMARIZE`, `EXPLAIN`, ...) run as written with the cap
+    /// applied as rows are read; their results are small by nature. A
+    /// failure while *executing* the wrapped statement is reported directly
+    /// — the raw statement is never re-run unbounded.
     ///
     /// Execution-time errors from the wrapped form name line numbers
     /// relative to the user's text plus the wrapper's first line.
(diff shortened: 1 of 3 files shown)


josefbacik:

Review mechanics: the repository's `rust-bpf-code-reviewer` agent definition was executed in-session over the full diff of this branch against `main` (three files: `src/mcp.rs`, `src/analyze/query.rs`, `src/analyze/mod.rs`; +131/−26 at commit `0b2ec8c8`), all five passes. No BPF, schema, recorder, Python-version or ELF-parsing code is touched; `Cargo.toml`'s version is untouched.

### Summary
Closes the two holes a cold read found in the previous change to the MCP `query` tool: the configuration lock now covers the database the server is started with (`mcp -d <path>`), through one `open_locked` helper both open sites use; and the statement classifier refuses any bare `$` outside a quoted run, because the binding's `prepare()` executes every statement but the last of the text it is given and a dollar-quoted string was the one construct under which the classifier could under-count statements.

### Critical Issues
None.

### Improvements
- (Taken before commit) The initial-database branch and the lazy-open branch each opened and cached a database on their own; folded into `open_locked` so a third open site cannot be added unlocked by accident.
- (Taken before commit) The `query` doc comment now states the `prepare()`-executes-all-but-last fact and the invariant it imposes — nothing reaches `prepare` unless the classifier has established exactly one statement — where the next change to this path will read it.

### Suggestions
- Refusing every bare `$` also refuses identifiers like `a$b`, which DuckDB's lexer allows; none exist in trace schemas, and the refusal message names the alternative. Noted in the body as the reviewer's call.
- The robust form — counting statements with `duckdb_extract_statements` before any `prepare` — needs a raw connection handle that duckdb-rs does not expose outside the crate; recorded in the body so the constraint is not rediscovered.
- The escape-string case (`E'\''`) is the remaining lexer divergence and goes the safe way (the classifier sees more separators and refuses); a unit case for it would pin that direction. Not added here.

### Positive Notes
- `test_initial_db_is_locked` exercises the real launch shape end to end — `DbHandle::new(Some(path))`, the database thread, a request with no path — and fails correctly on both failure modes (an unlocked initial database, or an initial open that failed and left no database to serve).
- The bypass test asserts the hidden `SET` did not run (`memory_limit` unchanged after the refused texts), not only that an error came back.
- The classifier's refusal is a single early return with no new state; the refused shapes are enumerated in a unit test alongside the shapes that must keep passing (a `$` inside a literal or a comment).

PR #219, opened 2026-08-25
memory recorder: VFIO DMA regions, IOMMU run-size histogram, THP splits and vmstat deltas [bump minor]

josefbacik:
memory recorder: VFIO DMA regions, IOMMU run-size histogram, THP splits and vmstat deltas

## What this answers

Three questions about device memory on a host, asked from the memory
recorder so the answer is one capture instead of three ad-hoc tools:

1. **How often are VFIO regions mapped and unmapped, by whom, how big?**
   `memory_vfio`: one row per `VFIO_IOMMU_MAP_DMA` / `UNMAP_DMA` region
   (kprobes on `vfio_dma_do_map` / `vfio_dma_do_unmap`), with the calling
   thread's stacks.
2. **How fragmented is the memory behind those mappings?** `memory_iommu`:
   a per-process, per-GiB-of-IOVA histogram of `iommu_map()` /
   `iommu_unmap()` run sizes from the `iommu:map` / `iommu:unmap`
   tracepoints. vfio_iommu_type1 pins a region one physically contiguous
   run at a time and maps each run with one `iommu_map()` call (and unmaps
   by "the largest contiguous physical chunk"), so the run size IS the
   contiguity of the backing memory: a THP-backed arena maps in ≥2 MiB
   runs, a 4 KiB-fragmented one in 4 KiB runs. The share of bytes at
   `size_order >= 21` is the huge-page-backed share; `count` per region
   (joined to `memory_vfio`) is the number of unmap/unpin walk entries the
   release will pay for; `iova_gib` says where in the device address space
   the fragmented part sits.
3. **How often do huge pages split, and do processes get them at all?**
   `memory_vmstat` (start/end `/proc/vmstat` samples of the THP, compaction
   and direct-reclaim counters — always on with the memory recorder, no
   probe), `memory_thp` (sampled split events with stacks: `pmd` =
   `__split_huge_pmd`, the public entry the `split_huge_pmd()` macro calls
   for a huge PMD — its static worker `__split_huge_pmd_locked` is inlined
   in some builds, which the VM test showed; `page` = the folio split entry
   with its result), and a per-process `AnonHugePages` sample (`memory_rss`
   member -6 `anon_huge`) at the end of the capture.

Why now: a device-unmap (model unload) regression between two OS images on
otherwise identical hosts, where the slower image shows a ~3× smaller
huge-page-backed share of the same pinned arena and a higher THP
fault-fallback rate. Whether the unpin/unmap walk is 512× longer because the
arena is 4 KiB-backed is exactly the `memory_iommu` read.

## How it degrades

Nothing refuses a capture. The symbols are probed once in `/proc/kallsyms`
(module symbols appear there while `vfio_iommu_type1` is loaded) and the
iommu tracepoint directory in tracefs, BEFORE the programs are selected for
load — a kprobe on a missing symbol would otherwise fail `skel.attach()`
and the whole capture. A missing leg is left unloaded and named in
`sysinfo.memory_vfio_leg` / `memory_thp_leg` (`on`, `on:pmd-only`,
`on:page-only`, `off:nosym`, `off:notracepoint`). The folio-split entry is
`split_huge_page_to_list_to_order` on 6.9+ and `split_huge_page_to_list` on
the 6.6 series; both programs exist and the one the kernel has is loaded.
Verified at the sources: the iommu tracepoints and the vfio/THP symbol
names are the same at v6.6, v6.12 and v6.18 (and `__folio_split` is where
6.18 counts `thp_split_page`, under the same public entry). Blast radius
of a verifier surprise on some fleet kernel: none of the nine new programs
is autoloaded unless `--memory-vfio` / `--memory-thp-sample-rate` is
given, so a rejection could only fail an opt-in run, never a default
capture (the skeleton's other programs are untouched).

## Cost

- `memory_vfio`: one ring event per DMA region (a model load is tens to
  thousands), stacks included. Unsampled.
- `memory_iommu`: NO ring traffic. A 146 GiB region backed by 4 KiB pages
  is ~38 M `iommu:map` events; each is one preallocated-HASH update
  (`(tgid, op, iova GiB, log2 size)` → `{count, bytes}`, 128 K entries,
  created only when the leg is on) in the mapping ioctl's own task
  context, drained once at the end. Classic tracepoint programs, never
  the perf_event sampler → no NMI writer on the map; no LRU_HASH anywhere.
- **The host-wide hot path (the #126 class):** `iommu:map` / `iommu:unmap`
  fire for EVERY `iommu_map()` on the host — on an IOMMU-translated host
  that is the DMA-API path of every NIC / NVMe driver, softirq/IRQ context
  included. Bounded three ways: (1) the tracepoint program's first act is
  ONE lookup in a small preallocated hash (`memory_vfio_inflight`, keyed
  by tgidpid) that only the `vfio_dma_do_map` / `_unmap` kprobes populate —
  for every task not inside a VFIO map/unmap ioctl, and for every
  IRQ-context map whose iova is outside the region that task is mapping,
  the program returns on that miss (tens of ns, no task deref chain, no
  histogram touch); the kretprobes close the window, so the histogram is
  VFIO's by construction. (2) Everything is preallocated: no allocation in
  the hot path; a full histogram counts the drop (`memory_iommu_overflow`,
  printed at the end of the capture) rather than stalling. (3) Scope:
  on-demand only, `--memory-vfio` opt-in, off by default, first contact on
  one named host; it is not a continuous leg and gets no agent knob until
  that read says what the lookup costs on a busy NIC host.
- `memory_thp`: 1 in N (`--memory-thp-sample-rate`, 0 = off), task
  context, kprobe + kretprobe.
- `memory_vmstat`: two `/proc/vmstat` reads. `anon_huge`: one
  `smaps_rollup` read per process that produced a memory event, at the
  end.

Everything is off by default: `--memory-vfio` and
`--memory-thp-sample-rate N` opt in; `memory_vmstat` rides the memory
recorder itself (two file reads).

## Design choices

1. **Histogram key** — per process AND per GiB of IOVA, so the rows say
   where the fragmentation sits, not only how much: `memory_iommu` is
   keyed `(process, op, iova >> 30, log2 size)`; the BPF map holds 131072
   preallocated entries (a 150 GiB arena touches ~150 IOVA GiBs × 19 size
   orders × 2 ops ≈ 5.7K entries, so ≈23 such processes per capture);
   runs that find it full are dropped and counted, so the table is a floor
   in that case.
2. **Per-event THP splits, sampled, with stacks** — to find who splits
   huge pages, which the vmstat counters cannot say. `memory_thp`, 1 in N
   (`--memory-thp-sample-rate`, 0 = off).
3. **Fragmentation unit = the IOMMU run size** from `iommu:map` /
   `iommu:unmap` (not the pinned-folio order from a
   `vfio_pin_pages_remote` kretprobe: a static, inlinable symbol, and the
   tracepoints also cover iommufd callers); `memory_vfio` (the region)
   stays on the `vfio_dma_do_*` kprobes because nothing else carries the
   ioctl arguments.
4. **Per-process huge-page share from `smaps_rollup`** — `memory_rss`
   member -6 `anon_huge`, one read per process that produced a memory
   event, at the end of the capture, beside its `anon` rows; the host-wide
   `nr_anon_transparent_hugepages` level rides `memory_vmstat` too.

## Schema

Schema 17 → 18 (`[bump minor]`): tables `memory_vfio`, `memory_iommu`,
`memory_thp`, `memory_vmstat`; `memory_rss` member -6 `anon_huge`;
`sysinfo` columns `memory_vfio_leg`, `memory_thp_leg`,
`memory_thp_sample_rate`. `SCHEMA_CHANGES.md` has the column-by-column
read rules and the three queries behind the reads.

## Verification

- Gates on this head: `cargo fmt --check`, `clippy --all-targets -D
  warnings`, `cargo test --lib` (parsers, schema round-trips, program
  selection) — PASS.
- VM test (v6.12 guest): `tests/memory_recorder.rs` all pass incl. the new
  `test_memory_thp_vmstat_e2e` — an madvise(MADV_HUGEPAGE) + partial
  MADV_DONTNEED workload: `memory_vmstat` present and monotonic,
  `memory_thp` pmd rows with kernel stacks under the host-wide
  `thp_split_pmd` delta, `sysinfo` legs named, zero utid orphans;
  `memory_vfio_leg = off:nosym` with the capture intact (no VFIO device in
  the guest). The first run of that test is what found the inlined
  `__split_huge_pmd_locked`.
- NOT exercised here: the VFIO/IOMMU ON path on a real device — the guest
  has none. First contact is an on-demand memory capture on one host of
  each OS image during a model load/unload (`--add-recorder memory
  --memory-vfio --memory-thp-sample-rate 1`); the three queries are in
  SCHEMA_CHANGES.md.

Checked against the known hazard classes: NMI-context map writers (none
here), probes on host-wide hot paths (the iommu tracepoints — bounded as
above), the brk guard legs (untouched), the degrade shape of #217 (reused).

Co-authored-by: Claude <noreply@anthropic.com>


josefbacik:
rust-bpf-code-reviewer pass — review of the #556 head (branch `tracer-mem/memory-vfio-thp`, base v1.14.0 `0cb69e38`; sha in the PR), 19 files

Reviewer checklist: `.claude/agents/rust-bpf-code-reviewer.md` (blob `3739c8b3`), applied pass by pass to the full diff. Gates on this sha: `cargo fmt --check`, `clippy --all-targets -D warnings`, lib tests 561/0; VM test (v6.12 guest) `memory_recorder` 8/8 incl. the new `test_memory_thp_vmstat_e2e`.

### Summary
A new, opt-in instrument set on the memory recorder: kprobe/kretprobe pairs on the vfio_iommu_type1 ioctl handlers (per-region events, and the window the histogram counts in), two classic tracepoint programs counting `iommu:map`/`unmap` run sizes into a preallocated hash (no ring traffic), sampled THP-split kprobes with stacks, a start/end `/proc/vmstat` sample, and a per-process `AnonHugePages` read — four tables, one `memory_rss` member, three `sysinfo` columns, schema 17 → 18. The code follows the existing memory-leg shapes (reserve/emit helpers, per-type header field reuse, `utid` everywhere, degrade-not-refuse via a kernel probe before program selection). No critical issue found.

### Critical Issues
None.

| Severity | Finding | Where | Disposition |
|---|---|---|---|
| Critical | — | — | none |
| High | — | — | none |
| Medium | `memory_vfio` rows are the request at handler ENTRY: an ioctl the kernel refuses still produces a row, and `VFIO_DMA_UNMAP_FLAG_ALL` carries iova/size 0. Undocumented in the first draft. | `systing_system.bpf.c` `memory_vfio_emit`; SCHEMA_CHANGES v18 | FIXED (docs: the row is the region as requested; the ALL flag's meaning stated). A kretprobe result column is a follow-up if a consumer needs the outcome. |
| Medium | Histogram bound: 131072 entries, `(tgid, op, iova GiB, log2 size)`; a full map drops updates silently from the table's point of view. | `memory_iommu_hist`, `memory_iommu_account` | Addressed: drops are counted per CPU (`memory_iommu_overflow`) and printed at capture end (`memory-vfio: N iommu map/unmap runs not counted`), SCHEMA_CHANGES says the table is then a floor; the bound (≈23 processes × 150 GiB arenas) is in the map comment and the PR body. Not in `sysinfo` — a follow-up if first contact ever hits it. |
| Medium | **Host-wide hot path (the #126 class).** `iommu:map`/`unmap` fire for every `iommu_map()` on the host (NIC/NVMe DMA-API on an IOMMU-translated host, softirq/IRQ context included). The first draft gated on `trace_task(current)`, which in an unfiltered capture admits every task — so every DMA-API map on the host would have taken the histogram's bucket lock. | `memory_iommu_account` | FIXED: the vfio kprobes open a per-task window (`memory_vfio_inflight`, tgidpid → the region being mapped/unmapped; the kretprobes close it) and the tracepoint program's first act is that one small-hash lookup — a miss for every non-VFIO task and every out-of-range IRQ-context map, tens of ns, no task deref chain, nothing else touched; only runs inside the region are counted, so the histogram is VFIO's by construction. Everything preallocated (no allocation in the hot path); a full map counts, never stalls. Scope: on-demand, `--memory-vfio` opt-in, off by default, first contact on one named host, no continuous knob. Residual: a kretprobe miss (maxactive) leaves a window open until the task's next vfio ioctl overwrites it; an IRQ on that task's CPU is counted only if its iova falls inside the region. |
| Low | PID reuse inside one capture: `emit_anon_huge_pages` reads `/proc/<tgid>/smaps_rollup` for every tgid seen and attributes it to `get_or_create_utid(tgid)`; a tgid recycled within the capture window would be read as the new process. | `memory_recorder.rs` | Accepted: same exposure as every end-of-capture `/proc` read in the tree; window is one capture (10–30 s). Noted. |
| Low | The PMD-split probe is `__split_huge_pmd` (public entry), so rmap-path PMD splits (`split_huge_pmd_locked` from `try_to_unmap`/`try_to_migrate`, 6.10+) are in the vmstat delta but not in `memory_thp`. | BPF comment, SCHEMA_CHANGES | Documented; attribution of reclaim/migration splits is a follow-up (a second kprobe would double count on 6.18 where `__split_huge_pmd` calls `split_huge_pmd_locked`). |
| Low | `vfio_dma_do_map`/`unmap` are static module symbols: the kallsyms probe covers "module not loaded" and "symbol inlined"; the iommufd back end (`iommufd_ioas_map`) has no per-region events — only the histogram would see it, and only inside a type1 window. | `probe_memory_kernel_legs` | Documented in the PR body; iommufd support is a follow-up when a host shows that back end. |

### Rust quality
- Idiomatic: `let … else` for the optional collector and probe reads, `then_some`/`then` for the sysinfo options, `filter_map` in `parse_vmstat`; no new `if let` chains that want a `match`.
- `unsafe`: exactly two new lines, `unsafe impl Plain for memory_iommu_key {}` / `memory_iommu_value {}` — the same pattern as every skeleton type in `systing_core.rs` (POD structs generated from the BPF object, read back with `plain::copy_from_bytes`, which checks the length). No other `unsafe`.
- Comments explain why (run size = contiguity; why the tracepoints count instead of emit; why the window gate; why the kernel probe runs before program selection; why `__split_huge_pmd` and not the `_locked` worker). No source-line references. The per-type header field reuse is called out at each emit site as the existing legs do.
- Error handling: every `/proc` read degrades to "absent" (`Vec::new()` / `None`); the drain skips unreadable entries; no `unwrap` on external data. `report_write_error` reused for every new collector write.
- Overengineering check: no new abstractions beyond `MemoryKernelLegs` (a three-field probe result with two string formatters) and `MemoryIommuHistEntry` (the drained row). Both are the minimum that keeps the selection / sysinfo / drain readable.

### BPF quality
- Ringbuf payload: `memory_vfio`/`memory_thp` reuse `memory_event` unchanged (header fields `addr/size/old_addr/member/flags` by type; no new struct, no new bytes on the wire). The histogram emits nothing to the ring.
- Map operations: the histogram path is one small-hash lookup (the window) and, inside a window, one lookup, one conditional `update(BPF_NOEXIST)` + re-lookup on a new key, and two atomic adds per `iommu_map()`; `memory_log2_u64` is a six-step straight-line binary search (no clz). The sampling counter is the fault leg's PERCPU_ARRAY pattern.
- Map/context pairing: the new maps' writers are `tracepoint/iommu/map|unmap` (classic tracepoint), `kprobe|kretprobe/vfio_dma_do_*` (the window), `kprobe/__split_huge_pmd`, `kretprobe/split_huge_page_to_list*` — task context; `bpf_prog_active` is raised around every one of them; the `perf_event` sampler does not touch them. Both `BPF_MAP_TYPE_HASH` maps are preallocated (default). No `LRU_HASH` anywhere under `src/bpf/` (grep empty).
- Verifier: the classic iommu tracepoint ctx structs mirror the fixed event layout (8-byte common header + u64 fields in `TP_STRUCT__entry` order, unchanged since the tracepoints were added; verified at v6.6, v6.12, v6.18 trees). The kprobe arg reads use `bpf_probe_read_kernel` on the handler's kernel pointer. Loaded and attached in the v6.12 guest (8/8).
- Checked against the `R0 unbounded memory access` class that a newer 6.12-series verifier raises on `systing_rss_stat_btf` (a map-value store through a register index whose bound the verifier lost across helper calls — `handle_rss_stat`'s `st->latest_seen[member]` / `st->last_emitted[member]`, fixed separately): none of the nine new programs stores into or loads from a map value, a ctx, or a BTF pointer through a runtime-computed index. The histogram value is two fixed-offset counters updated by `__sync_fetch_and_add`; the window map is looked up and updated whole (`memory_vfio_inflight` read as two fixed fields, `iova` / `size`); the key fields (`iova_gib`, `order`) are computed INTO the key struct, never used as an index; the tracepoint ctx fields are read at constant offsets; the vfio argument structs are copied whole with a constant-size `bpf_probe_read_kernel`; the sampled-split and vfio events are written through the existing header helpers at fixed offsets. The only runtime-indexed store in the memory legs stays `handle_rss_stat`. The v6.12.0 guest load is not evidence against that class (it is the verifier version that does not reproduce the rss_stat rejection) — the absence of any indexed map access is.
- Kernel-version handling is by symbol presence (kallsyms for the static/module functions, tracefs for the tracepoint dir), never by `uname`; the two folio-split entry names are separate programs, exactly one loaded.

### Schema & versioning
- `SCHEMA_CHANGES.md` v18 entry: present, column by column, with the read queries.
- `SCHEMA_VERSION` 17 → 18: present (`src/duckdb.rs`).
- `Cargo.toml` / `Cargo.lock`: untouched.
- `[bump minor]` in the PR title: present.
- DuckDB DDL, parquet schemas, import/export lists, `parquet_paths` and the stream table list all carry the four tables; the pre-1.15 parquet import test (sysinfo by-name) extended to the three new columns.

### Recorder consistency
- `utid` on every new table; `memory_iommu.utid` is the mapping process's main thread via `get_or_create_utid(tgid)` (tgid == the leader's tid), the same resolution the rss sampler uses. No raw pid column.
- Process attribution via `thread.upid` → `process` (the SCHEMA_CHANGES queries do exactly that).
- The e2e asserts zero utid orphans across `memory_thp`, `memory_vfio`, `memory_iommu`.

### Tests
- Unit: program selection follows the probe result in every combination (all / legacy entry / pmd-only / page-only / none / not asked), the sysinfo leg strings, the kallsyms probe against real symbols, `parse_vmstat` (order, filtering, bad lines), `parse_smaps_anon_huge`, sysinfo round-trip with the three new columns, pre-1.15 import.
- VM test: `test_memory_thp_vmstat_e2e` — madvise(MADV_HUGEPAGE) + partial MADV_DONTNEED: vmstat present and monotonic (21 counters), THP pmd rows with kernel stacks bounded by the host-wide `thp_split_pmd` delta, legs named in sysinfo, VFIO tables empty with `off:nosym` in a guest without vfio.
- NOT exercised: the VFIO/IOMMU ON path on a real device (no VFIO in the guest) — the window gate and the histogram's in-range test are unexercised code until first contact on a host with a VFIO device; the PR body says so.

### Positive notes
- The kernel probe before program selection is the right place for degrade: a missing module symbol never reaches `skel.attach()`, and the histogram map is not even created when its leg is off.
- The histogram's design note (why counting beats emitting, with the 38M-event arithmetic), the window gate's note (why the tracepoints must not trust `current` alone) and the `__split_huge_pmd` note (found by the VM test, not by reading) are the kind of comment a future reader needs.

### Verdict
**SHIP** — Critical: none; High: none; Medium: 3 (all addressed in the reviewed sha: entry-time row semantics documented, overflow accounting, the host-wide hot path gated on the VFIO window); Low: 3 (documented, follow-ups named). Verified on this sha: gates green, VM test 8/8, pairing check clean, no version-file touch, schema bump complete.

PR #220, opened 2026-08-25
bpf: re-bound the rss_stat member index at the use [bump patch]

josefbacik:
**What this fixes.** With the memory recorder enabled, the BPF object fails to load on newer 6.12-series kernels (and 6.18): the verifier rejects `systing_rss_stat_btf` with `R0 unbounded memory access, make sure to bounds check any such access` at the first `st->latest_seen[member]` store in `handle_rss_stat()`, and libbpf then fails the whole skeleton (`-13`), so every capture that includes the memory recorder exits 1 at start-up. Captures without the memory recorder are unaffected (the program is not autoloaded then).

**Why it happens.** `handle_rss_stat()` range-checks `member` on entry, then makes two map-helper calls (the per-tgid state lookup and the `BPF_NOEXIST` insert on a miss) before indexing `last_emitted[]` / `latest_seen[]`. The program's own code did not change since it last loaded cleanly; the surrounding object did (per-CPU ring sharding, the cgroup-filter split), which shifted register allocation so that the index at the store is rebuilt from the sign-extended tracepoint argument and the newer verifier no longer carries the entry bound across the helper calls.

**The change.** One helper, both attach paths (`tp_btf/rss_stat` and the classic `tracepoint/kmem/rss_stat` share it): mask the index at the point of use — `idx = member & (NR_MM_COUNTERS - 1)` — with a `_Static_assert` that `NR_MM_COUNTERS` is a power of two (it is 4), and `barrier_var(idx)` so the compiler keeps the masked value in a register at the access. The entry check stays (it still short-circuits out-of-range members); the mask is the bound the verifier sees on every kernel. No behaviour change for in-range members; no new maps, programs, or options.

**Verification.**
- Verifier log excerpt from an affected 6.12-series host (the failing load of the unfixed object): in the review verdict comment.
- `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo test` on the fixed tree: results in the verdict.
- Load check on the project's 6.12 VM guest: the `memory_recorder` e2e target (both rss_stat attach paths) on the fixed object — the guest's older verifier already accepted the unfixed object, so this is the "still loads where it loaded" check, not the reproduction.
- The fix is confirmed on an affected 6.12-series host once the fixed build loads the object with the memory recorder enabled.

`[bump patch]`

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -1078,6 +1078,12 @@ struct rss_stat_state {
 	s64 latest_seen[NR_MM_COUNTERS];
 };
 
+/* handle_rss_stat() re-bounds its member index with a mask at the point of
+ * use (see the comment there); a mask is only an exact bound when the array
+ * length is a power of two. */
+_Static_assert((NR_MM_COUNTERS & (NR_MM_COUNTERS - 1)) == 0,
+	       "NR_MM_COUNTERS must be a power of two for the rss_stat index mask");
+
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	/* Bounds concurrent TRACED processes (tgids), not threads or events.
@@ -5367,11 +5373,25 @@ static __always_inline int handle_rss_stat(struct task_struct *task,
 			return emit_rss_stat_event(task, member, size_bytes, rss_flags);
 	}
 
+	/* Re-bound the index AT THE USE, not only at the entry check above.
+	 * Two map-helper calls separate that check from these stores, and a
+	 * newer 6.12-series verifier (and 6.18) no longer carries the bound
+	 * across them once the compiler rebuilds the index from the
+	 * sign-extended tracepoint argument: it rejects the whole object with
+	 * "R0 unbounded memory access, make sure to bounds check any such
+	 * access" at the first st->...[member] store, and the memory recorder
+	 * cannot load at all. NR_MM_COUNTERS is a power of two (asserted at
+	 * the struct), so the mask is an exact bound the verifier sees on
+	 * every kernel; barrier_var() pins the masked value in a register at
+	 * the access so the compiler cannot fold it back onto the argument. */
+	u64 idx = member & (NR_MM_COUNTERS - 1);
+	barrier_var(idx);
+
 	/* Unsynchronized per-member s64 stores — multiple threads of a tgid
 	 * may race here. Benign: worst case is a duplicate emit or one stale
 	 * last_emitted (one extra emit); never a lost sample. */
-	st->latest_seen[member] = size_bytes;
-	s64 last = st->last_emitted[member];
+	st->latest_seen[idx] = size_bytes;
+	s64 last = st->last_emitted[idx];
 	/* First sighting for this member: always emit so consumers have a
 	 * baseline (last == -1 from the 0xff memset). */
 	s64 drift = last < 0 ? (s64)tool_config.memory_rss_threshold_bytes
@@ -5381,7 +5401,7 @@ static __always_inline int handle_rss_stat(struct task_struct *task,
 	if ((u64)drift < tool_config.memory_rss_threshold_bytes)
 		return 0;
 
-	st->last_emitted[member] = size_bytes;
+	st->last_emitted[idx] = size_bytes;
 	return emit_rss_stat_event(task, member, size_bytes, rss_flags);
 }
 


josefbacik:
**Review verdict — bpf: re-bound the rss_stat member index at the use**

**Scope checked.** One hunk in `src/bpf/systing_system.bpf.c` (`handle_rss_stat()`: `idx = member & (NR_MM_COUNTERS - 1)` + `barrier_var(idx)` used for the three `st->…[]` accesses; entry check kept) plus a `_Static_assert` that `NR_MM_COUNTERS` is a power of two beside `struct rss_stat_state`. No Rust changes, no map/program/option changes; both attach paths (`tp_btf/rss_stat`, `tracepoint/kmem/rss_stat`) share the helper.

**The failure being fixed (verbatim from an affected 6.12-series host running the unfixed object with the memory recorder enabled; the capture exits 1 at skeleton load, every time):**

```
… R0 unbounded memory access, make sure to bounds check any such access
processed 413 insns (limit 1000000) max_states_per_insn 1 total_states 36 peak_states 36 mark_read 14
-- END PROG LOAD LOG --
libbpf: prog 'systing_rss_stat_btf': failed to load: -13
libbpf: failed to load object 'systing_system_bpf'
libbpf: failed to load BPF skeleton 'systing_system_bpf': -13
Error: Failed to load BPF skeleton into kernel. Check dmesg for BPF verifier errors.
Caused by:
    Permission denied (os error 13)
```

Observed on 6.12.85-, 6.12.68-, 6.12.55-series and 6.18-series hosts alike (every affected host we have seen); the same object loads on a 6.12.0 guest, and the program's own source is unchanged from the last release that loaded cleanly there — the verifier-version sensitivity, not a kernel-specific ABI change.

**Gates on the fixed tree (identical bytes to the commit):**
- `cargo fmt --all -- --check`: exit 0
- `cargo clippy --all-targets` (jobs 4): exit 0, no code warnings (6m37s)
- `cargo test` (jobs 2): exit 0 — 557 + 11 + 28 passed, 0 failed (the VM-only integration targets `#[ignore]`d, run in the guest below)
- `cargo build --release` (jobs 4): exit 0 (11m08s) — the BPF object compiles with the mask; the `_Static_assert` holds on the vmlinux.h `NR_MM_COUNTERS`

**Load check.** VM guest (6.12.0, `bzImage-v6.12-bpf`), `memory_recorder` e2e target (both rss_stat attach paths) on the fixed object: **PASS — 7 passed, 0 failed** (`test_memory_recorder_e2e` and siblings; 738 s in the guest) — the "still loads where it loaded" check. The fix is confirmed on an affected 6.12-series host once the fixed build loads the object with the memory recorder enabled.

---

rust-bpf-code-reviewer pass — `handle_rss_stat` index re-bound (branch `tracer-oom/rss-stat-index-bound`, one hunk in `src/bpf/systing_system.bpf.c`, +23/−3 on v1.14.0 `0cb69e38`)

### Summary
The memory recorder's rss_stat programs fail to load on newer 6.12-series (and 6.18) verifiers: `handle_rss_stat` checks `member >= NR_MM_COUNTERS` at its entry, then two map-helper calls later stores through `st->latest_seen[member]` / `st->last_emitted[member]` on the map value, and the verifier no longer carries the entry bound to the store ("R0 unbounded memory access"). The fix re-bounds the index at the use — `u64 idx = member & (NR_MM_COUNTERS - 1); barrier_var(idx);` — and indexes all three accesses (two stores, one load) with `idx`, with a `_Static_assert` that the mask is exact (NR_MM_COUNTERS is a power of two). Both attach paths (`tp_btf/rss_stat` and the classic `tracepoint/kmem/rss_stat`) share `handle_rss_stat`, so one hunk covers both. Semantics unchanged: the entry check still returns for out-of-range members, so `idx == member` on every path that reaches the stores; `emit_rss_stat_event` still receives the original `member`.

### Critical Issues
None.

| Severity | Finding | Disposition |
|---|---|---|
| Critical | — | none |
| High | — | none |
| Medium | The mask is the right cure only because `NR_MM_COUNTERS` is 4 on every kernel the tree targets (MM_FILEPAGES, MM_ANONPAGES, MM_SWAPENTS, MM_SHMEMPAGES at 6.6, 6.12, 6.18); a future kernel adding a fifth counter would make the mask inexact. | Addressed by the `_Static_assert((NR_MM_COUNTERS & (NR_MM_COUNTERS - 1)) == 0)` against the vmlinux.h enum — a non-power-of-two count fails the build instead of mis-indexing. |
| Low | `barrier_var(idx)` (`asm volatile("" : "+r"(var))`, bpf_helpers.h) hides the masked value from the COMPILER so it cannot rebuild the index from the sign-extended argument; the verifier still sees the `AND` and carries `umax = NR_MM_COUNTERS - 1` into the `*8` offset (≤ 24 bytes into a 32-byte array) — the bound is on the instruction stream, which is what the verifier reads. | Noted; this is the intended mechanism. |
| Low | The only other runtime-indexed code in the memory legs, `read_mm_rss_pages`, indexes with compile-time constants through a `switch` — not affected. No other program in the object stores through a runtime index (checked the nine programs of the pending VFIO/THP change as well). | Noted. |

### BPF quality
- One hunk, no new maps, no new programs, no ringbuf payload change; the helper-call sequence around the check is untouched (the fix is the re-bound, not a restructure).
- The comment at the use names the verifier class and why the entry check alone stopped being enough; no source-line references.

### Schema & versioning
- No schema change; `Cargo.toml` / `Cargo.lock` untouched; `[bump patch]` in the PR title is right for a BPF-only fix.

### Evidence required on the PR before the verdict line is "verified"
1. The rejecting verifier's own log block (the `R0 unbounded memory access … processed 413 insns … END PROG LOAD LOG` excerpt for `systing_rss_stat_btf`) quoted in the verdict — present above.
2. The object still loads where it loaded before: the `memory_recorder` integration target in the v6.12 VM guest (7 tests; the only in-repo exercise of both rss_stat attach paths) green on the fixed object — the "still loads" half; the rejecting kernel class is confirmed on an affected 6.12-series host once the fixed build loads the object with the memory recorder enabled, not by this guest.
3. `cargo fmt --check`, `clippy --all-targets -D warnings`, lib tests on the pushed sha.

### Verdict
**SHIP** (on the evidence above) — Critical: none; High: none; Medium: 1 (addressed by the static assert); Low: 2 (noted). The cure matches the ruled shape (re-bound at the use, both attach paths).

PR #221, opened 2026-08-25
bpf: put the barrier before the rss_stat index mask so the compiler keeps it [bump patch]

josefbacik:
**What this fixes.** #220 did not fix the verifier rejection it targeted: in the release build its mask is optimized away, and the compiled `tp_btf/rss_stat` program of v1.15.1 is instruction-identical to v1.14.0's. With the memory recorder enabled the object is still rejected at skeleton load on the affected hosts (`R0 unbounded memory access, make sure to bounds check any such access` in `systing_rss_stat_btf`; libbpf fails the skeleton with `-13`; every capture that includes the memory recorder exits 1 at start-up). Captures without the memory recorder are unaffected (the program is not autoloaded then).

**Why #220 was a no-op.** `handle_rss_stat()` checks `member >= NR_MM_COUNTERS` on entry, so at the store site the optimizer knows `member < 4`, proves `member & (NR_MM_COUNTERS - 1) == member`, and deletes the mask — before `barrier_var()` runs, so the barrier pinned the unmasked value. The verifier then sees what it saw before: the entry check on a 32-bit copy of the argument (`w2 = w8`), and the index rebuilt from the zero-extended raw argument (`r8 <<= 32; r8 >>= 32; r8 <<= 3; r0 += r8; *(u64 *)(r0 + 0x20) = r7`) two map-helper calls later.

**Two variables, both real.** Which register the compare sits on is the compiler's choice: clang 21 at its default `-mcpu` (v3) — the release build — checks the copy and indexes with the other register; clang 18 at `-mcpu=v1` compares the index register itself. Whether the bound on the copy reaches the index register is the verifier's choice: a vanilla 6.12.0 propagates it (the release object loads there), the newer 6.12.y stable and 6.18 verifiers on the affected hosts do not. That is why the project's local build and its 6.12 test VM passed in #220's verification while the release object failed in the field, and why #220's "newer verifier" explanation was only half of it.

**The change.** Barrier first, then mask: `u64 idx = member; barrier_var(idx); idx &= (NR_MM_COUNTERS - 1);`. `barrier_var()` makes `idx` opaque, so the mask that follows cannot be folded and lands on the register the access uses. Compiled with clang 21 at the build's flags (`-g -O2 -target bpf -D__TARGET_ARCH_x86`, default `-mcpu`), the program gains exactly one instruction over v1.15.1 — `r8 &= 0x3` between the zero-extension and the index shift — and the verifier carries umax 3 into the three `st->…[idx]` accesses with no reliance on copy propagation, on every verifier. The entry check stays; the `_Static_assert` now also requires `NR_MM_COUNTERS` to be non-zero. Both attach paths (`tp_btf/rss_stat` and the classic `tracepoint/kmem/rss_stat`) share the helper. No Rust changes, no new maps/programs/options.

**Verification.**
- Disassembly diff (clang 21, build flags): v1.15.1 vs this change on `tp_btf/rss_stat` = one added instruction, `r8 &= 0x3`, at the index build; the store site before/after is quoted in the review comment.
- Load check in the project's 6.12 VM (a minimal loader autoloading only `systing_rss_stat_btf`): this change's object loads — and so does the unfixed release object, because that verifier propagates the bound; the VM is the "still loads where it loaded" check, not the reproduction. The reproduction class (the affected hosts' verifiers) is confirmed once the fixed release build loads the object with the memory recorder enabled there.
- `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo test` on the tree: results in the review comment.

**Follow-up, not in this change:** pin `-mcpu` in `build.rs` so the release build, CI and the test VM compile the same ISA and a verifier difference cannot hide behind a compiler default again.

`[bump patch]`

### DIFF
--- a/src/bpf/systing_system.bpf.c
+++ b/src/bpf/systing_system.bpf.c
@@ -1189,8 +1189,9 @@ struct rss_stat_state {
 
 /* handle_rss_stat() re-bounds its member index with a mask at the point of
  * use (see the comment there); a mask is only an exact bound when the array
- * length is a power of two. */
-_Static_assert((NR_MM_COUNTERS & (NR_MM_COUNTERS - 1)) == 0,
+ * length is a non-zero power of two. */
+_Static_assert(NR_MM_COUNTERS > 0 &&
+		       (NR_MM_COUNTERS & (NR_MM_COUNTERS - 1)) == 0,
 	       "NR_MM_COUNTERS must be a power of two for the rss_stat index mask");
 
 struct {
@@ -5482,19 +5483,30 @@ static __always_inline int handle_rss_stat(struct task_struct *task,
 			return emit_rss_stat_event(task, member, size_bytes, rss_flags);
 	}
 
-	/* Re-bound the index AT THE USE, not only at the entry check above.
-	 * Two map-helper calls separate that check from these stores, and a
-	 * newer 6.12-series verifier (and 6.18) no longer carries the bound
-	 * across them once the compiler rebuilds the index from the
-	 * sign-extended tracepoint argument: it rejects the whole object with
-	 * "R0 unbounded memory access, make sure to bounds check any such
-	 * access" at the first st->...[member] store, and the memory recorder
-	 * cannot load at all. NR_MM_COUNTERS is a power of two (asserted at
-	 * the struct), so the mask is an exact bound the verifier sees on
-	 * every kernel; barrier_var() pins the masked value in a register at
-	 * the access so the compiler cannot fold it back onto the argument. */
-	u64 idx = member & (NR_MM_COUNTERS - 1);
+	/* Re-bound the index AT THE USE, on the register the access uses.
+	 *
+	 * The entry check above is what the compiler compares, but which
+	 * register it compares is its choice: clang 21 at the default -mcpu
+	 * (v3) checks a 32-bit copy of the argument and later rebuilds the
+	 * index from the zero-extended raw argument in another register, two
+	 * map-helper calls later; clang 18 at -mcpu=v1 compares the index
+	 * register itself. Whether a bound on the copy reaches the index
+	 * register is then the verifier's choice: a vanilla 6.12.0 propagates
+	 * it, the newer 6.12-stable and 6.18 verifiers on the affected hosts
+	 * do not and reject the whole object at the first st->...[] store
+	 * ("R0 unbounded memory access, make sure to bounds check any such
+	 * access"). An explicit mask on the index register itself is bounded
+	 * on every verifier and under every compiler.
+	 *
+	 * The barrier comes FIRST. With the entry check dominating, the
+	 * optimizer proves `member & (NR_MM_COUNTERS - 1) == member` and
+	 * deletes the mask; barrier_var() makes `idx` opaque, so the mask that
+	 * follows it cannot be folded and lands on the very register used as
+	 * the index. NR_MM_COUNTERS is a power of two (asserted at the struct),
+	 * so the mask is an exact bound (umax NR_MM_COUNTERS - 1). */
+	u64 idx = member;
 	barrier_var(idx);
+	idx &= (NR_MM_COUNTERS - 1);
 
 	/* Unsynchronized per-member s64 stores — multiple threads of a tgid
 	 * may race here. Benign: worst case is a duplicate emit or one stale


josefbacik:
**Review verdict — bpf: put the barrier before the rss_stat index mask so the compiler keeps it**

**Scope checked.** One hunk in `src/bpf/systing_system.bpf.c` (`handle_rss_stat()`: `u64 idx = member; barrier_var(idx); idx &= (NR_MM_COUNTERS - 1);` — barrier first, then the mask; the three `st->…[idx]` accesses unchanged from #220; the comment rewritten to the compiler + verifier facts) plus the `_Static_assert` beside `struct rss_stat_state` now requiring `NR_MM_COUNTERS > 0`. No Rust changes; both attach paths (`tp_btf/rss_stat`, `tracepoint/kmem/rss_stat`) share the helper.

**Why #220 was a no-op — read at the compiled objects, not the source.** The BPF ELF carved from the release binaries of v1.13.16, v1.14.0 and v1.15.1 (the libbpf-rs skeleton embeds it in `.rodata`) has a 365-instruction `tp_btf/rss_stat` in all three; v1.14.0 vs v1.15.1 differ only in five `.rodata` load offsets (new config fields), not one added instruction. The store site in v1.15.1:

```
249: r8 <<= 0x20
250: r8 >>= 0x20
251: r8 <<= 0x3
252: r0 += r8
253: *(u64 *)(r0 + 0x20) = r7
```

The only `&=` in the program are `w1 &= 0x200000` (PF_KTHREAD) and `w0 &= 0x1`; the entry check `member >= NR_MM_COUNTERS` is compiled onto a 32-bit copy (`w2 = w8`), never onto `r8`. The entry check dominates the store, so the optimizer folded `member & 3` to `member` before `barrier_var()` ran.

**Two variables, both observed.** (1) Compiler: the same source compiled with clang 21.1.8 at the build's flags (`-g -O2 -target bpf -c -D__TARGET_ARCH_x86`, default `-mcpu` = v3) reproduces the release object's instruction stream for this program (the only differences are subprogram `call` targets from the libbpf linker's placement) — two independent compiles agree; clang 18 at `-mcpu=v1` compares the index register itself. (2) Verifier: in the project's 6.12.0 test VM, with a minimal loader that autoloads only `systing_rss_stat_btf` (81 other programs disabled), the release object, the clang-21 recompile of the v1.15.1 source, AND the fixed object all load (`LOAD-OK` ×3, loader rc 0 each) — that verifier propagates the copy's bound to the index register; the newer 6.12.y stable (6.12.55 and later) and 6.18 verifiers on the affected hosts do not, and reject the release object with `R0 unbounded memory access`. The VM is therefore the "still loads where it loaded" check only; the affected-host class is confirmed once the fixed release build loads the object with the memory recorder enabled there.

**The fix at the release compiler and flags.** Candidate A compiles to 366 instructions; the jump-target-insensitive diff against v1.15.1 is exactly one added instruction:

```
249: r8 <<= 0x20
250: r8 >>= 0x20
251: r8 &= 0x3        <- added
252: r8 <<= 0x3
253: r0 += r8
254: *(u64 *)(r0 + 0x20) = r7
```

The AND sits on the register the access uses (the `last_emitted[idx]` load and store reuse the same advanced `r0`), so the verifier carries umax 3 → a map-value offset ≤ 24 into all three accesses with no reliance on copy propagation — bounded under every verifier that rejects the unfixed shape and under the one that accepts it.

**Gates on the tree (identical bytes to the commit):**
- `cargo fmt --all -- --check`: exit 0 (before and after the commit)
- `cargo clippy --all-targets` (jobs 4): exit 0 twice (before and after the final comment edit), no code warnings
- `cargo test` (jobs 2): exit 0 — 561 + 28 + 11 + 7 + 4 + 1 + 1 passed, 0 failed (the VM-only integration targets `#[ignore]`d)

**Follow-up noted in the body:** pin `-mcpu` in `build.rs` so the release build, CI and the test VM compile the same ISA.

---

BPF reviewer pass (second reviewer): **SHIP.** Read the diff at the commit (`handle_rss_stat()`: barrier first, mask second, on the same `u64 idx` that indexes all three `st->…[idx]` accesses; the `_Static_assert` now refuses `NR_MM_COUNTERS == 0`; `emit_rss_stat_event()` still receives the original `member`; both attach paths share the helper; no map, program, payload or Rust change) and confirmed it at the compiled objects rather than the source: (1) the three code lines are byte-identical to the candidate I compiled with clang 21.1.8 at the build's flags and linked with the pystacks object the way `build.rs` does; that object's `tp_btf/rss_stat` differs from the v1.15.1 release program by exactly one instruction, `251: r8 &= 0x3`, on the register the access uses, between the zero-extension and `<<= 3; r0 += r8`; my clang-21 compile of the unfixed v1.15.1 source reproduces the carved release program instruction-for-instruction (365 insns, empty diff), so the compiler and flags are the release's. (2) Loaded in the project's 6.12.0 test VM with a loader that autoloads only `systing_rss_stat_btf`: the fixed object LOAD-OK — and so do the carved release object and the unfixed recompile, which is why that VM is the "still loads where it loaded" half and not the reproduction; the rejecting verifiers are the newer 6.12.y stable and 6.18 ones on the affected hosts, and the fixed release build loading there is the proof. (3) The bound does not depend on copy propagation: `AND 3` gives umax 3, `<<= 3` umax 24, the map-value offset is ≤ 24 into a 64-byte value on any verifier. (4) No other runtime-indexed map-value access exists in the object (the other `<<= k; r += ` sites are probe-read source addresses or constant-bound loops). Findings: Critical none / High none / Medium none / Low 1 — the mask relies on `NR_MM_COUNTERS` staying a power of two (the assert enforces it at compile time against the vendored header; a kernel with a fifth counter would still be dropped by the entry check, not mis-indexed). Follow-up agreed: fix `-mcpu` explicitly in `build.rs` so CI, the test VM and the release build compile the same ISA, and read every BPF verifier claim at the release object.

josefbacik:
**Correction to the record (the mechanism behind the rejection):** the wording above about verifier versions was wrong, and the rig data that produced it was a vacuous control. The per-member state path in `handle_rss_stat()` — the one with the indexed `st->latest_seen[…]` / `st->last_emitted[…]` accesses — is entered only when `tool_config.memory_rss_threshold_bytes != 0`. That field lives in `.rodata`; libbpf freezes `.rodata`, the verifier constant-folds the load and prunes the branch, so with the object's default (0) the stores are dead code at load and are never verified. That is why the unfixed object loaded on every kernel we tried in the test VM (a level-2 verifier log of those loads shows the store instructions were never visited). `systing` sets the threshold at runtime (`max(16 MiB, 64 * nr_cpus * page_size)`), which makes the path live.

Reproduction with the runtime shape (the `.rodata` map's initial value patched before `load()`: `memory_rss_threshold_bytes = 16 MiB`, `collect_memory = 1`, `page_size = 4096`), vanilla 6.12.85 guest:

- v1.15.1 object, `systing_rss_stat_btf`: **rejected** — `R0 unbounded memory access, make sure to bounds check any such access … processed 413 insns (limit 1000000) max_states_per_insn 1 total_states 36 peak_states 36 mark_read 14` (the same log, to the digit, as on the affected hosts).
- this PR's object (v1.15.2), same shape: **loads** (the explicit `r8 &= 0x3` on the index register).
- both objects with the default shape (threshold 0): load — the pruned-path control.

So with the object's default configuration the store path is pruned and every verifier accepts the unfixed object; with the configuration `systing` applies at runtime, a 6.12.85 verifier rejects v1.15.1 and loads v1.15.2. The fix stands as written (an explicit mask on the register the access uses is bounded under any verifier). The test gap is the one to close: the VM tests load the skeleton with the default `.rodata`, so any path gated by a runtime knob is unverified there — a follow-up adds a load test that iterates the runtime configuration shapes (and loads the release-flag build of the object), so a rejection like this one is caught before a tag.

PR #222, opened 2026-08-25
build: pin the BPF objects' compiler flags to the release build's, allow choosing the BPF compiler, and correct the rss_stat comment

josefbacik:
**What this changes.** Every BPF object is now compiled with `-mcpu=v3 -fwrapv`, and `build.rs` honours `SYSTING_BPF_CLANG=<path>` to pick the C compiler for the BPF objects. CI installs clang 21 from apt.llvm.org and sets `SYSTING_BPF_CLANG=clang-21` for the Test Suite and Clippy jobs. The comment at the rss_stat index bound is corrected (below). No runtime change: the release build already compiled at v3 with clang 21, so its object is unchanged.

**Why the pin.** The release build's clang 21 defaults to `-mcpu=v3`; the distro clang 18 used by local builds and CI defaults to v1. The two levels produce different programs from the same source — different register allocation around the same stores — and the verifier judges the program, not the source. #220's rss_stat index bound was a no-op in the release object while the clang-18 object compared the index register itself, so no local build ever exercised the program the release shipped (#221 has the details). Measured on the current source (`tp_btf/rss_stat`): clang 21 default ≡ clang 21 `-mcpu=v3` (366 instructions); clang 18 default ≡ clang 18 `-mcpu=v1` (370); clang 18 at `-mcpu=v3` (394) still differs from clang 21 at v3 in 160 lines, so the compiler version is a second variable — `-mcpu` alone gives the release's index-register shape (the AND lands on the index register in both v3 objects), and the compiler override closes the rest. A third variable turned up in the proof build: the release build's toolchain compiles with `-fwrapv` (its hardening set makes signed overflow wrap), and without it three programs — the memory recorder's brk/mmap/munmap exit handlers — differ from the release object in register allocation around their signed delta arithmetic (60 instructions across the three; the rss_stat programs are unaffected). Pinning `-fwrapv` removes that difference.

**Compatibility.** The source already requires v3 on clang 21: `-mcpu=v1` fails there with `error: Invalid usage of the XADD return value` (the `__sync_fetch_and_add` return needs v3 atomics). Every kernel the project targets (6.6 and later) runs v3 programs. `-fwrapv` only pins what the release already compiled with.

**The comment correction.** The comment added in #221 said the rejection depended on the verifier version. It does not: the store path the bound guards is live only when `tool_config.memory_rss_threshold_bytes` is non-zero, and at the object's default the verifier prunes it as dead code — which is why every load test at the default passed on every kernel while the same object was rejected in the field. With the threshold set, as the agent sets it before loading, a 6.12.85 verifier rejects the #220 object (`R0 unbounded memory access`) and loads the #221 one. The comment now says so, and that a load test must set the threshold to exercise the path at all. (A follow-up adds a load test that iterates the configuration values the program branches on.)

**Verification.**
- A local build with `SYSTING_BPF_CLANG` pointing at clang 21.1.8: the embedded BPF object's programs are instruction-identical to the ones carved from the release binary of the same source (every program section, call and jump targets included) — in the review comment.
- `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo test` on the tree; CI on this PR runs with the new toolchain step.

No tag bump is needed (the release object is unchanged; the comment is text); it rides the next tag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,11 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y linux-tools-common libelf-dev linux-libc-dev clang libbpf-dev make pkg-config jq libjemalloc2
+      - name: Install clang 21 (the release build's BPF compiler)
+        run: |
+          wget -q https://apt.llvm.org/llvm.sh -O /tmp/llvm.sh
+          sudo bash /tmp/llvm.sh 21
+          clang-21 --version
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.9
@@ -34,6 +39,7 @@ jobs:
           DUCKDB_LIB_DIR: "${{ github.workspace }}/libduckdb"
           DUCKDB_INCLUDE_DIR: "${{ github.workspace }}/libduckdb"
           LD_LIBRARY_PATH: "${{ github.workspace }}/libduckdb"
+          SYSTING_BPF_CLANG: "clang-21"
         run: cargo test --workspace --no-default-features
 
   fmt:
@@ -52,6 +58,11 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y linux-tools-common libelf-dev linux-libc-dev clang libbpf-dev make pkg-config jq libjemalloc2
+      - name: Install clang 21 (the release build's BPF compiler)
+        run: |
+          wget -q https://apt.llvm.org/llvm.sh -O /tmp/llvm.sh
+          sudo bash /tmp/llvm.sh 21
+          clang-21 --version
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.9
@@ -70,4 +81,5 @@ jobs:
           DUCKDB_LIB_DIR: "${{ github.workspace }}/libduckdb"
           DUCKDB_INCLUDE_DIR: "${{ github.workspace }}/libduckdb"
           LD_LIBRARY_PATH: "${{ github.workspace }}/libduckdb"
+          SYSTING_BPF_CLANG: "clang-21"
         run: cargo clippy --workspace --no-default-features -- -D warnings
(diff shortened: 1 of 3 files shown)


josefbacik:
**Review verdict — build: pin the BPF objects' compiler flags to the release build's, allow choosing the BPF compiler, and correct the rss_stat comment**

**Scope checked.** `build.rs`: `BPF_CFLAGS = ["-mcpu=v3", "-fwrapv"]` prepended to both objects' clang arguments (the pystacks object and the main object); `bpf_clang_override()` reads `SYSTING_BPF_CLANG` (with `cargo:rerun-if-env-changed`) and `pin_bpf_compiler()` applies it through `SkeletonBuilder::clang()` at both build sites; nothing else in the build changes. `.github/workflows/ci.yml`: an "Install clang 21" step (apt.llvm.org) in the Test Suite and Clippy jobs and `SYSTING_BPF_CLANG: "clang-21"` on their cargo steps. `src/bpf/systing_system.bpf.c`: the comment above the rss_stat index bound only (the code is unchanged). No Rust source, no maps, no programs, no runtime behaviour change.

**Why these two flags, measured.** Compiling the v1.15.2 BPF source with clang 21.1.8 at the build's flags: the default `-mcpu` is v3 (default vs explicit `-mcpu=v3`: identical across every program section); a distro clang 18 defaults to v1 and produces a different program (370 vs 366 instructions for `tp_btf/rss_stat`, the index compare on a different register). With `-mcpu=v3` and clang 21 alone, the object still differed from the release build's in three programs — `systing_brk_exit`, `systing_mmap_exit`, `systing_munmap_exit` (60 instructions of register allocation around their signed delta arithmetic); the release build's toolchain compiles with `-fwrapv` (signed overflow wraps), and with that flag the three match exactly while the rest are unaffected. `-fstrict-flex-arrays=1`, `-fPIC` and `-D_FORTIFY_SOURCE=3`, also in the release toolchain's set, change nothing for the BPF target.

**Proof.** A release build of this commit with `SYSTING_BPF_CLANG` pointing at clang 21.1.8: the embedded BPF object's disassembly is identical to the object carved from the release binary of the same source for every program section (17,942 lines, call and jump targets included); both `tp_btf/rss_stat` (366 instructions) and `tracepoint/kmem/rss_stat` (298) identical. Without the override (clang 18), the pinned flags still give the release's index-register shape (`r8 &= 0x3` on the index register at the use).

**Gates on the tree (identical bytes to the commit):** `cargo fmt --all -- --check` exit 0; `cargo clippy --all-targets` exit 0 (no warnings); `cargo test` exit 0 — 561 + 28 + 11 + 7 + 4 + 1 + 1 passed, 0 failed (the VM-only targets `#[ignore]`d); `cargo build --release` exit 0.

**Not in this change:** the load test over the runtime configuration shapes and the release-flag object (named in #221's correction comment) — a follow-up.

PR #223, opened 2026-08-26
memory: take the AnonHugePages sample only with the THP-split leg, count an unopenable VFIO window [bump patch]

josefbacik:
The memory recorder's end-of-capture `AnonHugePages` sample (`memory_rss` member -6, added in v1.15.0) read `/proc/<pid>/smaps_rollup` for every process that produced a memory event, on every `--memory` capture. Each of those reads is a page-table walk of that process in the kernel, so on a host whose processes map terabytes at 4 KiB the capture's stop phase spent seconds on a datum only the huge-page question needs. This moves the sample behind the THP-split leg: it is taken only when `--memory-thp-sample-rate` is greater than 0, which is the capture that asks that question; a plain memory capture no longer takes it. The continuous memory captures (`--memory` without the THP leg) therefore stop walking page tables at every stop.

Second, smaller fix in the VFIO leg: the kprobe that opens a VFIO ioctl's IOMMU-counting window ignored the return of its `bpf_map_update_elem`, so a window that could not be opened (the inflight hash full — more than 1024 traced tasks inside a VFIO ioctl at once) left that ioctl's runs uncounted with nothing saying so. It is now counted under `memory_iommu_overflow`, so `memory_iommu` reads as a floor with the count printed, the same way a full histogram already did. The message at capture end names both causes.

Documentation: SCHEMA_CHANGES v18 says the `anon_huge` member is present only when `memory_thp_leg` is set, and gains the reading rules a second review of the VFIO/THP leg turned up — the per-domain count when a container is attached to several IOMMU domains, `size_order` being floor(log2) of a run, the iommufd/cdev hosts and `.isra`/`.constprop` symbol suffixes that read as `off:nosym`, and the 6.9+ `split_huge_page_to_list_to_order` kretprobe covering every large-folio split while the 6.15+ `folio_split` path can bypass it. The `--memory-thp-sample-rate` help and the MCP trace-info description say the same.

Tests: the basic memory e2e now asserts that a capture without the THP leg writes no member -6 row; the THP e2e (which runs with the leg at rate 1) keeps its `anon_huge` read. Both run in the VM rig.

Not in this change, tracked as follow-ups: probing `__split_huge_pmd_locked` instead of `__split_huge_pmd` (which `split_huge_pmd_address()` also enters for non-huge PMDs on VMA adjust, over-counting `kind='pmd'`), and a window on the `vfio_iommu_type1_release` / detach path so unmaps that happen by closing the container are counted.


josefbacik:
**Review verdict — memory: take the AnonHugePages sample only with the THP-split leg, count an unopenable VFIO window**

**Scope checked.** Seven files, +70/−13: one Rust gate (`systing_core.rs`: `emit_anon_huge_pages` now under `if opts.memory_thp_sample_rate > 0`), one BPF hunk (`memory_vfio_emit`: the `bpf_map_update_elem` return on `memory_vfio_inflight` checked, a miss counted in `memory_iommu_overflow`), the capture-end message, the CLI help, the MCP trace-info description, SCHEMA_CHANGES v18, and a new assertion in the basic memory e2e. No map, program, payload or schema change; no new autoload; the classic and tp_btf rss_stat programs untouched.

**The gate.** `memory_thp_sample_rate` is the option that turns the THP-split leg on (`probe_memory_kernel_legs` at 785–803 uses the same test), so "anon_huge only with the THP leg" is one predicate in two places that already agree; the THP e2e runs with rate 1 and keeps its read, the basic memory e2e runs with rate 0 and now asserts no member -6 row. `emit_anon_huge_pages` itself is unchanged (per-tgid `read_anon_huge_pages`, dead tgids skipped at ENOENT). Continuous captures pass no THP rate, so their stop phase loses the walk entirely; on-demand THP captures keep it.

**The BPF hunk.** `bpf_map_update_elem` on a `BPF_MAP_TYPE_HASH` with `BPF_ANY` returns non-zero only when a NEW key cannot be inserted (E2BIG at `max_entries = 1024`) or on an error the verifier permits the helper to return; an existing key is overwritten and returns 0, so the counter increments exactly when no window exists for this task after the call — the only case the tracepoints then under-count. The counter is the existing per-CPU `memory_iommu_overflow` (lookup of index 0 on a `PERCPU_ARRAY`, non-atomic `+= 1` in kprobe context on the owning CPU, the same pattern as `memory_iommu_account` at 6186), read by the existing `sum_percpu_counter` at capture end. Semantics widened by one cause ("runs or windows"), stated in the message and in SCHEMA_CHANGES; the table was already documented as a floor. Verifier: what this read holds is inferred from the source only — a checked helper return and the existing per-CPU counter pattern, no runtime-indexed access, straight-line code before `reserve_memory_event`; the verifier's own verdict on the two `kprobe/vfio_dma_do_{map,unmap}` programs comes from loading the release-flag object (clang 21, `-mcpu=v3 -fwrapv`) with the memory recorder's runtime configuration, which is the check this object gets before it ships and is not claimed here. Map/context pairing: `memory_vfio_inflight` is written only from the kprobe (task context) as before; no NMI writer.

**Docs.** SCHEMA_CHANGES v18 now carries: anon_huge present only when `memory_thp_leg` is set (with the version note), the per-domain tracepoint count, floor(log2) `size_order`, iommufd/cdev hosts and `.isra`/`.constprop` symbol suffixes reading `off:nosym`, the 6.9+ kretprobe superset and the 6.15+ `folio_split` bypass. Help text and the MCP description agree with it.

**Gates on the tree (identical bytes to the commit):** `cargo fmt --all -- --check` exit 0; `cargo clippy --all-targets` exit 0, 0 warnings; unit suite 561 + 11 + 28 passed, 0 failed; VM rig (the project's 6.12 test kernel), `memory_recorder` target: 8 passed / 0 failed — `test_memory_recorder_e2e` (with the new assertion: no member -6 row without the THP leg), `test_memory_thp_vmstat_e2e` (the leg on: the `anon_huge` read still taken), the four memory-alloc tests and both rss-threshold tests.

**Findings:** Critical none / High none / Medium none / Low 1 — the kretprobe-miss window leak (a `maxactive` miss on `vfio_dma_do_map_ret` leaves a task's window open until its next ioctl) is unchanged and bounded by the window's `[iova, iova+size)` range check; noted for the follow-up that adds the release-path window.

PR #224, opened 2026-08-26
Close the still-running sched slices at the tracing gate, not after the stop sequence

josefbacik:
## What changes

The sched slices still open when a capture stops now close at the instant the tracing gate closed — the end of event collection — instead of at the time the trace is finally written. `run_tracing_loop` reads `CLOCK_BOOTTIME` right after it disables the gate and returns it; `systing()` hands it to `generate_parquet_trace(tpu_recorder, end_ts)`, which passes it to every recorder shard's `finish(end_ts)` in place of the clock read that used to happen there. `finish()` closes a slice whose start stamped past the gate with zero length instead of a negative one. Docs on the three functions and in the `sched aggregate` module say what the window end means.

## Why

`generate_parquet_trace()` runs at the end of the stop sequence: after the 1 s continuous-mode stop delay, the ring-buffer drain, the joins of every poller and recorder thread and the symbol loader. On a busy host that is seconds, and under a large event backlog minutes. The final slice per CPU (`end_state = NULL`, "still running at trace end") was being cut with `dur = now − start`, so each one absorbed the whole stop latency.

`sched aggregate` takes its default window from the trace's own extent (`MIN(ts)` … `MAX(ts + dur)` over `sched_slice`), so those slices stretched the window by the same amount: across a large continuous-profiling population the window was ≈11 s at the median for a 10 s capture (the stop delay plus fast joins), but on ≈0.5 % of captures it ran to tens of seconds and up to ≈5 minutes with zero dropped events — every per-window count, `*_per_s` rate, `slice_len` maximum and the work-conservation fraction on those rows were computed over a span that contained no events. The signature on the rows is exact: `slice_len_max_ns` ≈ the window's excess over the capture length on every row, normal ones included (≈1 s: the stop delay), which also means `slice_len_max_ns` was measuring the stop phase, never a real slice.

Moving the end to the gate makes the window the span events were collected over, which is what every per-window figure divides by.

## What the reviewer decides

1. **Where the gate timestamp is read.** Immediately after `tracing_enabled = false`, before the ring consumers are told to shut down — the last instant at which a BPF program could have stamped an event. The event clock (`bpf_ktime_get_boot_ns`) and the gate read are independent, so an in-flight event can stamp microseconds past the gate; `finish()` clamps that slice to zero length rather than emitting a negative duration, and the new unit test pins both the gate close and the clamp.
2. **No change to `sched aggregate`'s window query for traces written by earlier recorders.** Their open slices still end after the stop sequence. A belt would be to take the window end from the last *completed* slice or marker (`MAX(ts + dur) WHERE end_state IS NOT NULL`, `MAX(ts)` otherwise) — but for traces from this recorder that would end the window at the last switch instead of the gate and shorten idle hosts' windows, so it is left out; the module docs tell a reader of older rows to compare by `window_ns` and to read `slice_len_p99_ns` rather than the maximum. Happy to add the belt behind a flag if wanted.

No schema, version, or BPF change; the `sched aggregate` output shape is unchanged.

## Tests

`sched::tests::test_finish_closes_running_slices_at_the_gate` — a task still running closes exactly at the passed `end_ts`; a completed slice keeps its event-clock length; a task switched in past the gate gets a zero-length final slice. The existing `test_finish_drains_all_buffered_record_types` keeps its explicit `end_ts`. `cargo test --lib --bins sched`, `cargo clippy --all-targets`, `cargo fmt --check` — results in the PR's first comment.

### DIFF
--- a/src/analyze/sched_aggregate.rs
+++ b/src/analyze/sched_aggregate.rs
@@ -66,6 +66,14 @@
 //!   that its state is unknown); CPUs with no events at all are reported as
 //!   unobserved, never assumed idle or busy. Waits that do not end inside the
 //!   window are counted as censored, never extrapolated.
+//! - The default window is the trace's own extent: from its first slice or
+//!   runnable marker to the end of its last slice. The recorder closes the
+//!   slices still open when collection stops at the instant the tracing gate
+//!   closed, so that end is the end of collection. (Traces written by
+//!   recorders before that change closed them after the stop sequence — ring
+//!   drain and thread joins — so on a loaded host their window, `slice_len`
+//!   maximum and every per-window count or rate carry that stop latency;
+//!   compare such rows by `window_ns`, not by `max`.)
 //!
 //! Percentiles come from a log-linear histogram (16 sub-buckets per octave,
 //! upper bucket edge reported, so a percentile is at most 6.25% above the
--- a/src/sched.rs
+++ b/src/sched.rs
@@ -420,6 +420,13 @@ impl SchedEventRecorder {
     /// 2. Emit unpaired IRQs/softirqs with dur=0
     /// 3. Drain every locally buffered record type to the collector
     ///
+    /// `end_ts` is the instant event collection ended — the tracing gate, not
+    /// the time this method happens to run — and is where the final slices
+    /// close. A switch observed in the last microseconds before the gate can
+    /// carry a timestamp past it (the gate and the event clock are read
+    /// independently), so a slice that would end before it started is
+    /// closed with zero length instead of a negative one.
+    ///
     /// Returns the streaming collector so the caller can call finish() on it.
     pub fn finish(&mut self, end_ts: i64) -> Result<Option<Box<dyn RecordCollector + Send>>> {
         if self.streaming_collector.is_some() {
@@ -428,7 +435,7 @@ impl SchedEventRecorder {
                 .cpu_states
                 .iter()
                 .map(|(cpu, state)| {
-                    let dur = end_ts - state.start_ts;
+                    let dur = (end_ts - state.start_ts).max(0);
                     SchedSliceRecord {
                         ts: state.start_ts,
                         dur,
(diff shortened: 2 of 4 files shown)


josefbacik:

Review mechanics: the repository's `rust-bpf-code-reviewer` agent definition (blob `3739c8b3`, unchanged since the last review) was executed in-session over the full diff of this branch against `main` (v1.15.2 `d68f618`): `src/systing_core.rs`, `src/session_recorder.rs`, `src/sched.rs`, `src/analyze/sched_aggregate.rs`; +116/−10, all five passes. No BPF, schema, recorder-table, Python-version or ELF-parsing code is touched; `Cargo.toml`'s version is untouched; the `sched aggregate` output shape is unchanged (the only hunk there is module documentation).

### Summary
Moves the timestamp at which a capture's still-open sched slices close from "whenever the trace is written" (after the ring drain and every thread join) to the instant the tracing gate closed, by returning that instant from `run_tracing_loop` and passing it through `generate_parquet_trace` to `SchedEventRecorder::finish`. The plumbing is the smallest that works (one returned value, one new parameter, one clamp) and every caller is updated; the new unit test pins both the gate close and the clamp.

### Critical Issues
None.

### Improvements
- (Taken before commit) `finish()` clamps a final slice to zero length when an in-flight switch stamped past the gate, rather than writing a negative `dur`; the gate and the event clock are read independently, so that ordering can occur and is now covered by `test_finish_closes_running_slices_at_the_gate`.
- (Taken before commit) The gate timestamp escapes the skel-lifetime block through a deferred `let capture_end_ts: i64;` assigned exactly once, instead of turning a 700-line block into an expression or widening the block's scope — the compiler checks the single initialization.

### Suggestions
- The memory recorder's end-of-capture rows (`stop_phase("memory recorder end samples")`) still stamp `CLOCK_BOOTTIME` after the stop sequence. Those are point samples, not durations, so the stretch does not propagate into a window — but stamping them at the gate too would make "end of capture" one instant everywhere. Separate change if wanted.
- For traces written by recorders before this change, `sched aggregate` could take its window end from the last completed slice or marker instead of the last slice end. Left out deliberately (it would shorten idle hosts' windows on new traces); recorded in the PR body as the reviewer's call, and in the module docs as a read rule for older rows.
- The doc comment on `generate_parquet_trace` is four sentences on one parameter; it carries the why (the stop sequence's latency) rather than the what, which is the standard this tree holds comments to, but it could lose its last clause without losing the point.

### Positive Notes
- The gate read sits between `tracing_enabled = false` and `ringbuf_shutdown.signal()` — after the last instant a BPF program could stamp an event and before any consumer is told to stop — so the value means exactly what the name says.
- The test constructs the past-the-gate case explicitly (a switch stamped at 10_020 against a gate of 10_000) beside the normal gate close, so the new boundary is the thing under test, not an incidental.
- `cargo fmt --check`, `cargo clippy --all-targets` (clean) and `cargo test --lib --bins sched` (35 passed, 0 failed) on the branch; no new dependencies.

PR #225, opened 2026-08-26
test: load the BPF object at every configuration shape that ships

josefbacik:
A load-only test that opens the BPF object, configures it exactly as a capture does for each configuration shape that ships, and loads it into the kernel — attaching nothing — and fails on any rejection with the verifier log printed whole. A second, slower test unions the instructions the verifier visited across all shapes and names any instruction no shape reaches.

Why: the object's behaviour is selected by `.rodata` constants (`tool_config`: which recorders are on, sample rates, thresholds, filters) that userspace sets before load. libbpf freezes `.rodata` and the verifier constant-folds every branch on those constants, so a load verifies exactly the code the configuration enables and prunes the rest. The rss_stat rejection fixed in #220/#221 was invisible to every load we ran because the store path is gated by `memory_rss_threshold_bytes != 0`: at the object's defaults that threshold is 0, the branch is dead, and the program loaded everywhere; with the memory recorder's configuration the branch is live and the verifier rejected it. A load at one shape says nothing about another shape, and a shape nobody loads before a tag is a shape that is first loaded on a real host.

What is in the change:
- `systing_core::bpf_load_probe(opts, legs, log_level)`: the open → configure → load sequence of `systing()` (the same `configure_bpf_skeleton`, the same autoload selection, the same map sizing for the shapes without perf counters or cgroup targets), attaching nothing. `legs` selects the kernel-probed memory legs as the host allows or forces them on: a kprobe or tracepoint program verifies without its attach target, so a host without the vfio_iommu_type1 module can still verify the VFIO programs, which a capture on such a host never selects. `log_level` is consulted once per selected program: 0 is the plain load (libbpf refetches the log at level 1 on a rejection), 2 records every instruction the verifier visited. The report carries, per program: selected or not, instruction count (read from the loaded object, after subprogram appending), the visited set, and the program's section of the verifier log.
- `bpf_load_shapes`: the shape table (every row one `Config`, with the values a continuous capture runs with where they differ from the CLI defaults — CPU lane defaults and off arms, sw-event, build-id, syscalls/markers, pystacks, ring shards and sizes, the pid filter, a trace-event row; the memory recorder at its continuous shape, at rss threshold 0, at fault/map sample rates 1 and 97, the classic rss_stat twin, memory-alloc plain and sampled, VFIO + THP; the three network shapes; everything on; and two forced rows for the VFIO/THP legs, one per folio-split kretprobe twin), the level-2 log parsing (`split_prog_load_logs`, `visited_insns`), and the coverage arithmetic (`coverage_gaps`, `ranges`) — plain functions with unit tests.
- `tests/bpf_load_shapes.rs`, two `#[ignore]` root tests. `every_shape_loads` is the gate: every shape loads, and every program in the object is selected by some shape unless listed in `SELECTED_ELSEWHERE` with the reason (a twin the running kernel does not pick, or a program that needs an attach-time input such as a cgroup target); it loads at level 0 and takes a few minutes even on a KVM-less guest. `every_instruction_is_verified_by_some_shape` is the coverage read: any instruction of a loaded program that no shape visited fails unless listed in `ALLOWED_UNVISITED` with a reason; it asks for level 2 only on programs whose union is still incomplete after the shapes loaded so far, because a level-2 load costs a formatted line per instruction per state (the first cut logged every program at every shape and did not finish in 30 minutes under TCG).

A new knob is one new row in the table. The tests run where the other root tests run (`./scripts/run-integration-tests.sh bpf_load_shapes`, or the VM rig with `-f every_shape_loads` for the gate alone).

Test plan
- Unit tests for the parser, the coverage union and the table: `cargo test --lib bpf_load_shapes` — 7 passed.
- VM rig (the project's `bzImage-v6.12-bpf`, vanilla 6.12, TCG guest): `every_shape_loads` at this tree — 1 passed, 167 s in the guest; 30 shapes, 10–66 programs selected each, 2.3–25.7 s per shape.
- Positive control (not committed): the same test on the v1.15.1 `handle_rss_stat` form — #221's barrier-first mask reverted in the working tree — fails with `verifier rejected systing_rss_stat_btf at shape(s) memory-default, memory-sw-event, memory-sample-rates-1, memory-sample-rates-97, memory-alloc, memory-alloc-sampled, memory-vfio-thp, memory-ringbuf-shards-8, all-recorders, memory-vfio-thp-forced, memory-vfio-thp-forced-legacy`, each with the same verifier words the affected hosts reported, `R0 unbounded memory access, make sure to bounds check any such access … processed 407 insns`, while `memory-rss-threshold-0` and `memory-rss-classic` load. The same vanilla kernel that accepted that object at the defaults rejects it at the recorder's configuration, so the shape — not the verifier version — is what the earlier local and VM loads were missing.
- Not covered here: the other kernels in use (newer 6.12.y stable and 6.18 verifiers). The rig has only the vanilla 6.12 image today; a kernel matrix is the next step of the same program, and this test is the thing that runs on it.

### DIFF
--- a/src/bpf_load_shapes.rs
+++ b/src/bpf_load_shapes.rs
@@ -1,0 +1,508 @@
+//! Load-every-program × every-configuration-shape: the data and the
+//! arithmetic behind `tests/bpf_load_shapes.rs`.
+//!
+//! The BPF object's behaviour is configured through `.rodata` constants that
+//! userspace sets before load (`tool_config`: which recorders are on, sample
+//! rates, thresholds, filters). libbpf freezes `.rodata`, and the verifier
+//! constant-folds every branch on those constants, so the code a load
+//! verifies is exactly the code the configuration enables — a branch no
+//! shape turns on is never verified, and a defect in it ships. The
+//! instrument here is therefore two-fold: every shape that ships must LOAD
+//! (the verifier accepts the object at that configuration), and the union
+//! over all shapes of the instructions the verifier visited must cover every
+//! instruction of every program that loaded — an instruction no shape
+//! reaches names a configuration the table is missing (or dead code, which
+//! is then allowed explicitly, with a reason).
+//!
+//! [`crate::systing_core::bpf_load_probe`] performs one load at one shape and
+//! returns the per-program visited sets, read from the verifier's level-2
+//! log; this module owns the shape table, the log parsing, and the coverage
+//! arithmetic, all of which are plain functions with unit tests.
+
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+
+use crate::systing_core::Config;
+
+/// One program's outcome in a load probe.
+#[derive(Debug, Clone)]
+pub struct ProgramLoad {
+    pub name: String,
+    /// Selected for load at this shape (`get_required_bpf_programs`).
+    pub autoload: bool,
+    /// Instructions in the loaded program (main body plus the subprograms
+    /// libbpf appended to it); 0 when the object did not load or the program
+    /// was not selected.
+    pub insn_total: usize,
+    /// Instruction indices the verifier visited, from the level-2 log.
+    pub visited: BTreeSet<u32>,
+    /// The program's own section of the verifier log, when one was printed.
+    pub verifier_log: Option<String>,
+}
+
+/// The outcome of one load probe.
+#[derive(Debug, Clone)]
+pub struct LoadReport {
+    /// The whole object loaded (every selected program verified).
+    pub loaded: bool,
+    /// libbpf's error when it did not.
+    pub error: Option<String>,
+    pub programs: Vec<ProgramLoad>,
+}
+
+impl LoadReport {
+    /// The programs libbpf reported as rejected by the verifier at this
+    /// shape (their log section starts with libbpf's failure line).
+    pub fn failed_programs(&self) -> Vec<&ProgramLoad> {
+        self.programs
+            .iter()
+            .filter(|p| {
+                p.autoload
+                    && p.verifier_log
+                        .as_deref()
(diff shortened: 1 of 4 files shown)


josefbacik:
Reviewer pass (the repository's rust-bpf-code-reviewer rubric), run over the full diff before push.

### Summary
A load-only probe plus a table of configuration shapes, with the table, the log parsing and the coverage arithmetic as plain functions under unit tests, and two root tests that drive them. No BPF source changes, no schema changes, no Cargo.toml edit. The probe mirrors `systing()`'s own open → configure → load prelude rather than re-deriving it, so the shapes it loads are the shapes a capture loads.

### Critical Issues
None.

### Improvements
- `bpf_load_probe` swaps libbpf's process-global print callback and reads a process-global capture buffer; two tests in one binary could interleave their captures under libtest's default parallelism. Resolved before push: probes are serialized on a lock inside the function (`PROBE`), and the doc comment says so.

### Suggestions
- `split_prog_load_logs` is a chain of `if let`s on the libbpf line shapes; a `match` on a small parsed enum would read a little better. Left as is — three shapes, each a one-line test.
- libtest captures the per-shape `eprintln!` lines on success; a `--nocapture` run is the way to keep the per-shape timings in the log.

### Positive Notes
- The shape table is data: a new knob is one new row, and the table's own unit test pins that the memory, VFIO/THP, network and alloc shapes are present.
- Instruction counts come from the loaded object (after subprogram appending), so the coverage union indexes the program the verifier log indexes.
- The forced VFIO/THP rows exercise programs that a capture on the test kernel never selects; load ≠ attach, so they verify without their attach targets.
- The positive control in the test plan fails for the right reason, at the right shapes, and passes at the threshold-0 and classic-twin shapes that prune the store path.

PR #226, opened 2026-08-26
memory recorder: probe the PMD-split worker, count teardown unmaps, bound the AnonHugePages walk, attach the legs one at a time [bump minor]

josefbacik:
Follow-ups to the VFIO/THP legs (#219) from reading the kernel they probe, the two fixes the v1.15.4 gate (#223) left open, and a brk guard. Schema 19 (`[bump minor]`). Stacked on #225 (the shape test) so the new programs are covered by it; merge #225 first and this PR's diff is its last commit only.

### PMD splits

The `kind = 'pmd'` probe sat at `__split_huge_pmd`, the public entry, which `split_huge_pmd_address()` also calls for a PMD that is not huge (any populated PMD on a VMA adjust), so the rows over-counted on VMA churn and missed the rmap paths (`try_to_unmap` / `try_to_migrate` go through `split_huge_pmd_locked` on 6.10+). The probe now sits at `__split_huge_pmd_locked`, the static worker every PMD split reaches and where vmstat counts `thp_split_pmd`, whenever kallsyms exports it; a row is then one counted split, bounded by the vmstat delta, and carries the split's `freeze` flag (1: for migration or unmap) in `result`. A build that inlined the worker falls back to the entry probe and says so: `sysinfo.memory_thp_leg = on:pmd-entry`.

### Teardown unmaps

Closing the last file of a VFIO container (`vfio_iommu_type1_release`) unmaps every region, and a group detaching from a domain (`vfio_iommu_type1_detach_group`) unmaps the domain's; neither passes `vfio_dma_do_unmap`, so a model unload that just closed its device files left no unmap row and its `iommu:unmap` runs fell outside every histogram window. A kprobe/kretprobe pair on each opens a `[0, ~0)` window; the release also writes one `memory_vfio` unmap row with a teardown flag (bit 31 — `flags < 0` in the column; `size = -1`). When the two symbols are missing the leg reads `on:noteardown`.

### The AnonHugePages walk, bounded

`emit_anon_huge_pages` (gated behind the THP leg since v1.15.4) reads `/proc/<pid>/smaps_rollup` — a page-table walk of the process in the kernel — for every process that produced a memory event. It is now bounded: at most 64 processes, the ones whose memory events moved the most bytes first, under a 500 ms wall budget, and the capture records what it did in `sysinfo.memory_anon_huge_walk` (`complete:<read>/<candidates>`, `capped:…`, `budget:…`).

### Attach, one leg at a time

The kernel-probed memory legs were attached by `skel.attach()`, all or nothing: a symbol gone between the kallsyms read and the attach, or one the kernel will not probe, failed the whole capture, CPU profile included. They are now attached one leg at a time after `skel.attach()`; a leg whose probe fails is dropped for that capture (`off:attach`, `on:noteardown`) and the rest runs.

### Smaller

- Two facts a reader needed from stdout are in `sysinfo`: `memory_iommu_overflow` (0 = the `memory_iommu` histogram is complete, else a floor) and `memory_anon_huge_walk` above.
- `brk`: an `-EINTR` return (the process is being killed while the syscall waits for the mmap lock) no longer produces a row — the exit handler computed `ret - old_brk` with `ret = -4`, a delta of `-(old_brk + 4)`: tens of TB with `|bytes| ≡ 4 (mod 4096)`, seen about once a week.
- Docs: `kind = 'pmd'` ≡ `thp_split_pmd` now holds at the worker probe and SCHEMA_CHANGES says when it does not; the AnonHugePages gate's first release is v1.15.4 (v1.15.3 was tagged from #222 minutes before #223 merged); an iommufd host reads `memory_vfio_leg = off:nosym` only with the module unloaded.

### Tests and evidence

- The shape table's forced rows select both PMD-split twins and the teardown pair, so every new program is verified at the recorder's shape wherever the shape test runs. VM rig (vanilla 6.12 test kernel): `every_shape_loads` PASS (30 shapes, 169 s).
- `tests/memory_recorder.rs`: 9/9 on the rig. The THP e2e asserts the pmd-rows-within-vmstat-delta bound only at the worker probe. A new e2e keeps a huge-page-advised workload alive across two duration-stopped captures (mapping and unmapping a page each turn so every capture sees it) and asserts the AnonHugePages sample is absent without the THP leg and present, complete and non-zero with it — the command-mode test could not tell, because its workload was reaped before the end samples on either code path.
- Unit tests: the leg-selection matrix (every `memory_thp_leg` / `memory_vfio_leg` value incl. the attach-failure causes), the sysinfo parquet round-trip with the two new columns (and NULL when a leg did not run), `trace info` reading them.
- Reviewer pass against the repository's reviewer rubric: no critical issues; two improvements applied before this commit (the leg links detach inside the `detach bpf programs` phase; the new e2e's workload produces a memory event by construction).

Not covered here: the fallback-to-entry probe on a host whose build inlined the worker, and the teardown pair on a VFIO host, have been load-verified (the shape gate) but not attach-exercised on real hardware — the on-demand THP/VFIO capture on a VFIO host is the first contact for those two.


josefbacik:
# Reviewer pass — memory recorder follow-ups (PMD worker probe, teardown unmaps, bounded AnonHugePages walk, per-leg attach) [bump minor]

Rubric: the repository's `rust-bpf-code-reviewer` agent definition (structure → Rust → BPF → schema/versioning → integration passes). Diff reviewed: 14 files, +818/−126 on top of the shape-test commit (`16cac88b`); rig evidence: the shape gate PASS (every configuration shape loads, the forced rows selecting both PMD-split twins and the teardown pair) and `tests/memory_recorder.rs` 9/9 on the project's 6.12 test kernel.

## Summary

The change does what its description says: the PMD-split probe moves to the worker every split reaches (with a declared fallback), teardown-by-close unmaps get a window and a flagged row, the end-of-capture AnonHugePages walk is bounded and self-describing, the optional kernel legs attach one leg at a time so a failed probe no longer costs the CPU profile, and a brk error return no longer becomes a row. Schema 19 carries two `sysinfo` columns; every new `sysinfo` value form is documented in SCHEMA_CHANGES.

## Critical Issues

None. Schema/versioning rubric: SCHEMA_CHANGES has the v19 entry (and the corrected v18 text); `SCHEMA_VERSION` is 19; `[bump minor]` is in the title; `Cargo.toml`/`Cargo.lock` untouched.

## Improvements (applied before the commit)

1. **Leg links were dropped at scope end, not in the `detach bpf programs` phase.** `attach_memory_kernel_legs` returns its links into a binding that outlived the explicit `drop(skel)` step, so the VFIO/THP probes stayed attached through the drain and write phases (harmless — the tracing gate is closed — but outside the bracketed step that exists to account for every detach). Fixed: `drop(_memory_leg_links)` beside `drop(_probe_links)`.
2. **The new e2e relied on an incidental memory event.** The long-lived workload's idle loop wrote one already-mapped byte per turn; the capture saw the process only because the interpreter happened to allocate. The loop now maps and unmaps a page each turn, so every capture has a memory event from the workload by construction. Test-only; re-run on the rig after the change.

## Suggestions (not applied)

- `memory_anon_huge_walk` reports `complete:<read>/<candidates>` where `candidates − read` are processes gone at capture end; a reader wanting the gone count has to subtract. Acceptable as documented; a `gone` figure in the string would be a later schema note, not worth a fourth field now.
- `seen_tgids` weights processes by `|size|` summed across event types (rss deltas, map sizes); faults contribute 0. Heaviest-first ordering is a heuristic for which 64 processes to read under the cap, not a measurement, and the documentation says so.

## Positive Notes

- The teardown pair reuses the ioctl pair's window bookkeeping (`memory_vfio_open_window` factored out rather than duplicated), and the detach pair opens a window without writing a row — the comment explains why the two differ.
- `thp_leg_value` / `vfio_leg_value` enumerate every combination in one `match` each, with unit tests for every arm including the attach-failure causes.
- The brk guard sits after the scratch entry is consumed, so an error return leaves no stale scratch row.
- `MemoryRecorderEnd` is a `OnceLock` set before the trace is written; the sysinfo columns cannot race the writer.

## Checks

- BPF: the `__split_huge_pmd_locked` probe uses the worker's 4-argument signature (6.6, 6.12, 6.18); `freeze` is carried as the row's result. The teardown programs are `kprobe`/`kretprobe` on out-of-line static ops callbacks, selected only when kallsyms lists both symbols.
- Parquet/DuckDB: the two new columns round-trip (writer unit test), NULL when the leg did not run; `trace info` and the MCP description read them by name; pre-v19 imports stay by-name.
- Recorder consistency: member −6 rows carry `utid` via `get_or_create_utid`; no raw pid columns added.

PR #227, opened 2026-08-27
analyze: match wakeups to slices with an ASOF JOIN in the cpu_stats runqueue query

josefbacik:
SQL-only change to `cpu_stats`' runqueue-depth query; no schema change, no BPF change.

## Why

`cpu_stats` estimates per-CPU runqueue depth by matching every wakeup (`thread_state.state = 0`) to the `sched_slice` it precedes. The match was an inequality join — `w.ts BETWEEN COALESCE(s.prev_end_ts, 0) AND s.ts` over all slices of the same utid — followed by `DISTINCT`, with a `ROW_NUMBER` window to dedup wakeups. Under the locked DuckDB configuration the MCP server runs with (`memory_limit` derived from the cgroup, `threads` capped, a bounded spill directory), that is the plan an analyzer hangs on when the trace is large: the range join builds an intermediate DuckDB cannot spill, so on whole-node traces of tens of millions of rows the first `cpu_stats` call either crawls past the analyzer's deadline or fails with an out-of-memory error.

## What changes

Consecutive slices of one utid have disjoint `[prev_end_ts, ts]` windows, so at most one slice qualifies per wakeup. An `ASOF JOIN` — the first slice of the same utid that starts at or after the wakeup, with the window's lower bound kept as a post-filter — produces exactly the same matches in one ordered merge pass, with no `DISTINCT` needed (ASOF is 1:1). The wakeup dedup becomes `SELECT DISTINCT` over the same `(ts, utid, cpu)` key. The percentile stage (`QUANTILE_CONT` over the running depth) is untouched, so the numbers are the same.

## Measured

192-CPU, 6000-thread, 60 s synthetic trace in systing's schema (`sched_slice` + `thread_state`), one query per process, DuckDB `memory_limit` / `threads` / `temp_directory` / `max_temp_directory_size` set the way the locked config sets them:

| rows | bound | before (range join) | after (ASOF) |
|---|---|---|---|
| 62 M | 4 GiB / 4 threads | 58 s, no spill | 15 s, no spill |
| 62 M | 1.5 GiB / 2 threads | out-of-memory error | 30 s, spills 1.8 GiB |
| 186 M | 4 GiB / 4 threads | out-of-memory error | 81 s, spills 7.2 GiB |
| 186 M | 8 GiB / 4 threads | 428 s | 53 s, spills 3.9 GiB |

The cost was the join, not the quantile: swapping `QUANTILE_CONT` for `approx_quantile` on top of the ASOF form changed the wall time by 0.1 s, so the exact quantile stays.

## Tests

- The shape tests in `cpu_stats.rs` are updated for the new CTEs (`wakeups` via `SELECT DISTINCT`, `ASOF JOIN … WHERE w.ts >= COALESCE(s.prev_end_ts, 0)`, no `INNER JOIN`, no `ROW_NUMBER`).
- New `test_runqueue_query_asof_matches_range_join`: an in-memory DuckDB fixture (4 CPUs, 5 threads each, 300 slices per CPU; a wakeup before most slices, spurious wakeups inside the thread's previous run, cross-CPU wakeup targets, duplicate wakeup rows, a thread that wakes but never runs, non-wakeup `thread_state` rows) runs the old query verbatim as the oracle and asserts the `events` multiset is identical, the matched-wakeup counts are equal and sit strictly between 0 and the wakeup count (both branches exercised), and the per-CPU p50/p90/p99 agree within the one depth step that the `+1`/`-1` pair of a matched wakeup at one timestamp leaves to sort order.
- On the 62 M-row synthetic trace the ASOF form reproduced the range-join form's p50 on all 192 CPUs and p90/p99 on 190 of them (±1 on two, the same tie-order effect).

Gates on this tree (DuckDB v1.5.4, the version `libduckdb-sys` pins): `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` 563 lib tests + the integration targets passed, 0 failed.

### DIFF
--- a/src/analyze/cpu_stats.rs
+++ b/src/analyze/cpu_stats.rs
@@ -237,18 +237,28 @@ fn build_softirq_time_query(trace_id: Option<&str>) -> String {
 fn build_runqueue_query(trace_id: Option<&str>) -> String {
     let filter_ts = trace_id_filter(trace_id, "ts.");
     let filter_ss = trace_id_filter(trace_id, "ss.");
-    // Deduplicate wakeups per (utid, scheduling-window) so that spurious
-    // wakeups between the same pair of sched_slices produce only one +1 event.
-    // The -1 side already uses DISTINCT, so this keeps +1/-1 balanced.
+    // Deduplicate wakeups per (utid, target cpu, ts) so that spurious wakeups
+    // between the same pair of sched_slices produce only one +1 event. The -1
+    // side is one row per matched wakeup, so this keeps +1/-1 balanced.
+    //
+    // A wakeup is matched to the sched_slice it precedes: the first slice of
+    // the same utid that starts at or after the wakeup, provided the wakeup
+    // came after the thread's previous slice ended (a wakeup that lands inside
+    // the previous run is spurious and stays unmatched). Consecutive slices of
+    // one utid have disjoint [prev_end_ts, ts] windows, so at most one slice
+    // qualifies per wakeup, and an ASOF JOIN -- one ordered merge over both
+    // sides -- yields the same matches as the range join
+    // `w.ts BETWEEN COALESCE(s.prev_end_ts, 0) AND s.ts` it replaces. That
+    // range join built an intermediate DuckDB cannot spill: on a 192-CPU trace
+    // of 62 M rows it took 58 s and failed with an out-of-memory error under a
+    // 1.5 GiB memory_limit, and at 186 M rows it took 428 s at 8 GiB and
+    // failed at 4 GiB, while this form takes 15 s / 53 s and spills gracefully
+    // under the same limits.
     format!(
-        "WITH dedup_wakeups AS (\
-             SELECT ts.ts, ts.utid, ts.cpu as target_cpu, \
-                    ROW_NUMBER() OVER (PARTITION BY ts.utid, ts.cpu, ts.ts ORDER BY ts.ts) as rn \
+        "WITH wakeups AS (\
+             SELECT DISTINCT ts.ts, ts.utid, ts.cpu as target_cpu \
              FROM thread_state ts \
              WHERE ts.state = 0 AND ts.cpu IS NOT NULL{filter_ts}\
-         ), \
-         wakeups AS (\
-             SELECT ts, utid, target_cpu FROM dedup_wakeups WHERE rn = 1\
          ), \
          sched_with_prev AS (\
              SELECT ss.ts, ss.cpu, ss.utid, \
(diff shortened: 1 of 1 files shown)


josefbacik:
rust-bpf-code-reviewer pass — executed in-session over the full diff of this PR (`.claude/agents/rust-bpf-code-reviewer.md` at `65c4535`, read as a tracked blob), before commit/push, per the repo's review-before-commit rule. Mechanics: the authoring session ran the definition's five passes itself; no separate agent process.

### Summary
A SQL-only change to `build_runqueue_query` in `src/analyze/cpu_stats.rs`: the wakeup→slice match moves from an inequality (range) join + `DISTINCT` to an `ASOF JOIN` with the same window predicate as a post-filter, and the wakeup dedup from a `ROW_NUMBER` window to `SELECT DISTINCT` over the same three columns. The percentile stage is untouched. The change is minimal, the match semantics are argued in the comment and pinned by a fixture test that runs the old query as an oracle, and the measured motivation (58 s → 428 s superlinear / OOM at a tight bound vs 15 s → 53 s with graceful spill) is stated where the next reader will look.

### Critical Issues
None. No schema change (`SCHEMA_VERSION`, `SCHEMA_CHANGES.md` untouched), no `Cargo.toml`/`Cargo.lock` version edit, no BPF code, no recorder, no ELF parsing, no `unsafe`.

### Improvements
- None required. Equivalence argument checked: consecutive slices of one utid have disjoint `[prev_end_ts, ts]` windows, so the range join matched at most one slice per wakeup; `ASOF JOIN … ON w.utid = s.utid AND w.ts <= s.ts` picks exactly the first slice starting at or after the wakeup, and the `WHERE w.ts >= COALESCE(s.prev_end_ts, 0)` post-filter rejects the same in-previous-run wakeups the range join never matched. ASOF is an inner join, so wakeups with no later slice drop out as before. `SELECT DISTINCT (ts, utid, cpu)` is the same dedup key as the old `ROW_NUMBER() OVER (PARTITION BY utid, cpu, ts)` / `rn = 1`.

### Suggestions
- `legacy_range_join_runqueue_query()` could return `&'static str` instead of `String` (test-only; `.to_string()` is harmless).
- `fixture_db()` inserts ~2,500 rows one statement at a time; a DuckDB `Appender` would be faster, but the whole test runs in 0.85 s so this is not worth the code.
- The equality test's `events_multiset_query` / `matched_count_query` split the SQL text at CTE names; that couples the test to the CTE naming, which the structure test already pins, so it is acceptable — a future rename of `events`/`rq_depth` will fail loudly, not silently.

### Positive Notes
- The old query is kept verbatim inside the test as the oracle and compared on the `events` multiset (exact) plus the percentiles (≤ 1 depth step, with the tie-order reason written down) — the right two comparisons for a join rewrite whose downstream stage is unchanged.
- The comment explains why (disjoint windows → ASOF is exact) and carries the measured numbers rather than a vague "faster"; no line-number references.
- `trace_id` filters are preserved on both sides (`{filter_ts}` / `{filter_ss}`), and the shape tests were updated rather than deleted.

Gates on this tree (the author's host, prebuilt DuckDB v1.5.4 = the crate's pinned version, `CARGO_BUILD_JOBS=2`, niced): `cargo fmt --all -- --check` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib analyze::cpu_stats` 15 passed (incl. `test_runqueue_query_asof_matches_range_join`); full `cargo test` 563 lib tests + the integration targets passed, 0 failed.

PR #228, opened 2026-08-27
sched: hand each flush to the shared collector in one call instead of locking per record

josefbacik:
Rust-only change in the record collector contract and the sched recorder's flush path; no schema change, no BPF change.

## Why

The sched recorder's ring shards buffer records locally and hand them to the shared streaming collector in batches of up to 10,000 (`STREAMING_SCHED_FLUSH_THRESHOLD`), but `flush_pending` walked each batch record by record through `SharedCollector`, whose forwarding methods (`shared_delegate!`) take the mutex per call. One flush was therefore thousands of lock acquisitions per shard, and since the ring buffers are sharded per CPU (up to 64 consume threads on a big host instead of eight), the shards spend the length of every flush contending on the mutex's futex. On 192-vCPU hosts that run the sched recorder in the continuous lane, about 4–6 % of the recorder's own capture-window CPU, averaged over a mixed cluster, was futex wait/wake spin (`__pv_queued_spin_lock_slowpath` under `futex_q_lock` / `futex_wake`) beneath `SharedCollector::add_sched_slice` / `add_thread_state` ← `flush_pending` ← `consume_loop`; the average hides the tail, which is where the cost sits: on busy 128-vCPU hosts with 64 shards, single 30 s captures showed 21–28 cores of that spin, 85–90 % of the recorder's own CPU for the capture, while hosts with the legacy eight shards read ≈0. The profiler surfaced as a lock-contention subject in its own data.

## What changes

- `RecordCollector::add_sched_batch(&mut self, batch: SchedRecordBatch<'_>) -> usize` — a new trait method taking the seven scheduler buffers at once. The default implementation drains them through the per-record `add_*` methods with the warn-and-count semantics `flush_pending` had (a failed record is warned about and dropped, the batch continues, the failure count is returned), so every other collector behaves exactly as before.
- `SharedCollector` overrides it to lock once and run the inner collector's batch under that single acquisition.
- `flush_pending` builds the batch from its buffers by `&mut` borrow (drained in place, so their capacity survives the flush), skips the call when nothing is pending, and keeps its error contract (`failed to stream N sched/IRQ records`).
- `SchedRecordBatch` is re-exported from `record`.

Scope: the memory, network, stack and probe recorders each stream into their own writer and never shared a lock, so they are untouched; the other `SharedCollector` (the counter recorders) emits a few rows per tick. Allocator growth inside the writer during a flush happens under the lock either way and is not changed by this PR.

## Tests

- `shared_collector_hands_a_sched_batch_to_the_inner_in_one_call`: through a `SharedCollector`, a batch of 1,507 records (1,000 slices, 500 thread states, 7 process exits) reaches the inner collector as exactly one `add_sched_batch` call and zero per-record calls, and the buffers come back drained — the one-acquisition-per-flush property, since the override is a single `lock()` around that call.
- `default_sched_batch_drains_every_buffer_and_counts_failures`: a collector using the default implementation that rejects two of four thread-state records returns 2, delivers the rest in order, and leaves every buffer empty with its capacity intact.
- The existing sched recorder tests (`test_finish_drains_all_buffered_record_types`, `test_finish_closes_running_slices_at_the_gate`, …) pass unchanged.

Gates on this tree (DuckDB v1.5.4, the version `libduckdb-sys` pins): `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` 564 lib tests + the integration targets passed, 0 failed.

### DIFF
--- a/src/record/collector.rs
+++ b/src/record/collector.rs
@@ -155,6 +155,61 @@ pub trait RecordCollector {
     /// Add a TPU runtime metric record.
     fn add_tpu_metric(&mut self, record: TpuMetricRecord) -> Result<()>;
 
+    /// Add a batch of scheduler-lane records, draining every buffer in the
+    /// batch. Returns the number of records that could not be added; each
+    /// failure is warned about and dropped rather than aborting the batch, so
+    /// one bad record cannot silently discard the rest of a buffer.
+    ///
+    /// The default forwards record by record through the `add_*` methods.
+    /// A collector that guards shared state should override it so the whole
+    /// batch goes through one guard acquisition: the sched recorder's ring
+    /// shards flush tens of thousands of records at a time into one
+    /// [`SharedCollector`], and taking its mutex per record turns every flush
+    /// into a futex storm between the shards.
+    fn add_sched_batch(&mut self, batch: SchedRecordBatch<'_>) -> usize {
+        let mut failed = 0usize;
+        let mut warn = |what: &str, e: anyhow::Error| {
+            eprintln!("Warning: Failed to stream {what}: {e}");
+            failed += 1;
+        };
+        for record in batch.slices.drain(..) {
+            if let Err(e) = self.add_sched_slice(record) {
+                warn("sched slice", e);
+            }
+        }
+        for record in batch.thread_states.drain(..) {
+            if let Err(e) = self.add_thread_state(record) {
+                warn("thread state", e);
+            }
+        }
+        for record in batch.irq_slices.drain(..) {
+            if let Err(e) = self.add_irq_slice(record) {
+                warn("IRQ slice", e);
+            }
+        }
+        for record in batch.softirq_slices.drain(..) {
+            if let Err(e) = self.add_softirq_slice(record) {
+                warn("softirq slice", e);
+            }
+        }
+        for record in batch.wakeup_news.drain(..) {
+            if let Err(e) = self.add_wakeup_new(record) {
+                warn("wakeup_new", e);
+            }
+        }
+        for record in batch.sched_migrates.drain(..) {
+            if let Err(e) = self.add_sched_migrate(record) {
+                warn("sched_migrate", e);
+            }
+        }
+        for record in batch.process_exits.drain(..) {
+            if let Err(e) = self.add_process_exit(record) {
+                warn("process_exit", e);
+            }
+        }
+        failed
+    }
+
     /// Flush any buffered records to storage.
     fn flush(&mut self) -> Result<()>;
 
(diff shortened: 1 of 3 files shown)


josefbacik:
rust-bpf-code-reviewer pass — executed in-session over the full diff of this PR (`.claude/agents/rust-bpf-code-reviewer.md` at `65c4535`, read as a tracked blob), before commit/push, per the repo's review-before-commit rule. Mechanics: the authoring session ran the definition's five passes itself; no separate agent process.

### Summary
A collector-contract change with one behavioural goal: the sched recorder's per-shard flush reaches the shared streaming collector through one `add_sched_batch` call (one mutex acquisition) instead of one `add_*` call per record. The trait method has a default that preserves the old per-record semantics for every other collector, `SharedCollector` overrides it to lock once, and `flush_pending` builds the batch from its existing buffers by `&mut` borrow so their capacity survives. No schema, no BPF, no version edit. The diff is small where it matters (≈25 lines of production code in three files) and the rest is tests.

### Critical Issues
None. `SCHEMA_VERSION`, `SCHEMA_CHANGES.md`, `Cargo.toml`/`Cargo.lock` untouched; no recorder table added; no ELF parsing; no `unsafe`; no BPF code. The sched recorder's BPF side and its ring consumers are not changed — only what a flush does with the records it already buffered.

### Improvements
- None required. Checked: the `is_empty()` early return in `flush_pending` preserves the old behaviour for an empty flush (no collector call either way) and additionally avoids an acquisition at `finish()` when nothing is pending. The default `add_sched_batch` drains the buffers in the same order with the same warning text as the removed loop, and its failure count feeds the unchanged `bail!` contract.

### Suggestions
- `forward_all_but_sched!` in the tests repeats the method list that `shared_delegate!` already carries; a shared test-support macro would remove the duplication, but two mocks in one test module do not justify a new module.
- `SchedRecordBatch::len()` is computed from seven `Vec::len()` calls per flush — negligible, noted only so nobody "optimises" it into stored counts.

### Positive Notes
- The test `shared_collector_hands_a_sched_batch_to_the_inner_in_one_call` asserts the property the change exists for (one delegated call, zero per-record calls through the shared handle) rather than an implementation detail, and `default_sched_batch_drains_every_buffer_and_counts_failures` pins the failure semantics, the ordering and the capacity preservation.
- The trait doc says why a guarding collector should override the method (the futex storm between shards) without line-number references; the `flush_pending` doc was rewritten to match the new shape instead of left stale.
- The scope line is honest: recorders with their own writers are named as untouched rather than swept into the PR.

Gates on this tree (the author's host, prebuilt DuckDB v1.5.4 = the crate's pinned version, `CARGO_BUILD_JOBS=2`, niced): `cargo fmt --all -- --check` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib record::collector::tests` + `--lib sched` 37 passed; full `cargo test` result in the PR body.

josefbacik (commented):
Co-read as the owner of the continuous sched lane this recorder feeds — read the full diff at b5f2932 against v1.15.5.

What I checked, and it holds:
- `RecordCollector::add_sched_batch` default = the exact per-record loop `flush_pending` used to run (same warn text, same drop-and-continue, count returned), so every collector that does not override it behaves byte-for-byte as before.
- `SharedCollector` override is a single `lock()` around the inner's batch — one acquisition per flush; the inner's per-record adds run under it exactly as they did, just without 10,000 lock round-trips per flush per shard.
- `flush_pending` keeps its error contract (`failed to stream N sched/IRQ records` on the returned count) and skips the lock entirely when nothing is pending; all seven sched-lane buffers are in the batch (process exits included).
- Tests pin both halves: the shared path delivers 1,507 records as exactly one inner batch call with zero per-record calls, and the default path drains every buffer, keeps order, returns the failure count, and leaves capacity intact.

One observation, not a blocker: the lock is now held for a whole flush (the writer's time for up to 10,000 records) instead of per record, so the other shards wait once per flush rather than churning; at per-shard event rates on large hosts that pause is a few hundred ring entries, far inside a shard's ring. Net effect is the intended one: same total writer time, orders of magnitude fewer acquisitions.

PR #229, opened 2026-08-28
build: compile every BPF object through one helper that owns the flag pin, document the empty-override case, and drop CI's write scopes

josefbacik:
**What this changes.** `build.rs` only: the two BPF object builds now go through one helper, `compile_bpf_object(source, obj_path, object_args)`, which assembles the compiler argument list itself — the `-mcpu=v3 -fwrapv` pin from #222 first, then the object's include paths and defines — applies the `SYSTING_BPF_CLANG` override and runs the compile. `pin_bpf_compiler` is gone. Plus three doc/CI nits. No runtime change and no change to the compiled BPF objects.

**Why.** libbpf-cargo's `clang_args` *replaces* the builder's argument list. After #222 the pin held only because each of the two object builds happened to start its own `Vec` with `BPF_CFLAGS` and call `clang_args` exactly once; a third object, or a second call on the same builder, would have compiled without the pin and nothing would have said so. A first cut of this change moved the prepend into `pin_bpf_compiler` but still handed the builder back to the caller, which could call `clang_args` again — the same hole one step removed. With the helper owning the builder from `new()` to `build()`, the pin is a property of the code shape rather than of the call sites agreeing.

**The nits.** `SYSTING_BPF_CLANG=""` already meant unset (the override filters an empty value) — its doc comment now says so; the README Quick start gains one sentence on the pin and the override; `.github/workflows/ci.yml` gets `permissions: contents: read` (the jobs only build and test; the actions cache does not use the job token's scopes).

**Verification.**
- The BPF objects built from this tree and from the parent (v1.15.5's `build.rs`, same compiler, separate target directories) were compared program section by program section with `llvm-objdump`: identical instruction streams for `systing_system_tmp.bpf.o` (63 program sections), `pystacks.bpf.o` and the linked `systing_system.bpf.o` (63), and identical bytes outside `.BTF`/`.BTF.ext` (the only byte differences are `.BTF` source-path strings, which differ between the two checkouts).
- `cargo fmt --all -- --check` exit 0; `cargo clippy --all-targets -- -D warnings` exit 0; `cargo test` exit 0 — 562 + 28 + 11 + 7 + 4 + 1 + 1 passed, 0 failed (the VM-only targets `#[ignore]`d); CI on this PR runs with the new `permissions` block.

No tag bump is needed (build-side only); it rides the next tag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,11 @@ on:
     pull_request:
       branches:
         - main
+
+# The jobs only build and test; nothing here writes to the repository.
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Test Suite
--- a/README.md
+++ b/README.md
@@ -17,6 +17,11 @@ cargo build
 sudo ./target/debug/systing --duration 60
 ```
 
+The BPF objects are compiled with `-mcpu=v3 -fwrapv` (pinned in `build.rs`)
+so a local build produces the same programs as a release build; set
+`SYSTING_BPF_CLANG=/path/to/clang-21` to also use the release build's
+compiler version (clang 21), otherwise `clang` on `PATH` is used.
+
 Alternatively, install directly with cargo. Use `--locked` so the build uses
 the dependency versions from `Cargo.lock`:
 
--- a/build.rs
+++ b/build.rs
@@ -117,8 +117,8 @@ const BPF_CFLAGS: &[&str] = &["-mcpu=v3", "-fwrapv"];
 
 /// Optional override for the C compiler used for the BPF objects, so a
 /// local or CI build can use the exact compiler the release build uses
-/// (`SYSTING_BPF_CLANG=/path/to/clang-21`). Unset: libbpf-cargo's default
-/// (`clang` on PATH).
+/// (`SYSTING_BPF_CLANG=/path/to/clang-21`). Unset or empty: libbpf-cargo's
+/// default (`clang` on PATH).
 fn bpf_clang_override() -> Option<PathBuf> {
     println!("cargo:rerun-if-env-changed=SYSTING_BPF_CLANG");
     env::var_os("SYSTING_BPF_CLANG")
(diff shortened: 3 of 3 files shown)


josefbacik:
**Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head (`build.rs`, `README.md`, `.github/workflows/ci.yml`); verdict as it produced it.**

### Summary
A build-side change: the two BPF object compiles move into one helper, `compile_bpf_object`, which owns the `SkeletonBuilder` from `new()` to `build()` and assembles the argument list itself (`BPF_CFLAGS` first, then the object's include paths and defines), so no caller can drop or replace the pinned flags. No Rust source, no BPF source, no maps, no programs, no schema, no runtime behaviour change; the README gains one sentence and the CI workflow a `permissions: contents: read` block. The code is simpler than before (the two hand-built chains collapse into one function) and the pin is now structural.

### Critical Issues
None. No schema change (`SCHEMA_CHANGES.md` / `SCHEMA_VERSION` untouched, no `[bump minor]` needed), no `Cargo.toml`/`Cargo.lock` version edit, no recorder, no Python-version, no ELF parsing.

### Improvements
- *(Found on the first cut of this change and resolved before this verdict.)* The first cut prepended `BPF_CFLAGS` inside `pin_bpf_compiler` but returned the builder to the caller, which could still call `clang_args` again and replace the list — the same hole one step removed. Resolved by making the helper own the builder end to end (`compile_bpf_object`); callers pass only their object-specific arguments and never hold a builder that compiles.

### Suggestions
- The two call sites still build near-identical `object_args` vectors (include paths + arch define, with the optional multiarch include pushed afterwards). A small `bpf_include_args(...)` that returns the common prefix would remove the last duplication; not needed for this change and the vectors differ in their pystacks/`-DSYSTING_PYSTACKS` entries, so left as is.
- `bpf_clang_override()` prints `cargo:rerun-if-env-changed=SYSTING_BPF_CLANG` on every call (twice per build, one per object). Harmless — cargo deduplicates the directive — and unchanged from the previous shape.
- `.obj(obj_path)` now takes the `&Path` directly instead of `obj_path.to_str().unwrap()`; the `unwrap` on a non-UTF-8 `OUT_DIR` path is gone with it.

### Positive Notes
- The doc comment on `compile_bpf_object` states the *why* (replacement semantics of `clang_args`) rather than the *what*, and names the property the shape guarantees.
- The panic message names the source file that failed to compile, where the two previous `expect` strings did not distinguish the objects.
- The `SYSTING_BPF_CLANG` doc now matches the code (`filter(|v| !v.is_empty())`), and the README sentence is the first user-facing mention of the pin.

### Gates on the tree (identical bytes to the commit)
`cargo fmt --all -- --check` exit 0; `cargo clippy --all-targets -- -D warnings` exit 0; `cargo test` exit 0 (562 + 28 + 11 + 7 + 4 + 1 + 1 passed, 0 failed; the VM-only targets `#[ignore]`d); BPF object compare against a build of the parent's `build.rs` in a separate target directory: identical instruction streams for all three objects (63 program sections on the two system objects) and identical bytes outside `.BTF`/`.BTF.ext`.

PR #230, opened 2026-08-28
ci: link the tests against the DuckDB version the crate pins, read from Cargo.lock

josefbacik:
**What this changes.** CI only. `.github/workflows/ci.yml`'s two "download pre-built DuckDB" steps (Test Suite, Clippy) no longer hard-code `v1.4.2`: they run the new `scripts/libduckdb-version.sh`, which reads the `libduckdb-sys` version from `Cargo.lock` and decodes it (`1.MMmmPP.0` → `M.m.P`; `1.10504.0` → `1.5.4`), and download that release's `libduckdb-linux-amd64.zip`. The step logs the version it linked. No source change.

**Why.** The release build bundles the DuckDB the `duckdb` crate pins — v1.5.4 today — while CI's `--no-default-features` jobs linked the tests against v1.4.2. A green CI therefore did not exercise engine-semantics-dependent SQL (ASOF JOIN matching, window tie order, quantiles) on the engine that ships; #227's ASOF-vs-range-join equivalence test, for one, ran on 1.4.2. Deriving the version from `Cargo.lock` makes CI follow every crate bump without a second place to edit.

**Verification.**
- `bash scripts/libduckdb-version.sh` on this tree prints `1.5.4` (from `libduckdb-sys 1.10504.0`); the script refuses a lock without the crate or with an unexpected version scheme.
- CI on this PR: the download step's log line reads `libduckdb v1.5.4 (the DuckDB version Cargo.lock pins)` and the Test Suite / Clippy jobs run against it — that log line is the proof the change is meant to produce.

No tag bump needed (CI only).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: "true"
-      - name: Download pre-built DuckDB library
+      - name: Download the pre-built DuckDB library the crate pins
         run: |
-          wget -q https://github.com/duckdb/duckdb/releases/download/v1.4.2/libduckdb-linux-amd64.zip -O /tmp/libduckdb.zip
+          LIBDUCKDB_VERSION="$(scripts/libduckdb-version.sh)"
+          echo "libduckdb v${LIBDUCKDB_VERSION} (the DuckDB version Cargo.lock pins)"
+          wget -q "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
           unzip /tmp/libduckdb.zip -d ${{ github.workspace }}/libduckdb
       - name: Run Tests
         env:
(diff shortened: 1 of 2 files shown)


josefbacik:
**Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head (`.github/workflows/ci.yml`, `scripts/libduckdb-version.sh`); verdict as it produced it.**

### Summary
A CI-only change: both jobs' "download pre-built DuckDB" steps derive the libduckdb release from `Cargo.lock` through a new 26-line helper instead of a hard-coded `v1.4.2`, so the tests and clippy link against the engine the `duckdb` crate pins (v1.5.4 today) and a crate bump moves CI on its own. No Rust, no BPF, no SQL, no schema or `Cargo.toml` change. The helper is small, fails closed on a missing crate or an unexpected version scheme, and the workflow change is the minimum that reads it.

### Critical Issues
None. No schema change, no version edit, no recorder, no Python-version support, no ELF parsing.

### Improvements
None required.

### Suggestions
- `crate_version=$(grep -A1 … | sed -n …)` under `set -euo pipefail`: when the crate is absent, `grep` exits 1, the pipeline fails and `set -e` ends the script at the assignment — exit 1, as intended, but before the "not found" message prints. Appending `|| true` to that pipeline would let the explicit message and exit run.
- If a lock ever carried two `libduckdb-sys` entries the helper would print two lines and the decode would misparse; a `[ "$(printf '%s\n' "$crate_version" | wc -l)" -eq 1 ]` guard (or `head -n 1` with a comment) would make that fail closed too. Not reachable from one `duckdb` dependency.
- The two workflow steps are byte-identical; a composite action would remove the duplication, but two four-line steps are clearer than the indirection at this size.

### Positive Notes
- One source of truth: the version the tests link against is read from the same lock the release build resolves, and the step logs the version it used, so a CI run states which engine it exercised.
- The decode (`1.MMmmPP.0` → `M.m.P`) uses `10#` arithmetic, so a leading-zero minor or patch cannot be read as octal.
- The helper refuses an unexpected version scheme instead of decoding garbage, and the download URL is quoted.

PR #232, opened 2026-08-28
memory recorder: attach the syscall and page-fault hooks as fentry/fexit and raw tracepoints so detach takes milliseconds [bump minor]

josefbacik:
## What

Stopping a capture with the memory recorder on spent hundreds of milliseconds (seconds on 192-vCPU hosts) inside the `detach bpf programs` stop phase. The six `syscalls/sys_{enter,exit}_{mmap,munmap,brk}` tracepoints and the x86 `exceptions/page_fault_user` tracepoint were attached through `perf_event_open`, and closing a perf-attached tracepoint runs `perf_trace_event_unreg()` → `tracepoint_synchronize_unregister()` — one SRCU and one RCU grace period per program, serialized (kernel/trace/trace_event_perf.c and include/linux/tracepoint.h; the same at v6.6, v6.12 and v6.18). The syscalls tracepoints also kept every task of the host on the syscall-entry slow path (`SYSCALL_WORK_SYSCALL_TRACEPOINT`) for the whole capture.

This PR:

- attaches the six syscall hooks as **fentry/fexit trampolines on the arch syscall wrappers** (`__x64_sys_*`, `__arm64_sys_*`, `__riscv_sys_*`; the compat `__ia32_sys_*` wrappers are separate symbols and stay unattached, so native-only semantics hold as they did at the tracepoints), reading the arguments with `PT_REGS_PARMn_CORE_SYSCALL(regs)` and the return from the fexit. A trampoline call replaces the perf-sample build per event, puts nothing on other syscalls, and unlinks through `bpf_trampoline_update()`'s text poke and asynchronous RCU-tasks frees (kernel/bpf/trampoline.c).
- attaches the x86 page-fault tracepoint **raw** (`tp_btf/page_fault_user`), whose detach is `tracepoint_probe_unregister()`'s asynchronous `call_rcu` path (kernel/tracepoint.c).
- keeps the **classic tracepoint set in the object as the fallback, chosen at attach time**: both sets are loaded (the verifier checks both forms at every load), both are excluded from `skel.attach()`, and the trampoline programs are attached one by one afterwards; the first attach error drops the partial links and attaches the classic six. That is the failure the load-time gates cannot see — an arm64 kernel without `DYNAMIC_FTRACE_WITH_DIRECT_CALLS` resolves the BTF target, loads the program and refuses the link. A kernel whose kallsyms lacks the wrappers (riscv before 6.7), or whose vmlinux BTF has no FUNC entry for them, never loads the trampoline set (an fentry program's target is resolved at load; an unresolvable one fails the whole object).
- moves the network recorder's three TIME_WAIT hooks (`tcp_time_wait`, `inet_twsk_hashdance_schedule`, `inet_twsk_deschedule_put`) from kprobes to fentry trampolines the same way, with the kprobe trio as the fallback through the same helper (`attach_leg_with_fallback`), and **gates the leg on the kernel before the object loads**: a kernel without `inet_twsk_hashdance_schedule` (pre-6.11, kernel commit b334b924c9b7) runs its capture without TIME_WAIT tracking instead of failing every attach, CPU profile included.
- records which form ran in `sysinfo`: `memory_syscall_leg` = `fentry` | `tracepoint:nosym` | `tracepoint:nobtf` | `tracepoint:notramp` | `off:<cause>`, and `network_tw_leg` = `fentry` | `kprobe:notramp` | `kprobe:nobtf` | `off:<cause>` — **schema version 20** (`[bump minor]`; both columns nullable, NULL when the recorder did not run and in older traces). One read of kallsyms and vmlinux BTF per capture serves both lanes (`probe_kernel_symbols`).

The rows of `memory_map` and `network_packet` are unchanged in value: the trampoline and classic programs share one body per hook. `memory_fault_leg` keeps reading `tracepoint` on x86 for the raw-attached tracepoint (same event, same rows).

## Schema skew (both faces)

- Forward (a reader at this version importing an older `sysinfo.parquet`): the two columns read NULL — covered by the parquet writer's import-without-the-columns test.
- Reverse (an older reader importing a `sysinfo.parquet` written at this version): `parquet_to_duckdb` imports with `INSERT INTO … BY NAME`, which requires every source column to exist in the target, so an older reader fails the whole import on the first unknown column — the same face every column add since schema 14 has had. Readers that import parquet written by newer recorders need to be at this version first.

## arm64 note

The attach-time fallback exists for arm64: a trampoline needs direct-call ftrace, which a BTF/kallsyms gate cannot see, and the VM rig has no arm64 kernel. On such a host the two columns read `tracepoint:notramp` / `kprobe:notramp` — correct outcomes (the classic sets carry the rows); `off:*` would be the finding. A first read of the two columns per architecture is the check.

## Tests

- `bpf_load_shapes`: the shape table gains `memory-syscall-tracepoint-only` and `network-tw-kprobe-only`, so every new program is verified by some shape; `every_shape_loads` passes on the VM rig's 6.12 kernel (32 shapes) with the object compiled by clang 21 at the pinned flags.
- `tests/memory_recorder.rs`: the e2e asserts the recorded leg (`fentry` on the rig); new `test_memory_syscall_hooks_fall_back_to_tracepoints` forces the trampoline attach to fail through a testing-only `Config` switch (never a CLI flag) after the first real link, and asserts `tracepoint:notramp` and the classic set's mmap/munmap rows. 10/10 on the rig.
- `tests/trace_validation.rs`: the network suite asserts `network_tw_leg`; new `test_e2e_network_tw_leg_kprobe_fallback` exercises the kprobe fallback the same way (`kprobe:notramp` + TIME_WAIT rows through the kprobe set). 2/2 on the rig; the guest recorded `TIME_WAIT leg fentry` on the default run.
- Unit: leg selection and sysinfo values for both lanes, the vmlinux BTF probe, the sysinfo round trip; the whole lib suite (575) green; clippy `-D warnings`, rustfmt.
- Verifier read of the object compiled with the pinned flags: every new section's map-value accesses are constant-offset; the maps the new programs write are a preallocated HASH, per-CPU arrays and ringbufs (no LRU hash in the object); the NMI-context `perf_event` programs are untouched.

## Positive control

`detach bpf programs` stop-phase p50 on a many-core host with the memory recorder on: ≈459 ms (p99 618 ms) before → expected single-digit ms; the `attach bpf programs` line now includes the trampolines' text pokes, before the clock sampler starts. Per-tick missed stack/memory event shares should be unchanged.


josefbacik:
### Summary
The change moves the memory recorder's six mmap/munmap/brk syscall hooks from perf-attached `syscalls/` tracepoints to fentry/fexit trampolines on the arch syscall wrappers (classic set kept as an attach-time fallback), attaches the x86 page-fault tracepoint raw (`tp_btf`), and makes the network recorder's three TIME_WAIT kprobes fentry trampolines with the kprobe trio as fallback — so `detach bpf programs` no longer serializes synchronous grace periods. Both lanes share one attach helper (`attach_leg_with_fallback`), one kernel probe (`probe_kernel_symbols`: kallsyms + vmlinux BTF, read once per capture), and one schema bump (v20: `sysinfo.memory_syscall_leg`, `sysinfo.network_tw_leg`). Read over the full diff (15 files, +1621/−171) against v1.16.3; the BPF bodies are shared between the two forms of every hook, so the rows are unchanged by construction.

### Critical Issues
None found.
- Schema: `SCHEMA_CHANGES.md` carries the v20 entry (added columns, changed meaning, behaviour change); `SCHEMA_VERSION` 19 → 20 in `src/duckdb.rs`; the title carries `[bump minor]`; `Cargo.toml`/`Cargo.lock` `version` untouched.
- Recorder consistency: no new per-event table; the two new columns are on `sysinfo` (one row per trace, no thread attribution); every positional `INSERT INTO sysinfo` fixture updated (`profile_export.rs` +2 NULLs) and the name-keyed chain (`models.rs` → `schema.rs` → `parquet/writer.rs` round trip → `duckdb.rs` DDL → `analyze/mod.rs` SELECT + the legacy-DB test's DROPs → `mcp.rs` description) is complete.
- BPF verifier class: every new program section reads its map values at constant offsets (no runtime-indexed access; objdump of the release-flag object); the maps the new programs write are a preallocated HASH, PERCPU_ARRAYs and ringbufs — no LRU_HASH in the object — and the two `perf_event` (NMI) programs are untouched and write ringbufs only.
- No ELF/binary parsing added.

### Improvements (applied before the commit)
- One kernel read for both lanes: `probe_kernel_symbols(opts)` reads kallsyms once and parses vmlinux BTF once (only for trampoline targets kallsyms lists), and both `probe_memory_kernel_legs` and `probe_network_kernel_legs` decide from it — the first draft read kallsyms twice and BTF per lane.
- The memory syscall leg gained the BTF gate the network leg had (`tracepoint:nobtf`): an fentry program whose target is absent from vmlinux BTF fails the whole object at load, which an attach-time fallback cannot catch; the leg is left unloaded there instead.
- The classic-form bodies were deduplicated into `memory_{mmap,munmap,brk}_{enter,exit}_common` and the TIME_WAIT hooks into one `*_common` body each, so the trampoline and classic entry points are thin wrappers that only differ in how the arguments are fetched (`PT_REGS_PARMn_CORE_SYSCALL(regs)` / `ctx->args[n]`, the fexit `ret` / `ctx->ret`).
- The synthesized attach error of the testing switch names the switch generically (`forced fallback (testing switch)`) since both lanes share the helper.
- The e2e assertions accept `tracepoint:nobtf` beside `fentry` (and beside `tracepoint:notramp` under the switch) on a kernel that lists the wrappers, so a BTF-less rig kernel would not fail the suite for the wrong reason; the rig kernel (6.12, BTF on) recorded `fentry` / `tracepoint:notramp`.
- Folded from the v1.16.0 cold verify of the previous memory-recorder PR: the SCHEMA_CHANGES headings carry their release versions (v14 → 1.12.0 … v19 → 1.16.0); the `freeze` flag's documented meaning corrected (migrate path only, reclaim's unmap passes `false`); the PMD-split worker probe's comment notes it fires inside the page-table lock and takes the `bool` argument as a raw register tested on its low byte.

### Suggestions
- `attach_program_set` walks `progs_mut()` per name (O(programs × object) for six names) — negligible at attach time; a name-indexed map would only matter with far more programs.
- The `tracepoint:notramp` value is shared by a real trampoline refusal and the testing switch's synthesized failure; the stderr line carries the cause. If a future consumer needs to tell them apart, `forced` is already the internal cause string and could become its own sysinfo value.
- The vmlinux BTF parse (one `Btf::from_vmlinux()` per capture when a leg is on) is ~5 MB of reads; acceptable on a 10 s cadence, but worth caching across captures if a long-lived caller ever runs sub-second captures.

### Positive Notes
- The fallback is decided at attach time, where the failure actually happens (arm64 without direct-call ftrace resolves the BTF id, loads, and refuses the link), and both program sets are loaded on every memory capture, so the verifier checks both forms at every load and the load-shape gate exercises them.
- The shared `LegAttach` enum makes the three outcomes (primary / fallback with the primary's error / neither) explicit for both lanes, and partial links are dropped on the error path so a set is attached as a unit or not at all.
- The network leg's kallsyms gate turns a pre-6.11 kernel's whole-capture failure into a capture without TIME_WAIT tracking, recorded in `sysinfo`.

### Review mechanics
Rubric: the repository's `rust-bpf-code-reviewer` agent definition, executed by the author in-session over the full diff at commit time (three passes: Rust, BPF, schema/versioning), with the objdump read of the release-flag object and the map-writer census done beside it. Verifier evidence: the 32-shape load gate, the memory recorder e2e suite (10 tests) and the network e2e suite (2 tests) on the VM rig's 6.12 kernel with the release BPF compiler; the whole lib suite (575 tests); clippy `-D warnings`; rustfmt.

PR #234, opened 2026-08-28
stack_recorder: bound and deduplicate the build-id store's host-wide index walk

### LINKED ISSUE #233: build-id mode: bound and dedupe the host-wide /proc build-id index walk (stop phase 2.4 s → 21.6 s per capture on 192-vCPU many-container hosts)
## Summary

In build-id mode (`--collect-build-id`), the build-id store's "opportunistic fill" walks **every live process on the host** at symbolization time and reads the ELF build-id note of **every executable file-backed mapping of every process** (`stack_recorder.rs` ≈1258–1280 → `fill_store_from_maps` → `blazesym::helper::read_elf_build_id` through the `map_files` link). Nothing dedupes across processes: the same `libc.so.6` mapped by 3,000 processes is opened and parsed 3,000 times, and the walk has no bound in files, bytes or time.

On small hosts this is cheap (a 50/50 A/B on a class with ~few hundred processes per host reads ≈ +1 s per 10 s capture). On a 192-vCPU host class running many short-lived containers the same A/B reads, per 10 s continuous capture:

| | control | build-id |
|---|---|---|
| capture wall time, avg | 16.3 s | **35.3 s** |
| capture wall time, p99 | 28.8 s | **176 s** |
| `stop-phase: symbolize and write parquet`, avg | 2.4 s | **21.6 s** |
| stop phase, p50 / p90 / p99 | 2.1 / 3.9 / 9.8 s | **15.8 / 42 / 153 s** |
| captures over 60 s | 0.3 % | **8.8 %** |

The whole excess sits in the stop phase, where the fill walk runs; it does not warm up (the 2nd and 3rd captures on a host read the same as the 1st), and it costs no samples (rows per capture, failures, ring overflow and RSS are at parity). It is purely the per-capture host-wide read.

## Why it matters

A 10 s capture whose tear-down takes 20 s–3 min keeps the tracer resident and reading `/proc` and container images for most of the interval between captures. The earlier exited-process symbolization pass (#165) was reverted for exactly this class of per-capture, host-wide binary scanning; the build-id walk that replaced it (#173) carries a host-wide walk of its own, in userspace, with no bound.

## Proposed fix

1. **Dedupe across processes** by the mapping's `(dev, inode)` from the maps line (`MapEntry` does not carry them today; `exec_file_links` dedupes by path within one process only). One ELF note read per distinct file on the host instead of per process per mapping.
2. **Bound the walk**: a cap on distinct files read and on wall time (defaults generous; CLI-overridable), walking the **sampled** processes' mappings first (the frames we actually hold) and the rest of `/proc` only under the remaining budget. A frame whose binary was not indexed already falls to the existing `.build-id` directory / debuginfod / deferred `[buildid:…]` paths, so exhausting the budget degrades names, never samples.
3. **Report it** on the stop-phase line family (stdout, which the supervising agent forwards) — distinct files read, mappings skipped by the dedupe, whether the budget was hit and the elapsed time — so the cost is observable per capture instead of only as capture wall time.

Tests: the dedupe (two processes sharing a file → one read), the budget (cap of one → the second file skipped and reported), and the order (sampled tgids indexed first).

josefbacik:
## Why

With `--collect-build-id`, end-of-trace symbolization fills the build-id store from the binaries of still-running processes: it walks every live process on the host and reads the ELF build-id note of every executable file-backed mapping of every process, through the `map_files` link. Nothing deduplicated that across processes — the same `libc.so.6` mapped by 3,000 processes was opened and parsed 3,000 times — and nothing bounded it in files, bytes or time. On a host with a few hundred processes that is about a second; on a 192-vCPU host running thousands of short-lived containers a 50/50 A/B on continuous 10 s captures measured, per capture:

| | control | build-id |
|---|---|---|
| capture wall time, avg | 16.3 s | **35.3 s** |
| capture wall time, p99 | 28.8 s | **176 s** |
| `stop-phase: symbolize and write parquet`, avg | 2.4 s | **21.6 s** |
| stop phase, p50 / p90 / p99 | 2.1 / 3.9 / 9.8 s | **15.8 / 42 / 153 s** |
| captures over 60 s | 0.3 % | **8.8 %** |

The whole excess sits in the stop phase where the walk runs, it does not warm up (the 2nd and 3rd captures on a host read the same as the 1st), and it costs no samples — rows per capture, failures, ring overflow and RSS were at parity. A 10 s capture whose tear-down takes 20 s to 3 min keeps the tracer resident and reading `/proc` and container images for most of the interval to the next capture. The earlier exited-process symbolization pass (#165) was reverted for this class of per-capture, host-wide binary scanning; the build-id walk that replaced it (#173) carried an unbounded host-wide walk of its own.

## What

- `sandbox_maps`: `MapEntry` keeps the maps line's `dev`/`inode` as a `FileId` (the path cannot identify a file across mount namespaces; dev/inode can), and `exec_file_links_with_ids` returns it beside the link and display path (`exec_file_links` keeps its signature).
- `stack_recorder`: the walk becomes `BuildIdIndexWalk` — one ELF note read per distinct file host-wide, a budget of distinct files (default 10,000) and wall time (default 2,000 ms; `0` = unbounded for either), and a stats record. A binary the walk never reaches still resolves through the `.build-id` directories, debuginfod, or the deferred `[buildid:…]` rendering, so a hit budget costs names, never samples.
- The walk's cost is printed on stdout with the other per-phase cost lines — `build-id index: N files indexed, M without a build-id note, K mappings deduplicated, J mappings across P processes, in Tms (… budget hit …)` — instead of the stderr `Note:` line, so it is visible per capture wherever the stop-phase lines are.
- `--build-id-index-max-files` / `--build-id-index-max-ms` on the CLI and `Config`; README updated.

## Tests

`test_maps_line_carries_the_file_identity` (dev/inode parsing, the same file at two paths, inode 0, a malformed dev field), `test_build_id_fill_dedupes_files_across_processes`, `test_build_id_fill_stops_at_the_file_budget`, `test_build_id_fill_stops_at_the_time_budget`, `test_build_id_index_summary_line`. The fixtures' reads fail by construction (no such processes, no such files), so they pin the dedupe and budget decisions, which happen before the read.

No BPF source is touched; no schema change; no version edit.

Fixes #233.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,14 @@ symbolization still resolve — through a build-id-keyed store filled from
 binaries of still-running processes. Frames whose build-id no source knows
 render as `unknown ([buildid:<hex>]) <0x<offset>>`, a stable identity that can
 be resolved offline later. Costs ~40% larger stack-ring reservations while
-enabled; off by default and free when off.
+enabled; off by default and free when off. The still-running-processes source
+is an end-of-trace walk over every live process's executable mappings that
+reads each distinct file's build-id note once (deduplicated by dev/inode across
+processes) and is bounded by `--build-id-index-max-files` (default 10000) and
+`--build-id-index-max-ms` (default 2000; 0 = unbounded for either); it prints a
+`build-id index: …` line with what it read and whether a bound stopped it. A
+binary the walk never reaches still resolves through the other sources or
+keeps its `[buildid:<hex>]` identity.
 
 Note: `sleep-stacks` acts as a master switch for all sleep stack collection. When enabled,
 both uninterruptible (D state) and interruptible (S state) sleep stacks are collected by
--- a/src/main.rs
+++ b/src/main.rs
@@ -71,6 +71,17 @@ struct Command {
     /// "unknown ([buildid:<hex>]) <0x<offset>>" and stay resolvable offline
     #[arg(long)]
     collect_build_id: bool,
+    /// With --collect-build-id: the most distinct files the end-of-trace
+    /// build-id index may read the ELF note of (the walk over every live
+    /// process's executable mappings); 0 = unbounded. A binary the walk
+    /// never reaches still resolves through the .build-id directories,
+    /// debuginfod, or the deferred [buildid:...] rendering
+    #[arg(long, default_value_t = systing::stack_recorder::DEFAULT_BUILD_ID_INDEX_MAX_FILES)]
+    build_id_index_max_files: usize,
+    /// With --collect-build-id: the longest the end-of-trace build-id index
+    /// walk may run, in milliseconds; 0 = unbounded
+    #[arg(long, default_value_t = systing::stack_recorder::DEFAULT_BUILD_ID_INDEX_MAX_MS)]
+    build_id_index_max_ms: u64,
     /// Enable debug output for pystacks (Python stack tracing)
     #[arg(long)]
     pystacks_debug: bool,
@@ -251,6 +262,8 @@ impl From<Command> for Config {
             continuous: cmd.continuous,
             collect_pystacks: cmd.collect_pystacks,
             collect_build_id: cmd.collect_build_id,
+            build_id_index_max_files: cmd.build_id_index_max_files,
+            build_id_index_max_ms: cmd.build_id_index_max_ms,
             pystacks_pids: Vec::new(), // CLI doesn't expose this yet, uses discovery
             pystacks_debug: cmd.pystacks_debug,
             enable_debuginfod: cmd.enable_debuginfod,
--- a/src/sandbox_maps.rs
+++ b/src/sandbox_maps.rs
@@ -55,6 +55,37 @@ struct MapEntry {
     /// and meaningless for ELF math; file offsets are real).
     offset: u64,
     backing: Backing,
+    /// The maps line's `dev` and `inode` fields, which together identify the
+    /// backing file across every process that maps it (the path cannot: it
+    /// is namespace-relative, and one path can name different files in
+    /// different mount namespaces). `None` for anonymous entries and for
+    /// lines whose fields do not parse.
+    file_id: Option<FileId>,
+}
+
+/// Identity of a mapped file host-wide: the maps line's `(major:minor, inode)`.
+/// Two mappings with the same `FileId` are the same file, whatever process
+/// or mount namespace they were read from.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct FileId {
+    pub dev_major: u32,
+    pub dev_minor: u32,
+    pub inode: u64,
+}
+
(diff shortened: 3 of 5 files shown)


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head; verdict as it produced it.

**Structure.** The build-id store's "opportunistic fill" (build-id mode, end of trace) walked every live process and read the ELF note of every executable mapping of every process, with no cross-process deduplication and no bound. The change gives `MapEntry` the maps line's dev/inode as a `FileId`, wraps the walk in `BuildIdIndexWalk` (one ELF read per distinct file, a files + wall-time budget, a stats record), and prints the walk's cost on the stop-phase line family. Two CLI flags carry the budget; the defaults are generous and `0` means unbounded. No BPF source, no schema, no `Cargo.toml` version edit.

**Rust quality.**
- `BuildIdIndexWalk` owns the walk state (store, budget, start time, seen set, stats) — the earlier six-argument free function is gone; `fill_from` / `over_budget` / `run` each do one thing.
- `file_id.is_some_and(|fid| !self.seen.insert(fid))` is the idiomatic dedupe; a mapping with no parsable dev/inode is treated as unique rather than dropped.
- `exec_file_links` keeps its signature for its other caller and delegates to `exec_file_links_with_ids`; the one extra allocation is per process, not per mapping.
- Error handling unchanged on the read path (`read_elf_build_id` failures are counted, never propagated — a racing exit is a miss, as before).
- No `unsafe`, no new dependencies.
- Comments explain why (host-wide walk, why stdout, why the defaults), none reference line numbers.

**BPF quality.** No BPF code in the diff.

**Schema & versioning.** No schema change, `SCHEMA_VERSION` untouched, `Cargo.toml`/`Cargo.lock` untouched, no `[bump minor]` needed.

**Integration.** `systing_core` sets the budget beside `set_build_id_mode`; the `Config` defaults equal the CLI defaults (pinned by `test_build_id_index_summary_line`). The summary line moves from stderr (`Note: …`, not forwarded by supervising agents) to stdout, where the `stop-phase:` lines already are.

**Critical issues:** none.

1. `over_budget` reads `Instant::now()` once per non-deduplicated mapping; if a profile ever shows it, check the clock once per process in `run` instead (the per-mapping check is what makes the time bound tight on a process with thousands of mappings, so it is kept deliberately).
2. The walk still calls `ProcessMaps::load` (a `/proc/<pid>/maps` read + `readlink` of `exe`) for every process before the budget can stop it at the next process; a future change could stat the process count first and shrink the per-process work, but the time budget bounds this today.
3. Tests pin the dedupe, both budgets and the summary text on fixtures whose reads fail by construction; a fixture with a real ELF note (the test binary itself, where the toolchain emits one) would additionally pin `files_indexed`.

PR #235, opened 2026-08-28
build-side follow-ups: libduckdb-version.sh's refusals, and only -I/-D as BPF object arguments

josefbacik:
Two build-side follow-ups from the cold reads of #230 and #231; no source, BPF, schema or version change.

## `scripts/libduckdb-version.sh` — make its refusals real (#230)

The helper decodes the DuckDB version CI downloads from `Cargo.lock`'s `libduckdb-sys` entry. Two of the refusals it advertises did not fire:

- With `set -euo pipefail`, a lock without the crate made the `grep` pipeline fail inside the assignment, so the script died with rc 1 and an empty stderr before the "libduckdb-sys not found" branch could run.
- The scheme guard `1.*)` admitted the crate's older `1.x` line, which is still published in parallel (`1.4.5` beside `1.10504.0`): `1.4.5` decoded to `0.0.4` with rc 0 and the download step then 404'd silently under `wget -q`.

A pre-release suffix (`1.10504.0-alpha.1`) was also read as its release, and a lock naming the crate twice took the first entry.

Now: the extraction tolerates an empty match (`|| true`) and tests the count (none → "not found"; more than one → "appears N times (…); cannot pick one"); the scheme guard is the anchored encoded shape `^1\.([0-9]{5,})\.[0-9]+$`, which admits `1.10504.0` (→ 1.5.4) and `1.11000.0` (→ 1.10.0) and refuses `1.4.5`, `0.10.2`, `2.20000.0` and any pre-release suffix, naming the version and the expected form; the two libduckdb downloads in `ci.yml` use `wget -nv`, so a failed download prints its reason in the job log.

Checked on eleven synthetic locks (the real encoding, a two-digit minor, a crate patch, a pre-release, the old `1.4.x` line twice, `0.x`, `2.x`, a missing crate, two duplicate-entry locks) and on the repository's own `Cargo.lock` (`1.5.4`); the CI step shape (`bash -e`, the version in a command substitution) aborts with the message on every refusal and prints `reached: libduckdb v1.5.4` on the real lock.

## `build.rs` — admit only include paths and defines as BPF object arguments (#231)

The refusal added in #231 named three spellings that override the pinned flags (`-mcpu`, `-fwrapv`, `-fno-wrapv`), but the driver has more: on a newer clang, `-fstrict-overflow` after the pin drops `-fwrapv` from the cc1 line (an older clang keeps it — exactly the local-versus-release divergence the pin exists to close), `-ftrapv` beside it makes cc1 trap instead of wrap, `-fno-strict-overflow` adds `-fwrapv-pointer`, and `-Xclang` or an `@response-file` reach cc1 around any list. Both callers pass only include paths and defines, so the guard is inverted: an object argument must start with `-I` or `-D`, anything else is refused with the argument, the source and the pinned set named. The command line is unchanged; the compiled objects were compared against a build of the unchanged `build.rs` — identical instruction streams and identical bytes outside `.BTF`/`.BTF.ext`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           LIBDUCKDB_VERSION="$(scripts/libduckdb-version.sh)"
           echo "libduckdb v${LIBDUCKDB_VERSION} (the DuckDB version Cargo.lock pins)"
-          wget -q "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
+          wget -nv "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
           unzip /tmp/libduckdb.zip -d ${{ github.workspace }}/libduckdb
       - name: Run Tests
         env:
@@ -81,7 +81,7 @@ jobs:
         run: |
           LIBDUCKDB_VERSION="$(scripts/libduckdb-version.sh)"
           echo "libduckdb v${LIBDUCKDB_VERSION} (the DuckDB version Cargo.lock pins)"
-          wget -q "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
+          wget -nv "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
           unzip /tmp/libduckdb.zip -d ${{ github.workspace }}/libduckdb
       - name: Run Clippy
         env:
--- a/build.rs
+++ b/build.rs
@@ -135,14 +135,21 @@ fn bpf_clang_override() -> Option<PathBuf> {
 /// caller that assembled its own list (or called `clang_args` twice) would
 /// silently compile that object without the pin. Callers pass only their
 /// object-specific arguments and never hold the builder, which is what makes
-/// the pin structural. Because clang lets a later flag win, an object argument
-/// that would override a pinned flag is refused rather than silently applied.
+/// the pin structural. Because clang lets a later flag win — and its driver
+/// has more spellings that move a pinned flag than the pin itself has
+/// (`-fstrict-overflow` drops `-fwrapv` from the cc1 line on newer compilers,
+/// `-ftrapv` beside it makes cc1 trap instead of wrap, `-Xclang` and an
+/// `@response-file` reach cc1 around any list) — an object argument is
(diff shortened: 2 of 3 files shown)


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head (two commits); verdict as it produced it.

**Structure.** Two independent build-side fixes. (1) `scripts/libduckdb-version.sh` (from #230): two advertised refusals did not work — the "crate not found" branch was dead under `set -euo pipefail` (the empty `grep` aborted the assignment before the `if`), and the scheme guard `1.*` admitted the crate's older `1.x` line, which decoded to a version that does not exist and then failed only as a silent `wget -q` 404. The change makes both refusals real, adds the exactly-one-entry check, refuses a pre-release suffix instead of silently reading it as the release, and switches the two libduckdb downloads to `wget -nv`. (2) `build.rs`: the object-argument refusal from #231 enumerated three driver spellings; the driver has more that move the pinned flags, so the guard is inverted to admit only the two forms the callers use (`-I…`, `-D…`).

**Rust quality (`build.rs`).**
- One predicate replaces a growing deny-list: `arg.starts_with("-I") || arg.starts_with("-D")` is exact for what both call sites pass (every argument is a `format!("-I…")` or a `-D…` literal), and a future caller passing anything else fails at build time with the argument, the source and the pinned set named.
- The command line is unchanged — the pin is still prepended, the object arguments still follow — so the objdump identity proof against a build of the unchanged `build.rs` holds (identical instruction streams; identical bytes outside `.BTF`/`.BTF.ext`).
- The doc comment says why the set is not enumerated (the `-fstrict-overflow` / `-ftrapv` / `-Xclang` / response-file spellings) and what the guard does not cover (the compiler's own environment) — the why, not the what; no line-number references.

**Shell quality (`libduckdb-version.sh`).** `|| true` is the standard idiom for "empty is a value" under `pipefail`; the count is then tested explicitly; the anchored `[[ … =~ ^1\.([0-9]{5,})\.[0-9]+$ ]]` both guards the scheme and captures the encoded group, so the two parameter-expansion strips go away; the header comment states what is refused and why.

**BPF quality.** No BPF source in the diff; the objects are proven unchanged.

**Schema & versioning.** No schema change, `SCHEMA_VERSION` untouched, `Cargo.toml`/`Cargo.lock` untouched, no `[bump minor]` needed.

**Integration.** The CI steps consume the script unchanged (`LIBDUCKDB_VERSION="$(scripts/libduckdb-version.sh)"` under `bash -e`): a refusal now aborts the step with its message instead of proceeding to a doomed download — verified on eleven synthetic locks plus the repository's own (`1.5.4`). The `build.rs` guard is exercised by every build (both objects compile through it).

**Critical issues:** none.

1. The helper could carry a `--check` mode that verifies the release asset exists (a `HEAD` request) so a wrong encoded version fails before the unzip; today the `-nv` download's 404 is that check.
2. A unit test for the guard is not possible from `build.rs` itself; the object compare in this repository's review practice is the proof that a guard change leaves the objects alone.

PR #236, opened 2026-08-28
stack_recorder: index the build-id store across mounts, sampled first, within the store's capacity

josefbacik:
## Why

The end-of-trace build-id index walk from #234 (issue #233) bounded the walk's cost, but its coverage was wrong-ended on the host class that motivated it:

- **The dedupe key stopped at the mount.** The walk recognised a file across processes by the maps line's `(dev, inode)`. Since Linux 6.8 (`show_map_vma` reporting the user-visible inode) that line shows the overlay superblock's device for a container's files, so one image's `libc.so.6` in N containers is N distinct keys: the same note was read N times, `deduplicated` read low, `files indexed` overstated, and on a host running thousands of containers the 10,000-file / 2 s budgets tripped on essentially every capture — steady state, not an anomaly.
- **The walk order dropped the wrong end.** `/proc` in ascending pid order is oldest first, so a budget trip skipped the newest processes — the containers most likely to share an image with the just-exited processes whose frames only the store can name.
- **A large or zero file budget silently truncated.** `--build-id-index-max-files 0` (documented as unbounded) or a value above the store's capacity (16,384) kept reading while the store's LRU eviction discarded what had just been read, and the line printed no budget marker.

## What

- **A second dedupe tier.** On a `(dev, inode)` miss, one `stat` of the mapping's `map_files` link (falling back to the mapped path) keys the file by `(mapped path, size, mtime to the nanosecond)`; overlay passes the lower file's size and mtime through unchanged, so the same image's file collapses across every container, while two different files at one in-container path are kept apart unless they also share size and mtime. A false merge skips the second file's note read, so its frames keep their own build-id and find no entry in the store — a name lost, never a wrong one: the store is keyed by the note actually read, and a frame looks up its own id. The note is read only on a miss of both tiers; a `stat` failure (a process gone) falls through to the read, which fails the same way and is counted.
- **Sampled processes first, then newest first.** Each interner records the tgids it interned stacks for (on the miss path — once per unique stack, one small-set insert, not per event); at symbolization time the walk visits those first, then every other pid in descending order, so a budget trip drops the oldest, longest-lived processes. The doc comment says what a trip costs: names for the binaries of the processes not reached, never samples.
- **The file budget is clamped to the store's capacity** (`BuildIdStore::capacity()`, 16,384): `0` or anything larger means the capacity, and a walk stopped there says `file budget hit at the store's capacity of 16384` on its line.
- **The summary line** now reads `… N mappings deduplicated (M across mounts) …` so the second tier's effect is visible per capture, and says "without a readable build-id note" (unreadable files were always counted there).
- **Wording:** the time budget is checked between reads (a read in progress is not interrupted) — README, flag help and the `Config` doc say so.

## Tests

- `test_build_id_fill_dedupes_the_same_file_across_mounts`: one real file mapped by two fake processes at the same path with different devices → one read, one across-mount dedupe.
- `test_build_id_fill_reads_a_changed_file_at_the_same_path_again` (the negative control): the same path and size with a different mtime → two reads, nothing deduplicated.
- `test_build_id_index_walk_order`: sampled pids first, then the rest newest first.
- `test_build_id_fill_clamps_the_file_budget_to_the_store`: 0 / capacity + 1 / `usize::MAX` → the capacity; exactly the capacity is not a clamp; the line names it.
- `test_build_id_index_summary_line` updated for the new line; `test_finish_inner_build_id_dead_frame_resolves_via_live_fill` pins an unbounded budget instead of depending on the default against the host's process count.

What the tests cannot show is a real host's overlay behaviour itself — the first check on a many-container host is the `build-id index:` line: the across-mounts count high, and no budget marker at steady state.

No BPF, schema or version change; no flag added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,13 +148,18 @@ binaries of still-running processes. Frames whose build-id no source knows
 render as `unknown ([buildid:<hex>]) <0x<offset>>`, a stable identity that can
 be resolved offline later. Costs ~40% larger stack-ring reservations while
 enabled; off by default and free when off. The still-running-processes source
-is an end-of-trace walk over every live process's executable mappings that
-reads each distinct file's build-id note once (deduplicated by dev/inode across
-processes) and is bounded by `--build-id-index-max-files` (default 10000) and
-`--build-id-index-max-ms` (default 2000; 0 = unbounded for either); it prints a
-`build-id index: …` line with what it read and whether a bound stopped it. A
-binary the walk never reaches still resolves through the other sources or
-keeps its `[buildid:<hex>]` identity.
+is an end-of-trace walk over every live process's executable mappings — the
+sampled processes first, then the rest newest first — that reads each distinct
+file's build-id note once: a file is recognised across processes by the maps
+line's dev/inode and, across mounts (one container image's file shows a
+different device in every container), by its mapped path, size and mtime from
+one stat. The walk is bounded by `--build-id-index-max-files` (default 10000;
+0, or more than the store keeps, means the store's capacity of 16384) and
+`--build-id-index-max-ms` (default 2000, checked between file reads; 0 =
+unbounded); it prints a `build-id index: …` line with what it read, how much
+each tier deduplicated, and whether a bound stopped it. A binary the walk never
+reaches — a bound stops it at the oldest processes — still resolves through the
+other sources or keeps its `[buildid:<hex>]` identity.
 
 Note: `sleep-stacks` acts as a master switch for all sleep stack collection. When enabled,
 both uninterruptible (D state) and interruptible (S state) sleep stacks are collected by
--- a/src/build_id_store.rs
+++ b/src/build_id_store.rs
@@ -102,6 +102,13 @@ impl BuildIdStore {
         self
     }
 
+    /// How many build-ids the store keeps before evicting the least recently
+    /// used; a fill past this count evicts rather than grows, so a caller
+    /// filling the store has no reason to read more files than this.
+    pub fn capacity(&self) -> usize {
+        self.cap
+    }
+
     /// Resolve a build-id to the binary to symbolize against, probing
     /// trusted sources on first sight. Returns `None` when no source knows
     /// the id (the caller renders the deferred `[buildid:...]` form).
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,13 +73,15 @@ struct Command {
     collect_build_id: bool,
     /// With --collect-build-id: the most distinct files the end-of-trace
     /// build-id index may read the ELF note of (the walk over every live
-    /// process's executable mappings); 0 = unbounded. A binary the walk
-    /// never reaches still resolves through the .build-id directories,
-    /// debuginfod, or the deferred [buildid:...] rendering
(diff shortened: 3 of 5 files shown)


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head; verdict as it produced it.

**Structure.** A follow-up to #234's end-of-trace build-id index walk. Three changes in `BuildIdIndexWalk` (`src/stack_recorder.rs`): a second dedupe tier keyed by one stat of the mapping (mapped path, size, mtime) on a dev/inode miss, so one container image's file collapses across every container; a walk order that visits the sampled processes first (each `StackInterner` now records the tgids it interned stacks for) and then every other pid newest first, so a budget trip drops the oldest processes; and the file budget clamped to `BuildIdStore::capacity()` (a new accessor) with the clamp named on the summary line. The summary also reports the second tier's count. README, flag help and the `Config` doc follow.

**Rust quality.**
- The two dedupe tiers are the same shape (`is_some_and(|k| !seen.insert(k))`), read top to bottom in cost order: the free maps-line key, then one `metadata` call, then the ELF read. A `stat` failure falls through to the read rather than short-circuiting, which keeps "distinct files read" as the budget's unit.
- `StatKey` is the smallest struct that carries the identity; `mtime_ns` as `i128` is deliberate (seconds × 10⁹ plus nanoseconds without overflow reasoning) and is only ever compared, never arithmetic'd on.
- `build_id_index_walk_order` is a pure function over the pid list — `partition` then the reversed remainder — which is what makes the order testable without `/proc`.
- The interner's `sampled_tgids` insert sits on the miss path only (after the `stack_ids` lookup returns nothing), so the per-event cost is unchanged; the doc comment says so, which is the why a reader needs.
- The clamp lives in `BuildIdIndexWalk::new`, so `over_budget` loses its `max_files > 0` special case: a budget is now always a positive number.
- Doc comments explain why (overlay's per-mount device, what a trip costs, what a false merge costs) and reference functions and fields, never line numbers.

**Unsafe usage.** None added.

**BPF quality.** No BPF source touched; the objects are unchanged (no `build.rs` / `src/bpf/` change).

**Schema & versioning.** No schema change, `SCHEMA_VERSION` untouched, no new flag; `Cargo.toml`/`Cargo.lock` untouched. Rides the next tag.

**Integration.** `finish_inner` unions the interners' tgid sets before the walk and hands them to `run`; the interners are only borrowed there and drained afterwards as before. The summary line changes shape (`… deduplicated (M across mounts) …`, "readable build-id note") — anything parsing the old line by regex needs the new wording; the tests pin the new one. The live-fill integration test pins an unbounded budget, so it no longer depends on the default budget against the host's process count.

**Critical issues:** none.

1. The second tier's `metadata` call is one syscall per non-deduplicated mapping; on a host where the first tier already collapses everything (one mount) it is pure overhead, though microseconds against a note read. If the summary's `across mounts` count reads zero everywhere it runs, the tier could be gated on the device's major being 0 (an overlay or other virtual superblock).
2. `deduped_across_mounts` is the number to watch on a many-container host; a follow-up could also print the count of distinct files the tier keyed, to size the `seen_by_stat` set.

PR #237, opened 2026-08-28
memory/network recorders: the classic tracepoint and kprobe hooks are the default again; the fentry/fexit form is opt-in behind --kernel-hooks

josefbacik:
Base `ff69c98a` (v1.17.2); the fix-forward of #232.

## What this changes

systing#232 (v1.17.0) made the fentry/fexit trampoline form of the memory recorder's mmap/munmap/brk hooks and of the network recorder's TIME_WAIT hooks the default, on the claim that trampolines detach in milliseconds. The claim was inverted: measured, the trampoline form makes the capture's `detach bpf programs` stop phase 3.4–5.8× LONGER than the classic tracepoints/kprobes. This PR makes the classic form the default again and keeps the trampoline form as an opt-in behind ONE new flag, `--kernel-hooks {classic,trampoline}` (default `classic`), governing both legs. No schema change: `sysinfo.memory_syscall_leg` / `network_tw_leg` keep their meaning and gain the plain values `tracepoint` / `kprobe` for the default form.

## Why (the kernel, read; the cost, measured)

Dropping the last BPF program on a kernel-function trampoline runs `bpf_trampoline_update` → `unregister_fentry` → `unregister_ftrace_direct` → `unregister_ftrace_function` → `ftrace_shutdown()`, whose DYNAMIC/DIRECT tail waits `synchronize_rcu_tasks_rude()` and `synchronize_rcu_tasks()` — torvalds/linux v6.12 `kernel/trace/ftrace.c` lines 3181 and 3190 (v6.6 :3184/:3194, v6.18 :3183–3204 per the cold verifier's read) — two Tasks-RCU grace periods, each sleeping at least 100 ms when any task is runnable at its scan, once per hooked FUNCTION, serialized on the closing thread. The three TIME_WAIT kprobes share one `kprobe_ftrace_ops` and pay that shutdown once; three trampolines pay it three times. A perf-attached tracepoint's detach is `tracepoint_synchronize_unregister()` = one SRCU + one RCU grace period per program. Only a raw tracepoint (`tp_btf`, the x86 page-fault move in #232, kept here) detaches through `call_rcu`.

`stop-phase: detach bpf programs done in Nms`, 2-vCPU guests with two busy loops and a live mmap/TCP workload, 3-s whole-system captures, the RELEASE binaries v1.16.3 (classic) vs v1.17.1 (trampoline):

| kernel | recorder set | v1.16.3 classic | v1.17.1 trampoline |
|---|---|---|---|
| 6.12.0 (rig) | memory only | 105 / 114 / 118 / 128 (med 116) | 645 / 671 / 674 / 679 (med 672) |
| 6.12.0 (rig) | network only | 120 / 138 / 142 / 160 (med 140) | 610 / 613 / 630 / 631 (med 622) |
| 6.12.0 (rig) | default + memory + network | 908 / 954 / 960 / 986 (med 957) | 2024 / 2140 / 2152 / 2203 (med 2146) |
| 6.12.0 (rig) | default set alone | 69 / 75 / 85 | 55 / 71 / 105 |
| 6.12.85 (mainline) | memory only | 214 / 161 | 638 / 657 |
| 6.12.85 (mainline) | network only | 183 / 175 | 597 / 610 |
| 6.12.85 (mainline) | default + memory + network | 1444 / 1391 | 2348 / 2407 |
| 6.12.85 (mainline) | default set alone | 21 | 24 |

Caveat: both guests are 2-vCPU TCG; the classic teardown (SRCU/RCU, the tasklist walk, the static-key IPIs) and the Tasks-RCU pair both scale with CPUs and tasks, so the ratio need not transfer to a 180-CPU host — the decisive read is the by-phase p50/p99 on many-core hosts under each form once this release runs there, which the opt-in flag makes an A/B. The classic default is the known cost at scale (the classic form on v1.15.5 today: 459 ms p50 / 618 ms p99 on 180-CPU hosts with both recorders on, 2.4 s p50 on a large arm64 host class, and a `detach bpf programs` p99 reaching 15.6 s on at least one host class in a per-phase table read on 2026-08-28); the trampoline form's real saving is during the capture (nothing on other syscalls' entry path), not at stop.

## What the code does

- `Config.kernel_hooks: KernelHooks { Classic (default), Trampoline }`; CLI `--kernel-hooks classic|trampoline`. Under `classic` the trampoline programs are NOT loaded (the load set is v1.16.3's plus `tp_btf/page_fault_user` on x86) and no vmlinux BTF parse is made; under `trampoline` everything v1.17.0 did stands: the kallsyms + BTF load gates, the attach-time fallback to the classic set, the test-only force switches.
- `memory_syscall_leg`: `tracepoint` (default form), `fentry` (opt-in), `tracepoint:nosym|nobtf|notramp` (the classic set under the trampoline form, with why), `off:<cause>`. `network_tw_leg`: `kprobe` (default), `fentry`, `kprobe:nobtf|notramp`, `off:<cause>`.
- Load-shape table: `memory-trampoline`, `network-trampoline` rows and `all-recorders` opted in, so the trampoline programs still meet the verifier on every gate run; the forced classic-only rows now model a trampoline-form capture on a kernel without the symbols/BTF.
- Tests: the unit tests cover both forms' selection and values; the memory e2e asserts the exact default value and gains `test_memory_syscall_hooks_trampoline_opt_in`; the network suite asserts `kprobe` and gains `test_e2e_network_tw_leg_trampoline_opt_in`; the two forced-fallback e2es run in trampoline mode (the `fentry || tracepoint:nobtf` acceptance is now per-mode exact).
- Docs: SCHEMA_CHANGES schema 20 rewritten (mechanism as read + as measured, both kernels); #232's `unsigned long freeze` / `(u8)` cast on `__split_huge_pmd_locked` documented under schema 20 (it changes `memory_thp.result` on splits whose caller left stale upper register bits); "riscv before 6.7" → "before 6.6" (two places); the BPF source comments and the MCP `trace_info` description say the same.

## Not in this PR

- The launcher-side argument allowlist entry for `--kernel-hooks` (a separate small change in the launcher's own repository) — until it lands, launched captures run the default classic form.
- The alternative of moving the six link drops off the stop path (a detached drop after the parquet flush) — filed as a follow-up; it only helps a consumer that reads the output before the process exits.
- `vmlinux_btf_has_funcs` still uses libbpf-rs's `type_by_name::<Func>` (kind-checked by the `TryFrom`); an exact `btf__find_by_name_kind(FUNC)` lookup is a follow-up — a same-named non-FUNC type would read as `nobtf` and run the classic set, a safe failure.

## Gates (on the committed bytes)

- `cargo fmt --all -- --check` exit 0; `cargo clippy --all-targets -- -D warnings` exit 0; `cargo test --lib` 580 passed / 0 failed (the changed unit tests by name: `test_memory_syscall_hooks_selection_and_leg_values`, `test_network_tw_leg_follows_the_kernel_probe`, `shape_table_names_are_unique_and_cover_the_recorders`).
- VM rig, vanilla 6.12.0 guest, the object built with the release compiler (clang 21, `-mcpu=v3 -fwrapv`): `bpf_load_shapes::every_shape_loads` PASS (34 shapes; the two new trampoline rows and `all-recorders` load the fentry/fexit and TIME_WAIT fentry programs beside the classic sets; every default row loads the classic set alone); `memory_recorder` 11 passed / 0 failed (`test_memory_recorder_e2e` asserts `memory_syscall_leg = tracepoint`; `test_memory_syscall_hooks_trampoline_opt_in` → `fentry`; `test_memory_syscall_hooks_fall_back_to_tracepoints` → `tracepoint:notramp`); `trace_validation` network suite 3 passed / 0 failed with the guest's own lines `network recorder: TIME_WAIT leg kprobe` (default), `… fentry` (opt-in), `… kprobe:notramp` (forced fallback).
- The BPF object: a standalone release-flag compile of this tree's `systing_system.bpf.c` disassembles per program section STRICTLY IDENTICAL (85/85 sections) to the same compile of the v1.17.0 tree — the BPF source changes here are comments only, no instruction changes.

Rollback: revert this commit (the previous release, v1.17.2, then carries the trampoline default again; the release in use today is v1.16.3, the classic form).


josefbacik:
rust-bpf-code-reviewer verdict on `e5ddaf8f` (the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition, executed in-session over the whole diff `ff69c98a..e5ddaf8f`, 10 files, +645/−327; mechanics: the five passes below, the diff read hunk by hunk, the two probe/attach paths traced by hand for both hook forms, the BPF object compiled standalone with the release flags and compared per program section against the v1.17.0 tree's object)

### Summary
A one-switch change: `Config.kernel_hooks` (`Classic` default, `Trampoline` opt-in) threads through the two kernel-leg probes and the load set, and the sysinfo leg values gain the plain `tracepoint` / `kprobe` for the default form. The trampoline machinery from v1.17.0 is untouched behind the switch; the BPF object's instructions are byte-identical to v1.17.0's (comments only). The tests cover both forms at the unit, load-shape and e2e levels. No Critical issues.

### Critical Issues
None. No schema change (`SCHEMA_VERSION` stays 20; no table definition moved; no `[bump minor]` needed — a patch release). `Cargo.toml` / `Cargo.lock` untouched. No new `unsafe`. No ELF parsing. No recorder table change (utid rules unaffected).

### Improvements
1. `--kernel-hooks` is parsed as a `String` through `PossibleValuesParser` and then `.parse().expect(…)` into `KernelHooks` (`src/main.rs`). Deriving `clap::ValueEnum` on `KernelHooks` (clap is already a dependency of the crate) and declaring the field as `KernelHooks` with `value_enum` would drop the `expect` and the duplicated value list. Not done here to keep this fix-forward's diff to the mechanism; the `expect` cannot fire (clap admits only the two strings the `FromStr` accepts). Follow-up material.
2. `vmlinux_btf_has_funcs` still resolves a name with `type_by_name::<Func>` (libbpf-rs's `btf__find_by_name` + a kind-checked `TryFrom`), so a same-named non-FUNC BTF type would shadow the function and read as `nobtf` (the classic set then runs — a safe failure under either form). `btf__find_by_name_kind(FUNC)` is exact; follow-up, noted in the PR body.

### Suggestions
- `KernelHooks`'s doc comment carries the measured ratios; if a later measurement moves them the enum comment and SCHEMA_CHANGES both change — acceptable while the numbers are the reason the default is what it is.
- The `LegSelection::SyscallTracepointOnly` / `NetworkTwKprobeOnly` rows now only mean something under the trampoline form; their table rows set `kernel_hooks = Trampoline` explicitly, and `bpf_load_probe` would force the same selection under the classic form (a no-op there). A debug assertion that the forced rows carry the trampoline form would make the intent mechanical.

### Positive Notes
- The default form loads only the classic set and skips the vmlinux BTF parse entirely (`probe_kernel_symbols` takes the trampoline targets only under the opt-in), so a default capture pays nothing for the switch.
- The sysinfo values are backward-readable: every value v1.17.0–1.17.2 could write still means what it meant; the default form adds two plain values and the docs (`SCHEMA_CHANGES.md`, `analyze/mod.rs`, the MCP `trace_info` text) list them in one voice.
- The e2e tests assert exact per-form values (`tracepoint` / `kprobe` for the default, `fentry` for the opt-in, `tracepoint:notramp` / `kprobe:notramp` for the forced fallback) instead of the earlier either/or acceptance, and the load-shape table keeps the trampoline programs in front of the verifier through the two `-trampoline` rows and `all-recorders`.
- The SCHEMA_CHANGES entry says how each claim was established (the kernel lines read; the numbers measured, on which guests and kernels) rather than asserting a mechanism.

Verdict: APPROVE for merge as a patch release. Gates on these bytes: fmt 0 / clippy `-D warnings` 0 / lib 580 passed; rig `every_shape_loads` PASS (34 shapes), `memory_recorder` 11/11, network suite 3/3 with the guest's own leg lines; BPF object 85/85 program sections strictly identical to the v1.17.0 tree's.

PR #238, opened 2026-08-31
memory recorder: probe the global split_huge_pmd_locked funnel when the worker is inlined

josefbacik:
## What

The memory recorder's THP leg probes PMD splits at the static worker `__split_huge_pmd_locked` when kallsyms lists it (every split is counted there — `sysinfo.memory_thp_leg = on`). A build that inlined the worker fell straight back to the public entry `__split_huge_pmd` (`on:pmd-entry`), which is reached for any populated PMD on a VMA adjust (over-counting VMA churn) and, on 6.10+ kernels, never sees the rmap paths: `try_to_unmap_one` enters through the global `split_huge_pmd_locked` (rmap.c v6.12:1703).

This adds the middle tier: a kprobe program on the global funnel `split_huge_pmd_locked` (a `T` symbol on 6.10+; huge_memory.c v6.12:2867, v6.18:3051), selected when the worker is inlined and the funnel is listed, recorded as `memory_thp_leg = on:pmd-global` / `on:pmd-only:global`. The entry stays the last fallback for a kernel without the funnel.

## Why it is the right tier

Read at v6.12 and v6.18 (`mm/huge_memory.c`): `__split_huge_pmd` takes the PMD lock and calls `split_huge_pmd_locked(vma, range.start, pmd, freeze[, folio])`; the rmap paths call the same function directly; the function checks `pmd_trans_huge || pmd_devmap || is_pmd_migration_entry` (the folio check beside it on 6.12) and then calls the static worker. So:

- a `pmd-global` row set covers every split the worker counts, including reclaim's unmap side that the entry probe misses, and `result` carries the split's `freeze` flag as at the worker;
- the funnel is reached before its own huge check, so like the entry it also fires for a PMD that is not huge — `pmd-global` rows still over-count on VMA churn, and the vmstat `thp_split_pmd` delta stays the host-wide truth (the e2e's rows ≤ vmstat bound is asserted only at the worker, as before);
- it runs inside the PMD's page-table lock like the worker (the probe's ringbuf reserve and stack capture sit inside that spinlock hold, bounded by the 1-in-N sample rate); the entry probe fires before the lock.

Its parameters are `(vma, address, pmd, freeze, folio)` on 6.12 and `(vma, address, pmd, freeze)` on 6.18 — the first four are the same on both, with `address` and `pmd` swapped relative to the worker's `(vma, pmd, haddr, freeze)` — so it is its own program (`systing_thp_split_pmd_global`), with `freeze` read as a raw register and masked to its low byte like the worker's.

## Changes

- `src/bpf/systing_system.bpf.c`: `SEC("kprobe/split_huge_pmd_locked") systing_thp_split_pmd_global`; the THP header comment describes the three tiers.
- `src/systing_core.rs`: `THP_PMD_GLOBAL_PROG`; `split_huge_pmd_locked` in the probed symbol list; the selection order worker → funnel → entry; `thp_leg_value` names the tier (`on:pmd-global`, `on:pmd-only:global`); the manual-attach predicate; unit tests for the tier order (a synthetic kallsyms set per case, the worker winning when both are listed) and the leg strings.
- `src/bpf_load_shapes.rs`: a forced row `memory-vfio-thp-forced-global` so the new program meets the verifier at every load-shape gate.
- `tests/memory_recorder.rs`: the e2e derives the expected tier from the guest's own `/proc/kallsyms` (worker → funnel → entry, and the folio-split half) and asserts `sysinfo.memory_thp_leg` against it, so a kernel that inlines the worker exercises the funnel for real — the vanilla 6.12 test kernel does; the bound comment names the new tier.
- `SCHEMA_CHANGES.md` (schema 20, "from the next release, unreleased"), `src/analyze/mod.rs`, `src/mcp.rs`: the tier and its read rule. The v19 VFIO teardown note adds that an interrupt's `iommu:unmap` on the tearing-down task's CPU is counted under that task's window for the whole unmap-and-unpin walk (`vfio_iommu_unmap_unpin_all`, seconds to minutes for hundreds of GB pinned) — the IRQ-attribution class the ioctl windows already admit, over a longer window; not a defect.

No schema change; the new `memory_thp_leg` values are strings in an existing nullable column.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` (the tier-order test and the leg-string test by name): fmt clean, clippy 0 warnings, 585 lib tests passed (`test_thp_pmd_probe_tier_follows_kallsyms_in_order` and `test_memory_optional_legs_follow_the_kernel_probe` among them).
- VM rig, vanilla 6.12.0 guest: `bpf_load_shapes::every_shape_loads` — every shape incl. the new forced row loads (PASS, 251 s); the `memory_recorder` e2e suite — 11 of 11 passed (836 s), the THP test with its new tier assertion (the guest inlines the worker and exports the funnel, so the kallsyms-derived expectation the passing assertion matched is `on:pmd-global`; a passing test prints no leg line) and its pmd rows > 0 with stacks.

## What is not verified here

- The funnel on a real 6.10+ host that inlined the worker is exercised only on the test kernel; on a host whose kernel exports the worker the selection stays `on` (the tier-order unit test pins the precedence). The funnel's rows-vs-vmstat behaviour is documented, not asserted (the over-count on non-huge PMDs is by construction).
- A build that also inlines the funnel into its same-file caller `__split_huge_pmd` (the symbol stays exported for rmap.c) would leave the funnel probe with the rmap paths only; the recorder cannot tell, so the e2e's `madvise` split check (rows at the funnel on a worker-inlining kernel) is the detector — it passed on the 6.12 guest, and would fail rather than pass quietly on such a build. Documented in `SCHEMA_CHANGES.md` and the BPF comment.
- 6.18's four-parameter form is read at the source, not run here (no 6.18 guest in the rig).


josefbacik:
**Code review — `rust-bpf-code-reviewer` pass** (the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition, executed in-session by the author over the full diff before the push; mechanics disclosed: one read of the definition as a tracked blob, the whole `git diff` read hunk by hunk, the load-shape gate and the memory_recorder e2e suite run on a vanilla 6.12.0 VM guest).

### Summary
A small, well-bounded change: one new kprobe program on the 6.10+ global funnel `split_huge_pmd_locked`, selected between the worker and the entry from the same kallsyms read, recorded as its own `memory_thp_leg` tier, with the load-shape row, unit tests for the tier order and the leg strings, and an e2e assertion that derives the expected tier from the guest's own kallsyms. No schema change; docs describe the tier and its read rule honestly (covers every split, over-counts VMA churn).

### Critical Issues
None.
- Schema & versioning: no table or column changes (`SCHEMA_VERSION` unchanged is correct; the new values are strings in an existing nullable column); `SCHEMA_CHANGES.md` carries the entry under schema 20 as "from the next release, unreleased"; `Cargo.toml` `version` untouched; no `[bump minor]` needed.
- Map/context pairing: the new program writes only the memory ring buffer (through the shared `memory_thp_emit`) and the THP sample counter (a plain array, written by the THP kprobes/kretprobes only — no perf-event/NMI writer shares it); no `LRU_HASH` anywhere under `src/bpf/`.
- Argument reads: `freeze` is taken as the raw fourth register and masked to its low byte (the `_Bool` upper-bits rule already applied to the worker probe); `address` is the funnel's second parameter on both 6.12 (five parameters) and 6.18 (four), read at the source.

### Improvements (applied before the push)
- Documented the one residual the recorder cannot see: a build that also inlines the funnel into its same-file caller `__split_huge_pmd` (the symbol stays exported for rmap.c) would leave the funnel probe with the rmap paths only. Added to the BPF comment and to the `SCHEMA_CHANGES.md` entry, with the detector named: the e2e's `madvise` split check asserts rows at the funnel on a kernel that inlined the worker (it passed on the vanilla 6.12 guest, so the funnel is a real call there).

### Suggestions (not applied)
- `tests/memory_recorder.rs`: `has_text_symbol` rescans `/proc/kallsyms` once per name (five scans). A single pass collecting the five names into a set would read cleaner; left as is — a test-only cost of five linear scans, and the closure keeps the tier derivation next to the assertion that uses it.

### Positive Notes
- The tier selection stays a pure function of the kallsyms set, and `test_thp_pmd_probe_tier_follows_kallsyms_in_order` pins the precedence with synthetic symbol sets — including the case every 6.10+ build that kept the worker hits (worker AND funnel listed → the worker).
- `thp_leg_value` names the tier once instead of guarding each arm with a worker check, so a fourth tier would be one more match arm.
- The forced load-shape row `memory-vfio-thp-forced-global` gives the new program a verifier pass on every gate run, not only on hosts whose kernel selects it.

PR #239, opened 2026-09-02
Name every cause of an empty --cgroup slot, test the tag read on a built fixture, take the build-id index nits, and refuse PIVOT without an IN list

josefbacik:
Three follow-ups from the cold reads of #211/#214, #215 and #236, batched into one change, plus a `PIVOT` guard on the analyze query tool found by a cold read of #218. Two separable commits. No behaviour change for a trace that starts, and nothing here changes the arguments any caller passes.

## Commit 1 — `--cgroup` cause naming, the discovery tag read, the build-id index nits

**`--cgroup` (the #215 follow-up).** `install_cgroup_targets` named one cause when the iterator program left a target's `cgroup_target_refs` slot empty — a private cgroup namespace — which cannot hold on kernels carrying upstream 2c8951339506 ("bpf: Do not limit bpf_cgroup_from_id to current's namespace": v6.18, backported to 6.6.117+ and 6.12.58+). The empty slot is now explained from a fresh `cgroup_id()` of the same path: the cgroup was removed while systing started (ENOENT), it was replaced (a different id at the path), or the kernel refused the lookup — a cgroup v1 hierarchy, or the namespace on a kernel before 6.6.117 / 6.12.58 / 6.18 — with both cures still named. The wording is a pure function (`empty_cgroup_target_slot_cause`) with a unit test; README gets the same version hedge.

`test_e2e_cgroup_filter_outside_cgroupns_names_the_cause` no longer pins the pre-backport outcome: it puts a busy workload in the target and branches on what the kernel did — a refusal must name the cause and the fallback, a success must have run in kernel mode and traced that workload — and the forced-fallback run proves the trace by the workload's rows, not by exit status.

**pystacks discovery (the #211/#214 follow-ups).** The cache-tag read had no discriminating test on a runner whose libpython is stripped of `_PySys_ImplCacheTag`, so `test_read_impl_cache_tag_on_a_built_fixture` builds two stand-in runtimes with the host's C compiler — a shared object whose tag symbol is the string, and a non-PIE executable whose tag symbol is a pointer resolved at link time — under names the filename fallback cannot satisfy, and asserts the tag and the version (3, 99) read from them; under `CI` a missing compiler fails the test instead of skipping it (`CC` may carry a driver prefix such as `ccache cc`). `real_runtimes()` falls back to the executable when `Py_ENABLE_SHARED` says shared but the library is not installed (Debian's static `/usr/bin/python3`), and the ET_DYN assertion is keyed on what it means (a shared build's library must be one; a static build's PIE is reported). The section-range arithmetic in the tag read is checked rather than wrapping.

**build-id index (the #236 nits).** `BuildIdIndexStats` carries the file budget and its clamp from construction rather than only after `run()` (the struct field that duplicated it is gone; the clamp test reads the stats); the "newest first" order is documented as the pid-order heuristic it is (README and the walk's doc: a young process holding a recycled low pid after wraparound walks last); the second-tier stat before the budget check is explained; the live-fill test's comment says what a file budget of 0 means (the store's capacity).

## Commit 2 — `analyze`: refuse `PIVOT` without an `IN` list; a prepare-time memory failure is reported as the bound

duckdb-rs `prepare()` executes every statement but the last of the text it is handed. A `PIVOT` with no `IN (...)` list is expanded by DuckDB's parser into a `CREATE TYPE ... AS ENUM (SELECT DISTINCT ...)` ahead of the `SELECT`, so the enum scan ran at prepare time outside the row cap, ran twice on the failure path (the raw-prepare fallback), surfaced the engine's allocation text instead of the bound (and told the caller to raise a setting the lock refuses), and left one `__pivot_enum_<uuid>` type in the temp catalog per successful pivot. The classifier now refuses the shape (`StatementShape::PivotWithoutIn`; the same pivot with its values named is one statement and runs wrapped), both prepare error paths go through `classify_query_error`, and a prepare that hit the memory bound is never re-prepared. `PIVOT_WIDER`, the spelled-out form, now runs wrapped like `PIVOT` — it was missing from the row-reader list. The MCP `query` tool description names the rule beside the dollar-quote one.

## What was run

- `cargo fmt --all -- --check` clean; `cargo clippy --all-targets -- -D warnings` exit 0; the unit suite 589 passed / 0 failed, on the exact bytes of each commit.
- The integration tests in a VM (the repository's `bzImage-v6.12-bpf` guest, 6.12.0) on commit 1: `trace_validation` with the filter `test_e2e_cgroup_filter` — 3 passed / 0 failed. On that pre-backport guest the cgroupns test took the refusal arm ("refused with the cause named") and the forced fallback traced the workload; **the success arm (a kernel with the `bpf_cgroup_from_id` namespace fix, 6.6.117 / 6.12.58 / 6.18) is unexercised here** — a 6.12.85 or 6.18 host would take it.
- Commit 2's tests run on the host (DuckDB 1.5.4 via libduckdb-sys 1.10504.0): `test_statement_shape_refuses_pivot_without_an_in_list`, `test_query_reports_a_prepare_time_memory_failure_as_the_bound`, and the extended `test_query_comments_and_terminators_cannot_bypass_the_cap` (three refused pivots, `duckdb_types()` reads 0 `__pivot_enum%` after them, the `IN`-form runs wrapped and is capped at 10,000 of 25,000 rows).


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of `4e82954` (two commits, `def11e1` + `4e82954`, against `7902d87`, v1.17.5); verdict as it produced it.

### Summary
Userspace Rust and tests only: no BPF program, no schema, no `Cargo.toml` version edit, no new `unsafe`. The first commit names every cause of an empty `--cgroup` reference slot from a fresh stat of the path, makes the cgroup-namespace e2e test branch on the kernel's actual behaviour, adds a discriminating test for the CPython cache-tag read on compiler-built fixtures, and takes small documentation and stats nits in the build-id index walk. The second commit closes a real gap in the analyze query tool: a `PIVOT` without an `IN` list is parser-expanded into two statements, and the binding's `prepare` executed the first one outside the row cap. Both read as focused, well-tested changes; nothing here needs to change before merge.

### Critical Issues
None.

### Improvements
- `src/analyze/query.rs` `has_in_list`: the quote-skipping scan duplicates the scanning idiom already used by the comment stripper in this file. If a shared "next unquoted token" helper is ever factored out, this should ride it; as written it is correct and bounded (one pass over the text), so this is not a blocker.
- `src/analyze/mod.rs`: the refusal message for `PivotWithoutIn` is long for a single `bail!`. It reads well in the MCP client because it names the cure, so leaving it is defensible; a shorter first sentence with the mechanism second would scan faster in a log.

### Suggestions
- `src/pystacks/discovery.rs` fixture test: the non-PIE leg is skipped locally when the compiler lacks `-no-pie` and fails only under `CI`; a one-line note in the test's doc comment already says so. Consider asserting `checked == 2` under `CI` explicitly (today `checked >= 1` plus the CI-only `assert!` on the build already imply it).
- `src/systing_core.rs` `empty_cgroup_target_slot_cause`: the `_` arm folds "same id" and "stat failed for another reason" together on purpose (the comment in the test says why). A short doc sentence on the function saying that an EACCES/EIO stat is treated as "unknown, assume the kernel refused" would save a reader the trip to the test.
- `tests/trace_validation.rs`: the `--duration 2` run means the workload has one to two seconds of samples in the guest; the assertion is on `process` rows (any traced task has one), which is the right bar for "was it traced" and does not depend on sample counts.

### Positive Notes
- The cause wording is a pure function with a unit test that pins each arm and the negative (no namespace text on the removed/replaced arms) — the message a user sees is under test, not only the code path.
- The cgroupns e2e test now proves whichever outcome the kernel gives instead of pinning pre-backport behaviour; the workload in the target turns "exited 0" into "traced".
- The fixture test finally exercises the tag read on a runner whose distro libpython is stripped of the tag symbol, with filenames the fallback cannot satisfy, so the version can only come from the read under test.
- The `PIVOT` guard is placed in the classifier (before anything reaches `prepare`), the memory-bound classification is reused for both prepare paths, and the extended cap test checks the temp catalog for leftover enum types — the observable side effect, not just the error string.
- Removing the duplicated `file_budget_clamped` field and filling the stats at construction is the simpler shape the tests wanted.

PR #240, opened 2026-09-03
analyze: read the PIVOT value list from the pivot's own ON clause; --cgroup: name a target off cgroup2 as its own cause

josefbacik:
Two follow-ups to #239, batched (the analyze `query` guard and the `--cgroup` empty-slot cause), each its own commit.

**1. analyze: refuse a PIVOT unless its own ON clause lists every pivot column, and test the no-re-prepare arm.** The `query` tool's PIVOT check (`has_in_list`) scanned the whole statement text for `IN (`, so a PIVOT whose only `IN (` sat in its SOURCE subquery — `PIVOT (SELECT * FROM t WHERE id IN (3, 4, 5)) ON bucket USING count(id)` — was classified as a listed pivot and wrapped: DuckDB still expands it into `CREATE TYPE __pivot_enum_… AS ENUM (SELECT DISTINCT …)` + the SELECT, `prepare()` executes the enum scan outside the row cap, and the type stays in the read-only database's temp catalog (a `__pivot_enum` count of 0 → 1 through the shipped `AnalyzeDb::query`). The check is now `pivot_names_every_value_list`: the statement is tokenised at its own paren depth, the pivot's ON clause is the first top-level `ON` after the first word up to the first top-level `USING` / `GROUP` / `ORDER` / `LIMIT` or the end, and every comma-separated element of it must carry a top-level `IN (`; a source subquery's `IN (`, a WHERE, a comment, a literal, a second unlisted pivot column or no ON clause at all is refused as `PivotWithoutIn`, and so is a join source written without parentheses (its own `ON` ahead of the pivot's — the join predicate must not vouch for the pivot column; the same source in parentheses reads fine). The prepare-time memory-bound path is factored as `query_after_wrapped_prepare_failed(err, inner, must_wrap)` and the test drives it directly with a memory-bound-shaped `duckdb::Error`: the bound's fixed text and no rows (deleting the arm returns rows, so the test fails), while a Binder-shaped error runs the raw form or bails "could not be bounded" under `must_wrap`. The `… USING count(id) WHERE id IN (1, 2)` case moved to the refused list with its comment corrected (it was never runnable DuckDB syntax).

**2. --cgroup: name a target off the cgroup2 hierarchy as its own cause; cite DuckDB's one OOM class.** The empty-slot cause naming from #239 asks the filesystem what the target path says now, and `cgroup_id()` is the inode of whatever directory the path names: a mistyped `--cgroup /tmp/pod-x`, or a cgroup on a v1 mount, resolved to an id like any cgroup and was then reported as "a v1 hierarchy or the cgroup namespace". A `statfs` of the path (`cgroup::cgroup_fs`) now tells the causes apart: removed and replaced as before, a cgroup v1 hierarchy, a directory on no cgroup filesystem at all (named with the magic it did find), and — only when the path is on the cgroup2 hierarchy — the pre-backport namespace cause, worded without the v1 hedge; when the filesystem cannot be read the last two are named together, as before. The README sentence follows. The analyze classifier's doc cites the DuckDB 1.5.4 sources for both bounds raising `OutOfMemoryException` (the block-offload throw in `temporary_file_manager.cpp` for the spill bound, the buffer pool for `memory_limit`), which is why the one "Out of Memory Error" prefix covers `memory_limit` and `max_temp_directory_size` alike.

**Tests** (`cargo test --no-default-features` on the exact tree; fmt and clippy `--all-targets -D warnings` clean): `analyze::query::tests::test_statement_shape_refuses_pivot_without_an_in_list` (the probe shape, the two-column and the join-source cases refused; ten accepted shapes, a parenthesised source with `ON bucket IN (0, 1)` and a parenthesised join source among them), `analyze::tests::test_query_comments_and_terminators_cannot_bypass_the_cap` (engine-driven: the probe shape refused with `__pivot_enum` count 0 after it; the listed form on the same source runs wrapped, 3 rows), `analyze::tests::test_query_reports_a_prepare_time_memory_failure_as_the_bound`, `test_empty_cgroup_target_slot_cause_names_what_the_path_says_now` (every arm: removed, replaced, cgroup2 + namespace, unreadable fs + both causes, v1, a plain directory), `test_cgroup_fs_reads_the_filesystem_magic` (a temp dir reads as no cgroup fs, a missing path errors, the cgroup2 root reads V2 where mounted). No BPF, no capture-path or recorder change; the BPF object is untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,9 @@ lifted that limit): there, from a container with a private cgroup namespace and
 the host's cgroup filesystem mounted, a host path is visible but cannot be
 resolved, and systing says so at start — run it in the host cgroup namespace
 or use the fallback. Only the unified cgroup v2 hierarchy is supported; a
-target on a cgroup v1 hierarchy is refused the same way.
+target on a cgroup v1 hierarchy, or a directory that is not on a cgroup
+filesystem at all, is refused at start with that cause named (systing checks
+the target's filesystem before blaming the namespace).
 
 `--trace-event` - This will add an `instant` track event for each event that
 this tool captures.  The format is "<trace type>:<optional
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -539,42 +539,53 @@ impl AnalyzeDb {
                 }
                 Ok(result)
             }
-            // The binding's `prepare` is not parse-only: it executes every
-            // statement but the last of the text it is handed, so a prepare
-            // can already have run a scan (a DuckDB-expanded statement the
-            // classifier let through) and hit the memory bound. That failure
-            // is reported as the bound — never as the engine's allocation
-            // text, which would tell the caller to raise a locked setting —
-            // and the text is not prepared a second time, which would run
-            // the same scan again for the same answer.
-            Err(err) if is_memory_bound_error(&err) => Err(self.classify_query_error(err)),
-            Err(_) => {
-                // Prepare the user's own text so a SQL error names their
-                // line and column, not the wrapper's.
-                let mut stmt = self
-                    .conn
-                    .prepare(&inner)
-                    .map_err(|e| self.classify_query_error(e))?;
-                if must_wrap {
-                    // It prepares on its own but not inside the wrapper: a
-                    // shape that can read rows and that the cap cannot
-                    // bound, so it does not run at all rather than run
-                    // unbounded.
-                    bail!(
-                        "query could not be bounded by the {MAX_QUERY_ROWS}-row cap; rewrite it as a single SELECT statement"
-                    );
-                }
-                // A non-row-reading statement: run it as written. Its result
-                // is small by nature and already materialized, so the rows
-                // beyond the cap are drained and counted.
-                let (mut result, beyond_cap) = self.run_capped(&mut stmt, true)?;
-                if beyond_cap > 0 {
-                    result.truncated = true;
-                    result.total_row_count = Some(result.row_count + beyond_cap);
-                }
-                Ok(result)
-            }
+            Err(err) => self.query_after_wrapped_prepare_failed(err, &inner, must_wrap),
         }
+    }
+
+    /// The rest of [`Self::query`] once the wrapped form failed to prepare
+    /// with `err`. Split out so the decision below can be driven directly.
+    ///
+    /// The binding's `prepare` is not parse-only: it executes every
+    /// statement but the last of the text it is handed, so a prepare can
+    /// already have run a scan (a DuckDB-expanded statement the classifier
+    /// let through) and hit the memory bound. That failure is reported as
+    /// the bound — never as the engine's allocation text, which would tell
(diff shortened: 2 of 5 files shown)


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head at the time of the run (two commits on v1.17.6); verdict as it produced it.

### Summary
Two follow-ups to #239, each doing one thing: the analyze `query` tool's PIVOT guard now reads the pivot's own ON clause instead of the whole statement text, with the prepare-time memory-bound arm factored into a function a test can drive; and the `--cgroup` empty-slot cause consults the target's filesystem (`statfs`) before blaming the cgroup namespace, so a mistyped path or a cgroup v1 mount is named as such. No BPF program, map, recorder or schema is touched. The code is small, the comments explain why rather than what, and every new arm has a test.

### Critical Issues
None.

### Improvements
- `pivot_names_every_value_list` (query.rs): the first top-level `ON` was taken as the pivot's, which read the wrong clause for a join source written without parentheses (`PIVOT t JOIN u ON u.k IN (1) ON bucket USING count(*)` — the join predicate's `IN (` vouched for an unlisted pivot column, the one shape left where the enum-building expansion could still reach `prepare`). Resolved in the same commit: a second top-level `ON` before the clause keyword now refuses the pivot, with the two join-source shapes in the refused list and the parenthesised join source in the accepted list.

### Suggestions
- `cgroup_fs` (cgroup.rs): the two `unsafe` blocks are minimal and the `// SAFETY:` note covers both (a NUL-terminated path that outlives the call, a zeroed out-parameter the kernel fills). Matching `st.f_type` directly against `libc::CGROUP2_SUPER_MAGIC` / `CGROUP_SUPER_MAGIC` keeps the comparison cast-free on the 64-bit Linux targets this builds for; if a 32-bit target is ever added, the `f_type` width differs and this is the one place to revisit.
- `empty_cgroup_target_slot_cause` (systing_core.rs): the `namespace_cause` closure with an `also_v1` flag is the smallest way to keep one wording for the two arms; a second flag would be the point to turn it into a small enum.
- `test_cgroup_fs_reads_the_filesystem_magic`: the cgroup2-root assertion runs only where a cgroup2 hierarchy is mounted, which keeps the test hermetic on hosts without one — worth keeping that guard when the test is extended.

### Positive Notes
- The tokeniser tracks paren depth from the statement's own first word, so `( PIVOT ... )` and a parenthesised source subquery read the same as the bare form, and comments and quoted runs never contribute tokens — the refused/accepted lists cover each of those explicitly.
- `query_after_wrapped_prepare_failed` makes the "never re-prepared after a memory-bound failure" claim a tested fact rather than a diff-read: deleting the arm returns rows and fails the test.
- The `--cgroup` cause wording now says what the filesystem found (the magic, or the v1 hierarchy) and tells the user where a cgroup2 target lives, instead of a two-way hedge.

PR #241, opened 2026-09-04
analyze: apply the spill cap on the opened connection; prove it by behaviour

josefbacik:
## What was wrong

`AnalyzeDb::open_with_spill_cap` applied the spill cap as `Config::with("max_temp_directory_size", cap)` before `Connection::open_with_flags`. In DuckDB 1.5.4 (the version `libduckdb-sys 1.10504.0` pins) that value never reaches the buffer manager:

- `MaxTempDirectorySizeSetting::SetGlobal` with no live database only stores `config.options.maximum_swap_space` and returns (`src/main/settings/custom_settings.cpp`, the `if (!db)` branch); only the live-database branch calls `BufferManager::SetSwapLimit`.
- `DatabaseInstance::Initialize` builds `StandardBufferManager(*this, config.options.temporary_directory)` and nothing in the open path calls `SetSwapLimit` (`src/main/database.cpp`); the buffer manager's `temporary_directory.maximum_swap_space` is default-unset and `SetSwapLimit` is its only writer (`src/storage/standard_buffer_manager.cpp`).
- At the first spill `RequireTemporaryDirectory` hands the unset value to the temp-directory handle, which falls back to `GetDefaultMax` = 90 % of the filesystem's free space (`src/storage/temporary_file_manager.cpp`) — the exact `statvfs` misreport the cap exists to avoid.
- `MaxTempDirectorySizeSetting::GetSetting` returns the config value first, so `current_setting('max_temp_directory_size')` echoed the requested cap while the default was enforced. `test_open_with_spill_cap_applies_bound` asserted that echo and passed on a dead cap; `lock_configuration` then locked the unenforced number.

Observed in practice: MCP servers started with `--max-temp-directory-size 8GiB` wrote well over 10 GB of DuckDB temp files for a single query, which an enforced 8 GiB cap makes impossible.

## What this changes

- `open_with_spill_cap` keeps the Config line (it validates the size string at open and keeps the readback consistent) and, after the open, runs `SET max_temp_directory_size = '<cap>'` on the connection — the live-database path, which `SetSwapLimit`s the buffer manager so the bound takes effect at the first spill. It runs before any caller's `lock_configuration` (`src/mcp.rs` `open_locked` = open, then lock), so the MCP server's cap is now both enforced and frozen; the CLI's `--max-temp-directory-size` gets the same fix.
- `test_open_with_spill_cap_applies_bound` is replaced by a behavioural test: a 12M-group hash aggregate under `memory_limit = '128MiB'` must go out of core (an aggregate rather than an `ORDER BY`, because `AnalyzeDb::query`'s `LIMIT` wrapper would turn a sort into a top-N heap that never spills); under a `1MiB` cap the raw DuckDB error names `max_temp_directory_size` and `query` renders the fixed memory-bound message; the same aggregate under a `10GiB` cap completes, which pins the refusal on the spill bound rather than the memory limit. The test is red at the base commit (the aggregate spills under the 90 % default and completes) and green with the change; the `current_setting` echo is kept only as a secondary assertion.
- Doc-comments on `open_with_spill_cap` name the DuckDB behaviour so the next reader does not trust the readback.

## DuckDB-side note

This looks like a DuckDB bug worth reporting upstream (not filed here): a `max_temp_directory_size` given on the `DBConfig` before the database opens is validated, stored on `config.options.maximum_swap_space`, echoed by `current_setting`, and never applied — `DatabaseInstance::Initialize` does not push `config.options.maximum_swap_space` into the buffer manager the way it pushes `maximum_threads` and `maximum_memory`. Observed at DuckDB v1.5.2 and v1.5.4 (`src/main/database.cpp`, `src/main/settings/custom_settings.cpp`, `src/storage/standard_buffer_manager.cpp`).

## Verification

`cargo test --no-default-features --lib -- test_open_with_spill_cap test_spill_cap test_lock_configuration` against the prebuilt libduckdb v1.5.4: the new test fails with the enforcing `SET` removed (`expect_err` fires: "a spill past a 1MiB cap completed") and passes with it (3/3); `cargo clippy --no-default-features --all-targets -- -D warnings` clean; `rustfmt --check` clean.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -36,7 +36,7 @@ pub use sched_stats::{
     ThreadDetailStats, ThreadSchedStats,
 };
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context as _, Result};
 use duckdb::Connection;
 use serde::Serialize;
 use std::collections::HashSet;
@@ -366,6 +366,21 @@ impl AnalyzeDb {
     /// passes it here (any DuckDB size string, e.g. "8GiB") so an oversized
     /// aggregation fails that one query instead of killing the process.
     ///
+    /// The cap is applied twice on purpose. On the `Config` before the open,
+    /// where DuckDB validates the size string (a malformed cap fails the open
+    /// loudly) and records the value `current_setting` echoes — but DuckDB
+    /// 1.5.4, the version `libduckdb-sys` pins, never hands a pre-open value
+    /// to its buffer manager: `DatabaseInstance::Initialize` builds the
+    /// buffer manager from the temp directory alone and only the setting's
+    /// post-open path calls `BufferManager::SetSwapLimit`
+    /// (`src/main/settings/custom_settings.cpp`, `src/main/database.cpp`), so
+    /// a Config-only cap leaves the 90 %-of-free-space default in force while
+    /// reading back as the requested bound. Then with a `SET` on the opened
+    /// connection, which does reach the buffer manager and takes effect at
+    /// the first spill. Because the readback cannot tell the two apart, the
+    /// bound is proven by behaviour (`test_open_with_spill_cap_applies_bound`
+    /// makes a query spill past it), never by `current_setting`.
+    ///
     /// On its own this is a structural default, not a jail: unlike
     /// `temp_directory`, DuckDB does not lock this setting when external
     /// access is disabled, so in-session SQL can still raise it (covered by a
@@ -405,12 +420,12 @@ impl AnalyzeDb {
             None => config.with("temp_directory", "")?,
         };
         if let Some(cap) = spill_cap {
-            // Set on the Config, like temp_directory above, so the bound is
-            // in force from the first query. DuckDB validates the size string
-            // at open, so a malformed cap fails loudly here rather than being
-            // silently ignored.
+            // Set on the Config, like temp_directory above: DuckDB validates
+            // the size string at open, so a malformed cap fails loudly here
+            // rather than being silently ignored, and `current_setting`
+            // reports this value. It does NOT bound anything on its own (see
+            // the doc-comment): the enforcing SET follows the open below.
             config = config.with("max_temp_directory_size", cap)?;
-            eprintln!("duckdb: max_temp_directory_size={cap}");
         }
         if let Some(m) = mem_limit_mib {
             config = config.max_memory(&format!("{m}MiB"))?;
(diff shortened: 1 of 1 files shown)


josefbacik:
rust-bpf-code-reviewer verdict — run over the full diff of `03e39e15` (1 file, +85/−26) against the repo's `.claude/agents/rust-bpf-code-reviewer.md` definition before push. Mechanics: the five passes (structure, Rust quality, BPF, schema & versioning, integration) were executed in-session by the author's assistant session as the reviewer, not by a separate reviewer process.

### Summary
A one-function fix with a test that proves the bound it claims. `open_with_spill_cap` keeps the pre-open `Config` value (string validation, consistent readback) and adds the post-open `SET` that DuckDB 1.5.4 actually applies; the readback-only test is replaced by a behavioural one that fails without the `SET` and passes with it. Nothing touches BPF, the schema, recorders, or `Cargo.toml`.

### Critical Issues
None.
- Schema & versioning: no table or `create_schema()` change, `SCHEMA_VERSION` untouched, `Cargo.toml` / `Cargo.lock` `version` untouched, no `[bump minor]` needed.
- Unsafe: none introduced.
- Comments: no source-line references (the doc-comment cites DuckDB by file and function, not by line).

### Improvements
None required.

### Suggestions
1. The doc-comment on `open_with_spill_cap` is long for a doc-comment; it carries the engine behaviour a future reader must not rediscover the hard way (a readback that lies), which is the reason to keep it — a follow-up could move the DuckDB mechanics to a module-level note and leave the doc-comment with the rule ("applied post-open; proven by behaviour, never by `current_setting`").
2. The behavioural test runs the capped aggregate twice (once raw, once through `query`) to assert both the engine's naming of the setting and the tool-facing message; about 0.3 s each. If test time ever matters, the raw run alone pins the bound and the `query`-path rendering is already covered by the memory-bound end-to-end test in this module.

### Positive Notes
- The test's control (the same aggregate under a 10GiB cap completes with the same `memory_limit`) is what makes the refusal attributable to the spill bound rather than the memory limit; and choosing a hash aggregate over an `ORDER BY` is deliberate — `query`'s `LIMIT` wrapper would turn a sort into a top-N heap that never spills.
- Quote-doubling on the cap literal costs nothing and closes the only injection shape a size string could take, even though the value is validated by the `Config` first.
- Evidence: the test fails with the enforcing `SET` removed (exit 101, `expect_err` fired) and passes restored (3/3); clippy `--all-targets -D warnings` clean; built against the prebuilt libduckdb v1.5.4.

PR #242, opened 2026-09-04
network-packets: --packet-sample-rate N, a per-packet keep with the rate in sysinfo [bump minor]

josefbacik:
On a large host the network-packets recorder loses most of its events, and
nothing in the trace says what share. The tier records one `packet_event`
(~130 bytes) per skb at every transmit and delivery hook; a 192-vCPU host
running `--only-recorder network --only-recorder network-packets
--ringbuf-size-mib 256` for 90 s printed `Missed packet events: 81,428,216`
(about 1 M events/s) with `Missed network events: 0` on the same run. At
that flag the packet family is already 64 rings × 32 MiB = 2 GiB
(`plan_ringbufs`), i.e. ~16 M events of headroom, ~16 s of a 1 M/s deficit:
the ceiling is the userspace consumer draining the rings, not the rings, so
a larger ring only delays the overrun. The cure is to record fewer packets,
and to say how many.

## The knob

`--packet-sample-rate N` (`tool_config.packet_sample_rate`; 0 or 1 = every
packet, the default, so the flag is inert by absence) keeps 1 in N
**packets** of the data-path event types: `PACKET_ENQUEUE`, `PACKET_SEND`,
`PACKET_RCV_ESTABLISHED`, `PACKET_QUEUE_RCV`, `PACKET_BUFFER_QUEUE`, the UDP
send / receive / enqueue triple and the qdisc enqueue / dequeue pair.

The decision is made once per packet, not per event: `packet_sample_keep()`
hashes the socket pointer and the packet's own key — the TCP sequence
number (`tcb->seq` at `__tcp_transmit_skb` and `tcp_rcv_established`, the
header's seq at the other stages), the packet buffer's address for UDP
(`udp_packet_key()`: read as a value, so the key is a scalar in the
`tp_btf` programs too, where the skb itself is a BTF pointer the verifier
allows no arithmetic on; the buffer is shared by the clones the stack makes
on the way out) — so every stage of a kept packet is kept and the per-stage
latencies (enqueue → qdisc → send, receive → queue → buffer) stay pairable
under sampling; packet and byte counts scale by N. The qdisc dequeue side
is keyed off the enqueue's `pending_qdisc_packets` entry, so a packet
sampled out at enqueue is never inserted and the pair follows one decision.
No map is touched by the decision: it is a pure function of the event's
fields, checked before `reserve_packet_event()`, so a dropped packet never
reserves ring space; every key read that is not already part of the event
sits inside a `packet_sample_rate > 1` arm the verifier prunes at the
default.

The diagnostic event types are never sampled and keep exact counts at any
N: `PACKET_ZERO_WINDOW_PROBE` / `_ACK`, `PACKET_RTO_TIMEOUT`,
`PACKET_SKB_DROP`, `PACKET_CPU_BACKLOG_DROP`, `PACKET_MEM_PRESSURE`,
`PACKET_TX_QUEUE_STOP` / `_WAKE`, `PACKET_TCP_STATE_CHANGE`.

What a reader of a sampled trace should expect, because the key is
(socket, seq): a retransmission is kept or dropped with its original
(good for pairing), and the pure ACKs of a socket that sends no data all
carry the same seq, so per socket the ACK direction is all-or-nothing —
"counts and bytes scale by N" holds in aggregate over many flows, not per
flow. A diagnostic row (an RTO, a drop) names a seq whose data-path rows
were sampled out N−1 times in N. Three pairings can split under sampling:
the send-side segments of a software-segmented GSO super-skb after the
first hash on their own seq; a segment that `tcp_try_coalesce` merges into
the receive queue's tail has its copy-to-user stage keyed by the tail's
seq; and a UDP datagram whose buffer is reallocated between stages
(`pskb_expand_head` on encapsulation with short headroom, UDP GSO) re-keys
at that point — the qdisc enqueue → dequeue pair is one decision
regardless, through the pending-map entry.

With the knob at its default the verifier prunes the hash arm of every
data-path hook (rodata), so the default configuration costs one rodata
compare per packet event and records every packet as before.

## The rate in the trace, and the honesty line

- `sysinfo.network_packet_sample_rate BIGINT` (schema version 21;
  `SCHEMA_CHANGES.md`): the N the packets recorder ran with, 1 when it ran
  unsampled, NULL when it did not run and in traces from systing < 1.18 —
  the memory tier's `memory_*_sample_rate` precedent. `systing-analyze
  trace info` and the MCP `trace_info` tool report it under `system`.
- A `network recorder: packet sample rate 1/N` stdout line at start when
  N > 1.
- At exit, when the packets recorder dropped anything: `WARNING: packet
  events: <recorded> recorded, <missed> missed (<pct>% of the total) — the
  packet tables are a sample; on this host use --packet-sample-rate <N>`
  (stderr), beside the `Missed packet events` counter the summary already
  printed and the `Missed packet events` counter track the trace already
  carried (polled once a second, the settled total at the end). `<N>` is
  the rate that would have kept the consumer up with the traffic, on top of
  any rate already in force.

## Schema skew, both faces

Forward (a newer reader, an older `sysinfo.parquet`): the import is BY NAME,
the missing column reads NULL — covered by the existing
`test_sysinfo_duckdb_import_without_memory_columns`. Reverse (an older
reader, a newer `sysinfo.parquet`): `parquet_to_duckdb` imports every table
`BY NAME`, and DuckDB's BY NAME requires every source column to exist in
the target, so a reader older than this change fails the whole import of a
trace written by it — the sysinfo row is written by every capture, packets
tier or not. Readers move to this version before writers do.

## Tests

- `packet_sample_rate_defaults_to_every_packet` (main.rs): the flag parses,
  defaults to 1, accepts 0 and N.
- The sysinfo round trip (parquet writer → DuckDB) carries the new column
  (`test_sysinfo_duckdb_roundtrip`, rate 8); `trace info` reads it and reads
  NULL on a database that predates it (analyze/mod.rs); the positional
  `INSERT INTO sysinfo` fixture in `profile_export.rs` carries the new
  column.
- `test_e2e_network_suite` (the state tier only) asserts the column reads
  NULL when the packets recorder did not run — a consumer scaling by it must
  tell "unsampled" from "no packet tables"; the new
  `test_e2e_network_packet_sample_rate` records with the packets tier on at
  rate 4 and asserts the column reads 4 with the never-sampled TCP state
  changes present.
- The load-shape table (`bpf_load_shapes::shape_table`) gains
  `network-packets-sampled` (rate 4): the sampling arm is dead code at the
  default rate, so this row is the load that walks it in `every_shape_loads`.
- fmt / clippy `--all-targets -D warnings` / the lib suite clean at -j2 on
  the prebuilt-libduckdb build; the VM suite results are in the review
  comment.


josefbacik:
## rust-bpf-code-reviewer verdict — `network-packets: --packet-sample-rate N` (head `904030e0850599ff09c37b412896c6ae7c20a495`)

### Summary
The change adds a per-packet sampling gate to the network-packets recorder's ten data-path hooks, records the rate as a `sysinfo` column (schema version 21) and prints an honesty line at exit when the packets recorder dropped anything. The BPF side is a pure function of the event's fields, checked before the ring reserve, with no map touched; the Rust side follows the memory tier's `*_sample_rate` precedent end to end (config → rodata → `SysInfoRecord` → parquet → DuckDB → `trace info` / MCP). The pass found three things and the VM suite a fourth — a verifier refusal at the sampling shape — all fixed in the head this comment is attached to, and the VM suite is green on it.

### Critical Issues
None open.
- Schema gates: `SCHEMA_VERSION` 20 → 21 in `src/duckdb.rs`; `SCHEMA_CHANGES.md` carries the version-21 entry; the PR title carries `[bump minor]`; `Cargo.toml` / `Cargo.lock` are untouched.
- Recorder consistency: no new per-event table; the new column is per-trace (`sysinfo`), so no `utid` question arises.
- No ELF parsing, no Python-version support touched.

### Improvements (found by the pass and the VM suite, fixed in this head)
1. **`tracepoint__qdisc__qdisc_enqueue`: the sampling key's reads were unconditional.** The qdisc event carries no seq of its own, so the key needs `sk->sk_protocol` and, for TCP, the header's seq (`read_tcp_seq_from_skb`) — one to four extra `bpf_probe_read_kernel` calls per qdisc event, paid at the default rate too. Fixed: the reads sit inside `if (tool_config.packet_sample_rate > 1)`, an arm the verifier prunes at the default (rodata), so the default configuration adds nothing to that hook. Every TCP data-path hook already read its seq for the event; those reads only moved ahead of the reserve.
2. **`skb_copy_datagram_iovec` fires for UDP reads as well** (the event is labelled TCP either way — pre-existing, unchanged here), and the sampling key was the "TCP seq" read out of a UDP header, so a UDP packet's copy-to-user stage was sampled independently of its receive / enqueue stages. Fixed: under sampling a UDP socket keys on its packet buffer, the key the UDP stages hash — inside the same pruned arm.
3. **Exit line, nothing recorded:** the suggested rate was `total` (an absurd number when the consumer never kept up). Fixed: the line says nothing was recorded and asks for a much larger rate; the ratio is only computed with `recorded > 0`.
4. **The UDP key was the skb pointer, and the object did not load at the sampling shape.** The first rate-4 capture in the VM suite failed at `skel.load()` (`Invalid argument`): in the two `tp_btf` programs (`net_dev_start_xmit` → `emit_udp_packet_event`, `skb_copy_datagram_iovec`) the skb is a BTF pointer, and the hash multiplies the key — arithmetic the verifier refuses on such a pointer; the kprobe programs' skb arguments are scalars, which is why every other stage loaded. The default-rate loads never see the arm (rodata-pruned), so nothing short of a load at the sampling shape could have caught it. Fixed: `udp_packet_key()` reads the skb's buffer address (`skb->head`) through `bpf_probe_read_kernel` — a scalar in every program, and shared by the clones the stack makes on the way out, so the UDP stages still hash alike; `shape_table()` gains `network-packets-sampled` (rate 4) so `every_shape_loads` walks the arm on every run of the gate, and a rate-4 e2e records with it.

### Suggestions (not applied; noted for the reader)
- **GSO asymmetry, inherent to sampling at the transmit hook.** A TSO super-skb is one `PACKET_ENQUEUE` (keyed on `tcb->seq`) and, when software-segmented at dequeue, several `PACKET_SEND` events whose header seqs step by the MSS — only the first segment shares the enqueue's key, so under sampling the later segments are sampled on their own seq. The per-stage pairing already compares seqs, so nothing new breaks; the count asymmetry (≈ segments per super-skb) is documented rather than worked around.
- **From a second BPF read of this head, three more inherent asymmetries and one statistical note, now stated in the PR body:** (a) the receive mirror of GSO — a segment that `tcp_try_coalesce` merges into the receive queue's tail has its `PACKET_BUFFER_QUEUE` stage keyed by the tail's header seq, so it is kept or dropped with the tail, not with its own RCV / QUEUE rows (GRO super-skbs are one skb through all three stages and unaffected); (b) a UDP datagram re-keys whenever its buffer is reallocated between stages (`skb_cow_head` / `pskb_expand_head` on tunnel or VLAN encapsulation with short headroom, `skb_realloc_headroom`, UDP GSO segmentation) — the qdisc enqueue → dequeue pair stays one decision through the pending-map entry, only the send ↔ qdisc / xmit pairing can split; (c) the diagnostic types are never sampled, so an RTO / drop row names a seq whose data-path rows were sampled out N−1 times in N; (d) the key is (socket, seq), so a retransmission rides with its original and the pure ACKs of a socket sending no data all carry the same seq — per socket the ACK direction is all-or-nothing, and "counts and bytes scale by N" holds in aggregate over many flows, not per flow. A sampled-out enqueue never inserts into `pending_qdisc_packets`, so that map's occupancy falls by the same factor.
- `packet_sample_keep(u64 sk_ptr, u64 key)` takes a `u64` key and the TCP callers pass a `u32` seq (zero-extended) — fine; the helper's comment says which key each protocol uses.
- The hash cannot be exercised from Rust (it is BPF C); the rate-4 e2e asserts the column and the presence of the never-sampled rows, not the kept share. A ratio assertion — data-path rows ≈ 1/N of an unsampled run's on the same workload — is a follow-up worth having once a workload with a stable packet count exists in the suite.
- `(u32)h % rate` divides by a rodata value: at rate 1 the arm and the helper's body are dead code the verifier removes on frozen rodata; at rate N the early return is pruned and the ALU32 MOD is by a known non-zero constant (a runtime zero would be defined anyway — MOD by 0 yields the dividend); the two 64-bit multiplies are `lddw` immediates, and the inlined helper adds ~15 instructions per program. Every operand of the XOR and the multiplies is a scalar in the `tp_btf` programs too (the socket through `bpf_probe_read_kernel(&sk, …, &skb->sk)`, the key through the probe reads) — the first head's refusal was `PTR_TO_BTF_ID` arithmetic, which every kernel refuses alike, not a version difference. All of this is inferred from the source; the VM suite's loads at rate 1 and rate 4 (below) are that kernel's word on it, and the release-flag object's word comes at its own gate.

### Positive Notes
- The decision runs **before** `reserve_packet_event()`: a dropped packet never reserves ring space, so the knob lowers the consumer's load rather than only the trace's size — the actual ceiling on a large host.
- Per-packet keying (socket + the packet's own key) keeps the per-stage latency pairs intact — the property that makes a sampled packet trace still worth reading — and the qdisc pair follows one decision through the enqueue's `pending_qdisc_packets` entry without a second hash.
- The nine diagnostic hooks are untouched, so RTO / drop / zero-window / state-change counts stay exact at any rate.
- The sysinfo plumbing reaches every reader (parquet writer, DuckDB DDL, `trace info`, the MCP tool description, the pre-schema fallback path) with a test at each step, plus the positional fixture that a previous column addition missed.
- The load-shape table's own rule ("a new knob is one new row") earned its keep here: the sampling shape is now verified by the gate on every run, not only by whoever remembers to capture at rate N.

### Mechanics, disclosed
This verdict was produced by executing the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition in-session over the full diff of `904030e0850599ff09c37b412896c6ae7c20a495` (15 files, +442 / −30), by the same session that authored the change. Findings 1–3 came from the pass and finding 4 from the VM suite; all four were fixed and every gate re-run on this head: `cargo fmt --check`, `cargo clippy --no-default-features --all-targets -- -D warnings`, `cargo test --lib --no-default-features` (590 passed, 0 failed) on the prebuilt libduckdb v1.5.4 build; in the VM suite (the repository's v6.12 vmtest kernel, TCG guest): `test_e2e_network_suite`, `test_e2e_network_tw_leg_trampoline_opt_in`, `test_e2e_network_tw_leg_kprobe_fallback` and `test_e2e_network_packet_sample_rate` all pass, and `bpf_load_shapes::every_shape_loads` passes over the whole shape table, the new `network-packets-sampled` row included. Not run at review time: a load of the release-flag object (the release build's compiler and flags) at the recorder's rodata shape — the rig kernel's verifier accepted the object at rate 1 and rate 4; the release object's verdict comes from its own gate before it ships.

PR #243, opened 2026-09-11
analyze: refuse every PIVOT shape but one literal top-level statement; --cgroup: fstatfs the held directory

josefbacik:
Follow-ups to #240, each its own commit.

**1. analyze: accept a PIVOT only as one top-level statement with literal value lists.** A cold review of #240 found three shapes the `query` tool's PIVOT guard still let through to `prepare()`, where DuckDB (1.5.4, the version `libduckdb-sys` pins) hoists any pivot whose value list is not literal into `CREATE TYPE __pivot_enum_… AS ENUM (SELECT DISTINCT …)` + the SELECT and executes the enum scan outside the row cap (`transformer.cpp`: `if (HasPivotEntries()) stmt = CreatePivotStatement(stmt)`; `transform_pivot_stmt.cpp`: an entry per pivot column without literal entries, and for an `IN (<subquery>)` column the enum's SELECT is that subquery): (a) the guard keyed on the statement's first word, so a pivot nested under SELECT / WITH / FROM / a set operation was never examined; (b) `IN (subquery)` is legal grammar and counted as a value list, so an arbitrary user SELECT ran as the enum source; (c) the ON-clause scan tracked only `()` depth, so a CASE or a `[` / `{` list literal contributed a false `IN (`. The guard now whitelists the one shape it fully understands and refuses the keyword everywhere else: `pivot_keyword_count` counts `pivot` / `pivot_wider` as whole words outside quoted runs; the statement is accepted for wrapping only when that count is exactly 1, the keyword is the first word, and `pivot_is_fully_listed` holds — `[` and `{` nest like parens, each `IN (` is a token that carries the first word inside its list and whether the list is empty, every element of the one top-level ON clause needs exactly one list whose first word is not a query keyword (`select` / `with` / `from` / `values` / `table` / `pivot` / `pivot_wider` / `unpivot` / `describe` / `show` / `summarize`) and is not empty, and any `CASE` / `WHEN` / `THEN` / `ELSE` / `END` / `ON` / query keyword at the clause's top level refuses. A nested pivot, a subquery list, a CASE pivot column and an unparenthesised join source are therefore refused rather than read; the refusal text and the tool description say the accepted form in full (a top-level `PIVOT <table or (subquery)> ON <column> IN (<literal values>) USING …`, with a join source written as a parenthesised subquery). The exact alternative — asking DuckDB's own parser with `json_serialize_sql` (a parse, no execution: the nested and subquery-list shapes return `"error": true, "Only SELECT statements can be serialized to json!"`, the literal shape a `SELECT_NODE`) — needs the `json` extension, which the bundled build does not enable (`duckdb/bundled` + `duckdb/parquet` only), so it is named here and not taken.

**2. --cgroup: read the target's filesystem from the directory it holds open.** The fourth cause from #240 read the target path's filesystem with `statfs` on the path while the target itself holds the directory open (the descriptor the cgroup array and the iterator attach use); a path re-pointed between the resolution and the empty-slot read could answer for a different directory. The read is now `fstatfs` on that held descriptor (`cgroup::cgroup_fs_of_fd`), so it always describes the directory the kernel was asked about; the path is still re-resolved for the removed and replaced causes, which is how those show.

**Tests** (`cargo test --no-default-features` on the exact tree; fmt and clippy `--all-targets -D warnings` clean): `analyze::query::tests::test_statement_shape_refuses_pivot_without_an_in_list` (+19 refused — the four nesting forms including a listed nested pivot and a pivot inside a pivot's source, five `IN (<query>)` forms and the empty list, CASE / `[` / `{` / a double list; +4 accepted literal kinds; two `pivot_keyword_count` controls), `analyze::tests::test_query_comments_and_terminators_cannot_bypass_the_cap` (eight of the review's shapes through `AnalyzeDb::query`, the `__pivot_enum` type count 0 after each — on the previous guard the same shapes left one type each behind), `test_cgroup_fs_reads_the_filesystem_magic` (the temp dir and the cgroup2 root read by path and by the held descriptor, the same answer). No BPF, no capture-path or recorder change; the BPF object is untouched.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head at the time of the run (two commits on v1.17.7); verdict as it produced it.

### Summary
The analyze `query` tool's PIVOT guard moves from "find the pivot's ON clause and check it" to "accept exactly one understood shape and refuse the keyword anywhere else": a single top-level `PIVOT` / `PIVOT_WIDER` whose every pivot column names a literal `IN (...)` list, with `[` / `{` nesting tracked and CASE / join / query keywords refused inside the clause. The second commit makes the `--cgroup` empty-slot cause read the target's filesystem from the directory the target holds open (`fstatfs`) instead of the path. Both are small, cast-free, and every new arm has a test; no BPF program, map, recorder or schema is touched.

### Critical Issues
None.

### Improvements
None required. One boundary worth stating in the code comment (already implied by the doc): an `IN (` whose first token is `(` — `IN ((SELECT id FROM t))` — reads as `InList(None, false)` and is accepted; DuckDB parses that as an expression list holding one scalar subquery (entries present, so no enum statement is generated and `prepare` executes nothing), and the value then fails at bind time as a non-constant pivot value. Safe under the guard's purpose; a stricter guard could refuse a list that opens with `(` for symmetry with the query-keyword rule.

### Suggestions
- `pivot_is_fully_listed` (query.rs): `NOT_A_PLAIN_ELEMENT` and `QUERY_KEYWORD` are function-local consts; lifting them to module level beside `StatementShape` would let the tool description and the refusal text cite one source for the accepted form. Cosmetic.
- `cgroup_fs_of_fd` (cgroup.rs): the shared `cgroup_fs_kind(magic)` keeps the two readers from drifting; the `// SAFETY:` notes cover both `unsafe` blocks (a live descriptor for the call, a zeroed `struct statfs` the kernel fills). Nothing to change.
- The caller's comment in `install_cgroup_targets` says why the path is still re-resolved for the removed / replaced causes while the filesystem is read from the held descriptor — keep that split when either read is touched again.

### Positive Notes
- Refusing the keyword everywhere except the one fully understood shape is the right posture for a guard that sits in front of `prepare()`: the refused cases each name the accepted form, and the tests pin nineteen refused shapes (nesting under SELECT / WITH / FROM / a set operation / another pivot, five subquery-list forms, the empty list, CASE and the two literal kinds, a double list) beside four accepted literal kinds and two keyword-count controls.
- The engine-level test drives eight of the reviewed shapes through `AnalyzeDb::query` and checks the temp catalog for `__pivot_enum` types after each — the observable the whole guard exists to keep at zero.
- The `fstatfs` read answers for the directory the kernel was asked about, and the test reads the temp dir and the cgroup2 root both ways and expects the same answer.

PR #244, opened 2026-09-11
duckdb import: project away the columns a newer trace carries instead of failing the import; VFIO/THP doc items

josefbacik:
The parquet→DuckDB import is `INSERT INTO <table> BY NAME SELECT … FROM read_parquet(…)`, and DuckDB's `BY NAME` has two faces. A target column the parquet lacks — an older trace read by a newer systing — is NULL-filled (the case `test_sysinfo_duckdb_import_without_memory_columns` covers). A parquet column the target lacks — a newer trace read by an older systing — failed the WHOLE import with a binder error on the first unknown column, so a trace written after a schema bump was unreadable by every systing before it, and a reader that lags the writer by one release lost the trace rather than the new columns. This PR makes that face degrade instead of fail, and carries the remaining doc items from the second review of the VFIO/THP legs (#219 / #226).

## The import guard

- `duckdb::import_column_list` reads the target table's columns (`duckdb_columns()`, the form the DuckDB→DuckDB import already uses) and the parquet's (`DESCRIBE SELECT * FROM read_parquet(…)`, so the same source form works for one file or a list), compares them case-insensitively as DuckDB binds them, and returns the select list: `*` when every source column is known; the known columns quoted, with one `warning:` line naming the table, the unknown columns and this systing's `SCHEMA_VERSION`, otherwise; `None` — the table skipped — when the source shares no column with the target.
- `ImportOptions { strict_schema }` (default off) turns the projection into a refusal with the same names in the message; `ImportReport { dropped_columns }` carries what was left out, per table in source order.
- Both import sites use it: `duckdb::import_tables` (`parquet_to_duckdb` keeps its signature; `parquet_to_duckdb_with_options` takes the options and returns the report) and `systing-util convert`'s two statement shapes — the multi-file `read_parquet([…])` for inputs that already carry `trace_id` and the per-trace injection — which gains `--strict-schema`.
- The database an import produces is a complete database of the READER's schema: `_schema_version` names it, and a consumer of that database sees nothing amiss. Only the importing process learns, through the report and the warning line, that the trace carried more. Neither the parquet directory nor `_traces` records the writer's schema version (`_traces.systing_version` is the converting binary's — the v12 entry says so), so the recorder's version in the parquet directory stays the planned follow-up; this PR adds no column and no schema version.

## Tests

- `test_sysinfo_duckdb_import_with_unknown_column` (beside the forward-face test): a `sysinfo.parquet` carrying two columns this schema does not have imports on the default with the known values intact, the report naming the table and both columns, `_schema_version` at this schema; the same directory under `strict_schema` refuses with the table, the columns and the schema version in the message.
- `test_import_column_list_projects_unknown_columns`: over parquet files DuckDB writes itself — every column known → `*`; an unknown column → the quoted known list, the report, the strict refusal, and the projected `INSERT … BY NAME` that the importers run; the multi-file list form binds the same way; a source sharing no column is skipped.
- The whole lib suite and `clippy --all-targets -- -D warnings` clean; no BPF program, map or instruction changes (the `src/bpf/` hunks are comments), so no load-shape or guest run.

## The doc items

- The `memory_iommu_hist` comment said its writers are task-context tracepoint programs. They are not (a driver mapping under a softirq fires the same `iommu:map`); the map is safe from those contexts because the bucket lock is a raw spinlock taken with interrupts disabled (`htab_lock_bucket`, read at kernel/bpf/hashtab.c v6.12) and no NMI-context program writes it — the comment now says that.
- What one `iommu:map` / `iommu:unmap` event is, read at drivers/vfio/vfio_iommu_type1.c v6.12: `vfio_iommu_map` issues one `iommu_map` per pinned run PER DOMAIN in the container's domain list (a container with two domains counts every map run twice); on unmap only the first domain is unmapped run by run and each further domain gets one `iommu_unmap` of the whole region. `size_order` is floor(log2) of the run. Both at the leg's comment and in SCHEMA_CHANGES' `memory_iommu` entry.
- The folio-split kretprobe on 6.9+ (`split_huge_page_to_list_to_order`) sees every large-folio split, file and shmem folios included — a superset of the vmstat `thp_split_page` family — and 6.15+'s `folio_split()` bypasses it; at the THP comment.
- The VFIO/IOMMU legs instrument the vfio_iommu_type1 container path only: an iommufd host reads `on` with the module loaded and sees nothing, and `.isra` / `.constprop` clones of the handlers read `nosym` like a missing module; at `MemoryKernelLegs::vfio_off`.
- A tid the kernel reuses inside a capture resolves to the first thread's utid (no mapping is retired), so rows attributed after the reuse — the end-of-capture `anon_huge` sample, a drained `memory_iommu` row — land on the old thread; at `UtidGenerator::get_or_create_utid`.
- A kretprobe that misses its return leaves a VFIO window open (the task's later runs inside the range count until its next VFIO path or the capture's end); a perf-attached kretprobe's miss count is not exported, so it is a documented bound at `memory_vfio_open_window`. The window that cannot be opened was already counted in `memory_iommu_overflow`.
- Two release stamps in SCHEMA_CHANGES read from the tags: the classic-default restoration shipped in 1.17.4 (1.17.0–1.17.3 carried the trampoline default — the text said 1.17.2), and the `split_huge_pmd_locked` funnel tier shipped in 1.17.5.

Per-leg attach (`set_autoattach(false)` + `attach_program_set` per leg, `off:attach` recorded) was verified present from #226 — no change.

### DIFF
--- a/SCHEMA_CHANGES.md
+++ b/SCHEMA_CHANGES.md
@@ -443,6 +443,30 @@ old behaviour — a capture without its CPU stack sampler is not a capture.
 `systing-analyze trace info` (and the MCP `trace_info` tool) report the four
 new fields under `system`.
 
+## Parquet import across schema versions (no schema version change)
+
+The parquet→DuckDB import (`parquet_to_duckdb`, `systing-util convert`) is
+`INSERT INTO <table> BY NAME SELECT … FROM read_parquet(…)`, and DuckDB's
+`BY NAME` has two faces. A target column the parquet lacks — an OLDER trace
+read by a NEWER systing — is filled with NULL, which is how every "NULL in
+traces from systing < 1.N" note in this file works. A parquet column the
+target lacks — a NEWER trace read by an OLDER systing — used to fail the
+whole import with a binder error on the first unknown column, so a trace
+written after a schema bump was unreadable by any systing before it. The
+import now reads the parquet's column list first and projects the unknown
+columns away, one warning line per table naming them and this systing's
+schema version, and the database is a complete database of the READER's
+schema (`_schema_version` says which; the newer columns are simply not in
+it). A caller that wants the old refusal — a test, or a pipeline that must
+not analyse a trace with columns it cannot see — passes
+`ImportOptions { strict_schema: true }` (`systing-util convert
+--strict-schema`), which fails with a message naming the table, the unknown
+columns and the reader's schema version. Neither the parquet directory nor
+`_traces` records the WRITER's schema version (`_traces.systing_version` is
+the converting binary's, see the v12 entry), so the report of what was
+dropped reaches only the importing process — the recorder's version in the
+parquet directory remains the planned follow-up.
+
 ## Schema Version 21 (systing 1.18.0) — 2026-09-04
 
 The network-packets recorder learns to sample, and says so. On a big host
@@ -497,9 +521,9 @@ on the arch syscall wrappers (`__x64_sys_mmap` …, `__arm64_sys_*`,
 the network recorder's TIME_WAIT hooks (`tcp_time_wait`,
 `inet_twsk_hashdance_schedule`, `inet_twsk_deschedule_put`: fentry in
 place of kprobes), as the DEFAULT, on the claim that trampolines detach in
-milliseconds. That claim was wrong, and the next release (unreleased at
-the time of writing; 1.17.0–1.17.2 shipped the trampoline default) makes
-the classic form the default again with the trampoline form opt-in
+milliseconds. That claim was wrong, and 1.17.4 (1.17.0–1.17.3 shipped the
+trampoline default) makes the classic form the default again with the
+trampoline form opt-in
 (`--kernel-hooks trampoline`; `sysinfo.memory_syscall_leg` /
 `network_tw_leg` say which ran). What the kernel does, read at v6.6, v6.12
 and v6.18: a perf-attached
@@ -531,8 +555,8 @@ The tables keep their columns; the new columns are nullable.
 ### Added columns
 - `sysinfo`: added `memory_syscall_leg VARCHAR` — how the mmap/munmap/brk
   hooks attached for the capture: `tracepoint` (the classic tracepoints as
-  the default form, from the correction release), `fentry` (the trampoline
-  set — the default in 1.17.0–1.17.2, opt-in after), `tracepoint:nosym` (the
+  the default form, from 1.17.4), `fentry` (the trampoline
+  set — the default in 1.17.0–1.17.3, opt-in from 1.17.4), `tracepoint:nosym` (the
   classic tracepoints under the trampoline form, because the arch syscall
   wrappers are not in kallsyms — riscv before 6.6), `tracepoint:nobtf` (the
   classic tracepoints under the trampoline form, because vmlinux BTF has no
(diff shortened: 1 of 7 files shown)


josefbacik:
rust-bpf-code-reviewer pass over the full diff, executed in-session by the author against the repository's reviewer definition. The pass read the diff at +461/−30; the committed diff is +465/−31 (7 files): the only lines after the pass are one comment in `systing-util`'s multi-file import (the "list binds to the first file's schema" sentence Suggestion 2 below asks for — four comment lines added and one removed by the reflow, no code), so the gated bytes (fmt / clippy `-D warnings` / the whole lib suite, 592 passed; fmt re-checked bare on the committed tree) are the committed bytes.

### Summary
A reverse-skew guard on the parquet→DuckDB import — one helper that compares the parquet's columns with the target table's and projects the unknown ones away (or refuses under `strict_schema`), used by both import sites — plus the doc items left from the second review of the VFIO/THP legs and two release stamps read from the tags. No schema change; the `src/bpf/` hunks are comments.

### Critical Issues
None.
- Schema pass: no table or column change (`create_schema()`, `src/trace/schema.rs` untouched); `SCHEMA_VERSION` correctly unchanged; `SCHEMA_CHANGES.md` gains a "no schema version change" section on the precedent of the perfetto-callsite one; no `version` edit in `Cargo.toml` / `Cargo.lock`; no `[bump minor]` needed.
- Recorder pass: no recorder or per-event table touched; utid attribution unchanged (its bound is now documented at `get_or_create_utid`).
- BPF pass: `systing_system.bpf.c` changes are three comment blocks (0 non-comment lines in the diff); no program, map or instruction moved, so no load-shape or pairing question arises.
- Binary/ELF pass: not applicable.

### Improvements
None applied — the two candidates were judged not worth a re-gate:
1. `ImportReport::dropped_columns` is a `Vec<(String, Vec<String>)>`; a named pair (`table`, `columns`) would read better at a call site that does more than print it. Today's callers only print or assert it, so the tuple stays until a consumer needs the names.
2. `import_column_list` runs two metadata statements per table per import (`duckdb_columns()` and `DESCRIBE … read_parquet(…)`, the latter a footer read) — ≈45 tables per trace, negligible beside the imports themselves, and reading the parquet's column list is the only way to know before the `INSERT` binds; a cache across the multi-trace importer's per-file loop would save a catalog scan per file, not measurable.

### Suggestions
- The strict refusal's tail ("… or non-strictly to take the known columns only") could name the CLI switch when the caller is `systing-util convert`; the library message stays switch-agnostic on purpose.
- The multi-file source form (`read_parquet([a, b])`) binds to the first file's schema as before; a list mixing two schema generations still fails inside DuckDB's scan as it did before this change — the guard covers the reader-older-than-writer case, not mixed inputs in one statement (documented at the call site's comment).

### Positive Notes
- The projection is decided before any `INSERT` binds, with the same source expression the statement then reads from, so the column set the guard saw is the one the import uses; a source sharing no column with its table is skipped rather than inserted as bare `trace_id` rows.
- The target's columns come from `duckdb_columns()` scoped to `current_database()` / `main` — the form the DuckDB→DuckDB import already uses — so an attached database of the same table names cannot answer for the target.
- The tests cover both faces from the parquet side (a file DuckDB writes itself, a `sysinfo.parquet` written through arrow) and assert the projected statement the importers actually run, the report's order and the strict message's three names.
- The doc items each name the kernel file and version they were read at (`kernel/bpf/hashtab.c`, `drivers/vfio/vfio_iommu_type1.c`, both v6.12), and the two release stamps were read from the tags' ancestry rather than predicted.

PR #247, opened 2026-09-11
memory recorder: an opt-in raw-tracepoint form of the mmap/munmap/brk hooks

josefbacik:
memory recorder: an opt-in raw-tracepoint form of the mmap/munmap/brk hooks (`--kernel-hooks raw-tracepoint`), measured

## What

The memory recorder attaches its mmap/munmap/brk hooks as six classic `tracepoint/syscalls/sys_{enter,exit}_{mmap,munmap,brk}` programs through `perf_event_open` (the default, `--kernel-hooks classic`), or as fentry/fexit trampolines on the arch syscall wrappers (`--kernel-hooks trampoline`, opt-in since #237). This PR adds a third form, `--kernel-hooks raw-tracepoint`: ONE `SEC("tp_btf/sys_enter")` and ONE `SEC("tp_btf/sys_exit")` program dispatching on the syscall number, with the six classic programs kept in the object as the fallback (selected when vmlinux BTF lacks the `btf_trace_sys_enter` / `btf_trace_sys_exit` typedefs, or when the pair fails to attach). It is opt-in, not the default, for the reason measured below.

Rows are byte-identical whichever form produced them: the pair reads its arguments from `pt_regs` exactly as the trampoline form does (`PT_REGS_PARMn_CORE_SYSCALL`) and builds every row through the shared `memory_{mmap,munmap,brk}_{enter,exit}_common` helpers. No schema change; `sysinfo.memory_syscall_leg` gains the VALUES `raw_tracepoint` (the pair ran), `tracepoint:nobtf` (the classic set under the raw form: no typedefs in BTF) and `tracepoint:noraw` (the classic set under the raw form: the pair failed to attach). The network recorder's TIME_WAIT leg is untouched by the new form (it stays on its kprobes, as under `classic`).

## Why the form exists: the detach

A perf-attached tracepoint program's link teardown waits two grace periods in sequence, ONCE PER PROGRAM: first the perf link's release runs `perf_event_detach_bpf_prog()`, which waits a synchronous tasks-trace RCU grace period (`synchronize_rcu_tasks_trace()`) under the global `bpf_event_mutex` (`kernel/trace/bpf_trace.c`, read at v6.6.97 and v6.12.85), then `perf_trace_event_unreg()` → `tracepoint_synchronize_unregister()` — one SRCU plus one RCU grace period (`kernel/trace/trace_event_perf.c`, read at v6.6, v6.12 and master) — so the classic set pays six such rounds on the capture's stop path, serialized. A raw tracepoint's last probe removal takes `tracepoint_remove_func()`'s RCU-state snapshot and frees the probe array through `call_rcu` (`kernel/tracepoint.c`, `release_probes`), with no grace period waited on the stop path — and it never enters `perf_event_detach_bpf_prog()`. The x86 fault leg already made this move (`tp_btf/page_fault_user`, #217); this form makes it for the syscall hooks. Where a host's grace periods are long — many CPUs, a busy RCU, or a tasks-trace reader that runs for seconds (a tasks-trace grace period stalled for minutes turns six programs into tens of minutes of detach) — the memory recorder's six programs are the largest share of the `detach bpf programs` stop phase.

## Why it is opt-in: the entry

Both forms register the same two tracepoints (`sys_enter`, `sys_exit`), so `syscall_regfunc()`'s task walk and the static-key flip at the first registration and last removal are paid either way, and every task carries `TIF_SYSCALL_TRACEPOINT` for the capture under either form. The difference is per syscall: under the classic form the kernel's `perf_syscall_enter()` returns on its enabled-syscall bitmap test for every unhooked syscall; under the raw form every syscall of every task enters the recorder's two programs, whose FIRST act is the syscall-number test (before `trace_task()` and any helper) — but an entry each. Measured (leg B): that entry costs more than the detach saving is worth as a default.

## Measured (a vanilla 6.12.0 guest, 4 vCPU / 8 GB, under TCG — the DIRECTION of a cost difference; magnitudes are the guest's)

- Shape gate: `every_shape_loads` — every shape loads; the new `memory-raw-tracepoint` shape selects 21 programs (the pair beside the classic set), `memory-default` stays at the classic six (19).
- Leg A, `stop-phase: detach bpf programs` per form (one capture of the allocator workload per boot, the same 58 big-mmap + 60 munmap rows under every form): **the raw pair 81 / 70 / 79 / 62 / 66 ms (median 70) vs the classic set 146 / 161 / 141 / 125 / 135 / 137 ms (median 138)** — the pair's detach at ≈0.5× the classic set's on a guest where six grace periods are cheap. Every run PASS with its expected `memory_syscall_leg` (`tracepoint` under the default; `raw_tracepoint` opted in; `tracepoint:noraw` on the forced fallback).
- Leg B, `test_memory_syscall_storm_cost` — a four-thread `getpid()` storm, each second stamped in CLOCK_BOOTTIME and placed against the trace's own `memory_vmstat` attach→detach window (so the traced seconds are known, not guessed), three runs (the last on the object exactly as committed): **1,667–1,729 K calls/s with no capture; 1,398–1,516 K traced under the classic set (−12 to −16 %: the sys_enter/sys_exit slow path and perf's bitmap probe); 818–868 K traced under the raw pair (−49 to −52 %)** — per call, none ≈0.6 µs, classic ≈0.7 µs, raw pair ≈1.2 µs: two BPF program entries per syscall, ≈0.25 µs each on this guest. The pre-attach seconds of the same runs (the capture loading beside the storm) read 1,309–1,425 K/s.

The trade: a stop-path saving paid by the tracer alone, bought with a per-syscall tax on every task of the host for as long as the hooks are attached. That is the wrong default for a continuous capture on a busy host, so the pair is opt-in — for a host whose captures stop slowly and whose workload makes few syscalls.

## Selection and fallback

`probe_kernel_symbols` asks vmlinux BTF for the two typedefs only under the raw form (`KernelSymbols::raw_syscall_tracepoints`; the same one-parse rule `select_rss_stat_prog` applies to `tp_btf/rss_stat`); `probe_memory_kernel_legs` records the pair's state in `MemoryKernelLegs::syscall_raw_off` (`None` = selected; `classic` / `trampoline` / `nobtf` / `noraw` / `forced`); `get_required_bpf_programs` loads the pair only when selected (a tp_btf target resolves at LOAD — an unresolvable one would fail the whole object, which no attach-time fallback can catch); `attach_memory_kernel_legs` attaches trampoline → raw → classic through the existing `attach_leg_with_fallback`, and the testing-only `memory_syscall_force_fallback` switch fails whichever primary set is selected after its first real link, so the drop path is the real one. Every set is `set_autoattach(false)` and attached by hand (`is_memory_kernel_leg_program`).

## Tests

- Unit: `test_memory_syscall_hooks_selection_and_leg_values` covers the three forms (the sets loaded per state, the rendered values, the host probe's verdict on this kernel); `test_network_tw_leg_follows_the_kernel_probe` pins the TIME_WAIT leg to its kprobes under the raw form.
- Integration (`tests/memory_recorder.rs`): `test_memory_recorder_e2e` expects `tracepoint` (the default); NEW `test_memory_syscall_hooks_raw_tracepoint_opt_in` (the pair opted in: `raw_tracepoint` with the rows, or `tracepoint:nobtf` on a kernel without the typedefs); NEW `test_memory_syscall_hooks_raw_fall_back_to_tracepoints` (the forced fallback: `tracepoint:noraw`, the rows still carried); NEW `test_memory_syscall_storm_cost` (leg B, prints its rates and asserts the forms it ran under).
- The shape gate (`tests/bpf_load_shapes.rs`, `every_shape_loads`) gains the `memory-raw-tracepoint` shape so the pair meets the verifier on every run of the gate.

## Precedent (the class of change: a program on every syscall's entry path; teardown-path work)

- In-tree precedent for a program on EVERY syscall: `tracepoint/raw_syscalls/sys_enter` (the marker / `--syscalls` path) — a classic perf tracepoint program that builds a perf sample per syscall; the raw pair is cheaper than that per syscall by construction (a direct call, no sample), and still opt-in.
- #126's class (a hash-map update taking locks with IRQs off on a host-wide path): the pair writes `memory_syscall_scratch` only for the three hooked syscalls, in task context on the syscall path, never from IRQ context; no map written by the `perf_event` (NMI) program is written by the pair.
- Stop-transition work: under the raw form six synchronous grace-period pairs leave the stop path and nothing is added to it.

## Docs

`SCHEMA_CHANGES.md` (schema 20's entry: the new values and the measurement, no version bump), the `sysinfo` column comments in `duckdb.rs`, `trace/models.rs`, `analyze/mod.rs`, `session_recorder.rs`, the `trace_info` MCP description and the `--kernel-hooks` help text.


josefbacik:
rust-bpf-code-reviewer pass over the full diff (11 files, +950/−139), executed in-session by the author against the repository's reviewer definition; the review's edits were applied before the commit and re-gated (fmt / clippy -D warnings / the lib suite; the load-shape gate and the memory-recorder integration tests in a 4-vCPU 6.12.0 guest).

### Summary
A third `--kernel-hooks` form for the memory recorder's mmap/munmap/brk hooks — one `tp_btf/sys_enter` + one `tp_btf/sys_exit` program — gated at load on the two `btf_trace_sys_*` typedefs, attached through the existing fallback helper with the classic six as the fallback, recorded in `sysinfo.memory_syscall_leg`, and left OPT-IN on its own measurement (two program entries on every syscall while attached). The selection is a `match` over the enum, the rows come out of the shared `*_common` helpers, and the tests pin all three forms plus the fallback path. No schema change.

### Critical Issues
None.
- Schema pass: no table or column changes (`create_schema()`, `src/trace/schema.rs` untouched); `SCHEMA_VERSION` correctly unchanged; `SCHEMA_CHANGES.md` carries the new VALUES under the schema-20 entry; no `version` edit in `Cargo.toml` / `Cargo.lock`; no `[bump minor]` needed.
- Recorder pass: no new per-event table; the rows the pair emits are the existing `memory_map` rows through the existing helpers (utid attribution unchanged).
- Binary/ELF pass: not applicable.
- The `perf_event` (NMI-context) program shares no map the pair writes: the pair writes `memory_syscall_scratch` only for the three hooked syscalls in task context.

### Improvements (applied)
1. `systing_syscall_nr()` on the `sys_exit` side read `orig_ax` / `syscallno` / `a7` through `BPF_CORE_READ` — a `bpf_probe_read_kernel` helper call on EVERY syscall exit while the pair is attached, before the number test. The tracepoint's `pt_regs` is a BTF-typed pointer, so the field is a direct CO-RE-relocated load; changed to `regs->orig_ax` (and the arm64 / riscv fields), one instruction instead of a helper call on the hottest path of the form.
2. The attach-time fallback cause was computed twice in `attach_memory_kernel_legs` (once per primary set); folded into one `cause` selection (`forced` / `noraw` / `notramp`) assigned to the field of whichever primary set was selected, so the two opt-in forms cannot drift.

### Suggestions
- `test_memory_syscall_storm_cost` prints its rates and asserts only the forms and a minimum of three traced seconds; it is a measurement, not a bound, and says so in its doc comment — if a future change wants a regression guard, a ratio ceiling (raw pair vs classic) would need a KVM guest, since TCG magnitudes drift with the host.
- The `KernelHooks` doc comment now carries the measured figures for both opt-in forms; when either form is re-measured on a different guest the two sentences should move together with `SCHEMA_CHANGES.md`'s.

### Positive Notes
- The pair's programs test the syscall number as their FIRST instruction and return before any helper or task filter; the arguments are read exactly as the trampoline programs read them, so a row is byte-identical whichever form produced it (verified by the integration tests: 58 big-mmap + 60 munmap rows under every form).
- The load-time gate on the typedefs mirrors the tree's existing rule for `tp_btf/rss_stat` — an unresolvable `tp_btf` target fails the whole object, so the pair is never loaded where it cannot resolve, and the classic set runs with the reason recorded (`tracepoint:nobtf`).
- The testing-only `memory_syscall_force_fallback` switch now fails whichever primary set is selected after its first real link, so the fallback e2e drops a real link rather than simulating one.
- The storm test places its seconds against the trace's own `memory_vmstat` attach→detach window rather than a guessed sleep — the traced seconds are known, and the pre-attach seconds of the same run are printed as the in-run reference.

josefbacik:
Rebased onto main at v1.19.0 after #259 merged. One conflict, src/mcp.rs: the trace_info tool description that both this PR and #259 extend — resolved as the union of both sides (main's trace_versions sentence kept; this PR's raw_tracepoint / tracepoint:noraw clause inserted at its place). No other file changed in the rebase; the commit message is unchanged. Range-diff against the previous head 072ce5dd shows only that one description line.

PR #248, opened 2026-09-11
network-packets: take the parquet encode and the fan-in channel off the packet consumer

josefbacik:
## Summary

On a busy host the `network-packets` recorder loses most of its events to its own userspace consumer, not to the kernel ring buffers. Two 90-second captures on 192-vCPU hosts (`--only-recorder network --only-recorder network-packets --ringbuf-size-mib 256`, i.e. 64 rings × 32 MiB per family) reported `Missed packet events` of 81 M and 93 M while `Missed network events` on the same runs was 0. A ring-size step cannot help: at 216 B per event the 2 GiB budget is ≈10 s of that deficit.

This PR removes the two structural ceilings on the consumer side. No schema change; the parquet file that comes out has the same schema, encodings, compression, row-group bounds and metadata as before, so every reader is unchanged.

## What was measured

The ignored benchmarks added here (`network_recorder::packet_consumer_bench`, `parquet::writer::packet_encode_bench`; run with `cargo test --release --lib <name> -- --ignored --nocapture`) split the per-event cost on one 64-core host. The absolute numbers are that host's; the split is the finding.

Before:

- building one packet record: **177 ns/event** (5.6 M/s on one thread) — not the ceiling;
- the production path (record build → `SharedCollector` → `StreamingParquetWriter`): **865 ns/event = 1.16 M/s**, of which the ten inline `flush_network_packets` calls over 2 M events took 1,434 ms of 1,729 ms — **83 %** of the consumer's time was building the arrow batch and running the parquet encode for 200,000 rows on the consumer thread, under both the recorder mutex and the collector mutex, while every ring poller sat blocked on the channel (worst single stall 229 ms);
- the encode itself: `build_network_packet_batch` 267 ns/row + `ArrowWriter::write` 447 ns/row across ~45 columns on one thread, and the codec is not the cost (UNCOMPRESSED 395, LZ4_RAW 426, ZSTD-1 462, ZSTD-3 447 ns/row — ZSTD-3 with the existing encodings is both the fastest compressed form and the smallest file, so the writer properties stay as they are);
- the single `sync_channel(100_000)` that all 64 ring pollers sent into: 11 M events/s with one producer, 2.5 M/s with 8, **1.5 M/s with 64** — a ceiling from send-side contention alone.

## What changed

1. **One channel and one consumer thread per packet ring** (`packet_rec_{i}`), the shape the sched rings already use. No shared channel, no contention. Each ring's channel holds 8,192 events (two consume batches, ≈1.8 MB per ring): the single channel held 100,000, and the same capacity per ring would commit ≈1.4 GB at capture start on a 64-ring host, because the bounded channel initialises every slot when it is created.
2. **The record build runs with no lock.** `NetworkRecorder::prepare_packet_batch` turns a consumer's batch of events into packet records (id unassigned) plus the socket sightings they need; the consumer then takes the recorder once per batch for `append_packet_batch` — the first-sight socket records, the ids in arrival order, the minimum timestamp — and hands the whole batch to the collector through a new `RecordCollector::add_network_packet_batch` (one collector lock per batch instead of one per record; the default implementation loops, `SharedCollector` and the parquet writer override it). `handle_packet_event` is the same two calls with a batch of one, so the row-level behaviour is unchanged, and a test feeds the same mixed batch through both paths and compares the packet rows, socket rows and minimum timestamp.
3. **The parquet encode moves off the consumer into an encode lane** (`src/parquet/lane.rs`). `flush_network_packets` hands the flushed `Vec` of rows over a bounded channel and returns with a fresh buffer; the lane thread builds the arrow batch in parallel slices and fans the leaf columns out to a fixed set of column-encoder threads through parquet's parallel column-writer API (`get_column_writers` / `compute_leaves` / `ArrowColumnWriter`), closing a row group at the writer's `max_row_group_size` exactly where `ArrowWriter` closed one and writing the same `ARROW:schema` metadata. The lane's queue is bounded to two batches on purpose: a lane that cannot keep up shows as missed events in the trace, never as unbounded memory. A consumer that finds that queue full waits on it while holding the recorder lock, so the other consumers and the syscall path wait behind it exactly as they waited behind the inline flush — no new lock class, only a shorter and rarer wait. Errors surface on the next flush and at `finish`.

After, on the same host: the single-event path costs **267 ns/event** with **no consumer-side stall over 1 ms** (worst 0.3 ms), and the per-ring batch path **195 ns/event per consumer thread** (47 ns lock-free, 148 ns under the recorder lock), with the lane draining in ~15 ms after the last event.

## Tests

- `parquet::lane::tests`: rows pushed in several batches come back whole and in order with the row groups closed at the configured bound and the arrow schema in the file's metadata; an empty lane writes a well-formed empty file; a failing sink is reported at `finish`.
- `network_recorder::packet_batch_tests`: the batch path produces the same packet rows, socket rows and minimum timestamp as the single-event path on a mixed batch (three sockets, one port-less first sighting that is upgraded, an event with no socket id, an event of an unknown type); ids stay contiguous across batches; the socket record is emitted once per socket plus the one upgrade; an empty batch is a no-op.
- The existing socket-record emission tests and the full lib suite pass unchanged (596 passed, 8 ignored = the benchmarks); `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean.
- Two integration targets run in a VM on the repo's 6.12 test kernel: `analyze_commands` (records a 3-second trace with the packets recorder on, validates the DuckDB, and runs every `systing-analyze` subcommand — `network socket-pairs` asserts a non-empty pair list, so packet rows went through the per-ring consumers, the encode lane and the parquet import) and `trace_validation`'s `test_e2e_network_suite` (the state tier on the shared recorder), both PASS. The VM cannot generate a packet rate that stresses the consumers; the before/after under real load is measured on the hosts that showed the loss.

## Open choices for the reader

- The lane is adopted for `network_packet` only. `sched_slice`, `thread_state` and `stack_sample` flush the same way and would take the same lane with a one-line change each; left for a follow-up so this change stays on the packets path.
- The remaining serialized cost per event is the recorder lock (the socket probe, the id, the buffer extend); reusing the flushed buffer's allocation instead of a fresh `Vec` per flush is the next trim if a host still runs ahead of it.
- Kernel-side work (a leaner per-packet event, folding the same-CPU stage pairs) is a separate change with reader implications and is not in this PR.

### DIFF
--- a/src/network_recorder.rs
+++ b/src/network_recorder.rs
@@ -296,6 +296,44 @@ impl SocketMetadata {
     }
 }
 
+/// A consumer's batch of packet events prepared OUTSIDE the recorder lock
+/// ([`NetworkRecorder::prepare_packet_batch`]): the parquet records with
+/// their `id` unassigned, and beside each the socket sighting the recorder
+/// needs to emit the socket record on first sight. Appended under the lock
+/// by [`NetworkRecorder::append_packet_batch`].
+pub struct PreparedBatch {
+    records: Vec<NetworkPacketRecord>,
+    sockets: Vec<SocketSighting>,
+}
+
+impl PreparedBatch {
+    /// Packet records in the batch (events without a socket id or of an
+    /// unknown type were dropped at preparation).
+    pub fn len(&self) -> usize {
+        self.records.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.records.is_empty()
+    }
+}
+
+/// What a packet event says about its socket: the tuple the socket record
+/// carries, whether it is complete (see [`socket_tuple_is_complete`]), the
+/// namespace and the sighting time.
+struct SocketSighting {
+    socket_id: SocketId,
+    protocol: u32,
+    af: u32,
+    src_addr: [u8; 16],
+    src_port: u16,
+    dest_addr: [u8; 16],
+    dest_port: u16,
+    complete: bool,
+    netns_inum: u64,
+    ts: u64,
+}
+
 pub struct NetworkRecorder {
     pub ringbuf: RingBuffer<network_event>,
 
@@ -325,7 +363,7 @@ pub struct NetworkRecorder {
     /// arrives with the complete tuple, ONE upgraded record is emitted
     /// so the trace carries the real endpoints. Readers keep one row
     /// per socket by preferring the complete row.
-    seen_sockets: HashMap<SocketId, bool>,
+    seen_sockets: SeenSockets,
     /// Collector for streaming records during recording
     streaming_collector: Option<Box<dyn RecordCollector + Send>>,
     /// Next record ID counters for streaming
@@ -355,6 +393,34 @@ fn socket_tuple_is_complete(af: u32, src_port: u16, dest_addr: &[u8; 16], dest_p
     src_port != 0 && !(dest_port == 0 && parse_ip_addr(af, dest_addr).is_unspecified())
 }
 
+/// The seen-socket map, probed once per packet event under the recorder
+/// lock: a `u64` socket id hashed by one multiplication (Fibonacci hashing)
+/// instead of SipHash. The ids are BPF-assigned counters, so the identity
+/// would cluster hashbrown's tag bits; the multiply spreads them.
+type SeenSockets = HashMap<SocketId, bool, std::hash::BuildHasherDefault<SocketIdHasher>>;
+
+#[derive(Default)]
+struct SocketIdHasher(u64);
+
+impl std::hash::Hasher for SocketIdHasher {
+    fn finish(&self) -> u64 {
+        self.0
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        // Only `u64` keys reach this hasher; anything else would be a bug
+        // in how the map is keyed, and is hashed byte-wise so it still
+        // works.
+        for &b in bytes {
+            self.0 = (self.0 ^ u64::from(b)).wrapping_mul(0x9E37_79B9_7F4A_7C15);
+        }
+    }
+
+    fn write_u64(&mut self, id: u64) {
+        self.0 = id.wrapping_mul(0x9E37_79B9_7F4A_7C15);
+    }
+}
+
 /// Decide whether a NetworkSocketRecord should be emitted for this
 /// sighting of `socket_id`, updating the seen-map. True on first sight,
 /// and ONCE more as an upgrade when the first record was emitted from a
@@ -365,7 +431,7 @@ fn socket_tuple_is_complete(af: u32, src_port: u16, dest_addr: &[u8; 16], dest_p
 /// only ever shows degraded tuples (a real listener, an unconnected UDP
 /// socket) emits exactly one record.
 fn should_emit_socket_record(
-    seen_sockets: &mut HashMap<SocketId, bool>,
+    seen_sockets: &mut SeenSockets,
     socket_id: SocketId,
     complete: bool,
 ) -> bool {
@@ -387,7 +453,7 @@ impl NetworkRecorder {
             hostname_cache: HashMap::new(),
             resolve_addresses,
             min_ts: None,
-            seen_sockets: HashMap::new(),
+            seen_sockets: SeenSockets::default(),
             streaming_collector: None,
             next_syscall_id: 1,
             next_packet_id: 1,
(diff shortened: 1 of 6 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `network-packets: take the parquet encode and the fan-in channel off the packet consumer` (head `ac7c5b4383152fb60f70deeb4dc6a2011821482e`)

Mechanics: the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session by the author over the full diff against `v1.17.8` (`6b106c8`, six files, +1397/−126), in the definition's five passes, with `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and the lib test suite (596 passed, 8 ignored benchmarks) as the mechanical half, plus two integration targets run in a VM (`analyze_commands`, which records with the packets recorder on and validates the DuckDB, and `trace_validation`'s network suite). The issues found were fixed before this head; they are listed so the reader can check them.

### Summary

A contained restructuring of the network-packets consumer path with no BPF, schema or reader change: per-ring SPSC channels and consumers, the record build outside the recorder lock with one batch append under it, and the `network_packet` table's parquet encode moved to a lane with parallel column writers that produces a byte-compatible file (same schema metadata, encodings, compression and row-group bounds). The measurements the PR rests on are the two ignored benchmark modules it adds; the tests cover the equivalence of the new batch path with the old single-event path and the lane's round trip, empty case and error path.

### Critical Issues

None.

- Schema & versioning: no table definition, `SCHEMA_VERSION`, `SCHEMA_CHANGES.md` or `Cargo.toml` version change — a patch bump, no `[bump minor]` needed. The lane writes the `ARROW:schema` metadata `ArrowWriter` wrote (`add_encoded_arrow_schema_to_metadata`) and closes row groups at `max_row_group_size`; the round-trip test asserts the schema read back equals the arrow schema and the group count.
- Recorder consistency: `network_packet` rows carry no thread attribution by design (the packet hooks fire in softirq context; pre-existing), so the `utid` rule does not apply; `NetworkRecorder` keeps its `Arc<UtidGenerator>` for the syscall path.
- BPF: untouched. Binary/ELF parsing: none.

### Improvements (found in the pass, fixed in this head)

1. `parquet/lane.rs` — `RowGroup::handles` was a four-deep generic type (clippy `type_complexity`); now the `IndexedChunk` alias.
2. `parquet/lane.rs` — an encoder thread's failure surfaced as the generic "a column encoder stopped" from the send that found the closed channel, hiding the encoder's own error; `RowGroup::abort` now joins the encoders and reports the first real error.
3. `network_recorder.rs` — `maybe_emit_socket_record` destructured the sighting into ten locals and re-tupled them; it reads the fields directly now.
4. `network_recorder.rs` (test) — `packet_event::default()` followed by field assignments (clippy `field_reassign_with_default`); now the struct-update form.
5. `systing_core.rs` — the per-ring packet channels first inherited `CHANNEL_CAPACITY` (100,000 events) each; a bounded channel initialises every slot when it is created, so 64 rings would have committed ≈1.4 GB at capture start where the single channel committed 22 MB. Now `PACKET_CHANNEL_CAPACITY` = two consume batches (8,192 events, ≈1.8 MB per ring), with the trade-off stated in its comment (a second reader's finding).

- `run_lane` takes eight parameters under a `too_many_arguments` allow; a `LaneConfig` struct would read better if the lane grows a second adopter.
- `flush_network_packets` replaces the flushed buffer with a fresh `Vec::with_capacity(batch_size)` (~72 MB) each flush; the page faults of first touch land on the consumer thread. Returning the encoded batch's buffer to the writer over a second channel would reuse it. Measured as tolerable here (148 ns/event under the lock in total); worth doing when the next host class runs ahead of it.
- `encode_workers()` derives the encoder count from `available_parallelism` (2–8); a command-line knob is not warranted until a host class shows the lane, not the consumers, as the ceiling.
- The lane is adopted for `network_packet` only; `sched_slice`, `thread_state` and `stack_sample` flush identically and are one-line adoptions each — a follow-up keeps this change on the packets path.

### Positive Notes

- `prepare_packet_batch` is a pure function of the events, which is what lets the consumers run in parallel without widening any lock; the batch/single equivalence test pins the row-level behaviour (socket-record emission and its upgrade rule, ids, minimum timestamp) so the refactor cannot drift from the old path.
- The lane's bounded queue makes overload visible as missed events rather than memory growth, and its `Drop` finishes the lane so a writer dropped without `finish` still closes the file.
- The benchmarks are ignored by default and self-describing; the encode split (codec vs column encode) is the datum that justified parallel column writers over a codec change.

PR #249, opened 2026-09-11
pystacks: skip CPython 3.12+ entry frames instead of emitting "[Frame Error]"

josefbacik:
## Summary

Python stack walking on CPython 3.12+ emits a `[Frame Error]` placeholder
frame at every C → Python re-entry and spends a slot of the stack budget
on it. The placeholder is CPython's entry ("shim") frame; this PR makes
the walker step over those frames and bounds the walk on emitted symbols
so deep stacks get their depth back.

Measured on production Python 3.13 workloads (a 20-minute window of CPU
samples from two large production populations): 99 % of Python-bearing stacks carry
at least one `[Frame Error]` frame, typical stacks carry 5–11, deep ones
up to 126; the placeholder is 20–34 % of all Python frames recorded; and
1.5–4.6 % of Python stacks sit at the 127-frame budget with shim frames
inside, i.e. their real outer frames were truncated away. An example
stack (trimmed) shows where they land:

```
…
:tenacity.__init__:Retrying.__call__ (python) [__init__.py:473]
:[Frame Error] (python) [unknown]
:tenacity.__init__:BaseRetrying.wraps.<locals>.wrapped_f (python) [__init__.py:331]
…
:requests.sessions:Session.send (python) [sessions.py:706]
:[Frame Error] (python) [unknown]
:requests.sessions:Session.request (python) [sessions.py:592]
…
```

## The mechanism

Since 3.12, `_PyEval_EvalFrameDefault` pushes an entry frame onto the
`previous` chain each time it is entered — every call into Python from C
(`PyObject_Call`, a bound method's `__call__`, `functools.partial`,
decorators implemented in C, …). That frame is owned by the C stack —
`owner = FRAME_OWNED_BY_CSTACK` (3) in 3.12 and 3.13; 3.14 renames the
value `FRAME_OWNED_BY_INTERPRETER` and adds `FRAME_OWNED_BY_CSTACK` = 4 —
and its `f_executable` is `None` (`Include/internal/pycore_frame.h`,
`pycore_interpframe_structs.h`).

`pystacks_get_frame_data` walked every frame on the chain as a
code-bearing frame: the code pointer read yields `None`, the qualname
pointer read from it fails the address check, and `get_names` writes
`[Frame Error]` with an unknown file, after which `add_symbol_to_buffer`
spends one of the `stack_max_len` (127) slots. The `owner` offset was
already resolved per version (`PyFrameObject_owner` =
`PY_INTERP_FRAME_OWNER`: 70 / 70 / 74 for 3.12 / 3.13 / 3.14) but never
read by the BPF side.

## The fix

Two commits, one BPF file (`src/pystacks/bpf/pystacks.bpf.c`), no Rust,
no schema, no map, no new program:

1. **Skip shim frames.** For interpreter frames on 3.12+,
   `pystacks_get_frame_data` reads the frame's `owner` first and, when it
   is `FRAME_OWNED_BY_CSTACK` or above, marks the frame skipped: no code
   object or name reads, no symbol emitted, the walk follows `previous`
   as before. Frames owned by a thread, a generator or a frame object
   are real frames and unchanged; Python ≤ 3.11 (no shim frames on the
   chain) and the shadow-frame path are untouched.
2. **Bound the walk on emitted symbols.** The loop terminates on
   `py_msg->stack_len < stack_max_len` rather than on visited frames, so
   a stack of N real frames interleaved with shim frames still yields up
   to `stack_max_len` symbols; the visited-frame bound stays for the
   verifier at twice the symbol budget (a shim frame never follows
   another on the chain — each hosts at least one real frame above it —
   so at most every other visited frame is skipped).

## Validation

- **Compile**: the pystacks BPF object compiled with the build.rs recipe
  (`-g -O2 -target bpf -mcpu=v3 -fwrapv`, the pinned flags) for x86_64
  and aarch64 at the merge base and on this branch; the branch's
  `pystacks_get_frame_data` carries the one-byte owner read and the `< 3`
  compare feeding the skip, `walk_and_load_py_stack` the emitted-symbol
  test. `cargo fmt -- --check` clean (no Rust changed).
- **Load and runtime (VM, observed on this branch's bytes applied to the
  v1.17.7 source — the pystacks tree is identical at v1.17.8)**: an x86_64
  KVM guest on a 6.12.93 kernel with BTF. Load: `every_shape_loads` 34 of
  34 shapes loaded, no verifier refusal; the pystacks-on shape loaded in
  270.8 ms (the unpatched control in the same recipe: 431.3 ms on a
  different host — no cost signal from commit 2's doubled visited bound).
  Runtime: a python 3.13.13 workload (`spin.py`) under
  `systing --duration 3 --sw-event --collect-pystacks`, patched vs
  unpatched — samples carrying a `[Frame Error]` frame **0 of 3,129**
  (unpatched: 2,998 of 3,139, i.e. every Python-bearing sample); samples
  with a `.py` frame 3,004 of 3,129 (96.0 %) vs 2,998 of 3,139 (95.5 %);
  the hot function `spin_hot_loop_m4` in 3,004 samples with both source
  lines vs 2,998; `with_pystack=3004 without_pystack=125 unknown=0`
  vs `2998 / 141 / 0`; the walker's per-event `stack_len` 2 vs 3 (the
  shim slot gone); distinct Python frames 3 vs 4 (the placeholder entry
  gone; the native frame counts differ by run noise only).
- **aarch64 (observed)**: the same patch on top of the aarch64 pystacks
  TLS fix (a separate PR; without it aarch64 yields no Python frames at
  all), in an emulated aarch64 guest on the same 6.12.93 kernel with BTF:
  `every_shape_loads` 34 of 34, no verifier refusal, the pystacks-on
  shape in 15.1 s under emulation (17.3 s with the TLS fix alone); the
  python 3.13.13 workload — samples carrying a `[Frame Error]` frame
  **0 of 2,691** (TLS fix alone: 2,637 of 2,905, every Python-bearing
  sample); `.py` frames 2,377 of 2,691 (88.3 %) vs 2,637 of 2,905
  (90.8 %); `spin_hot_loop_m4` in 2,377 samples with both lines vs 2,637;
  `stack_len` 2 vs 3 per event; distinct Python frames 3 vs 4. The
  Python-bearing share moves +0.5 pt on x86_64 and −2.5 pt on emulated
  aarch64 between runs of different length — run noise in both
  directions; the one real reclassification is a sample caught mid-entry
  whose chain holds only a shim frame, which now reads as no Python
  stack instead of a lone `[Frame Error]`.
- **First contact after release**: the `[Frame Error]` share of Python
  frames on the populations above goes to 0 and the at-budget share
  drops.

Note: the `[Frame Error]` text stays for the case it was written for — a
frame whose code object genuinely cannot be read.

### DIFF
--- a/src/pystacks/bpf/pystacks.bpf.c
+++ b/src/pystacks/bpf/pystacks.bpf.c
@@ -164,11 +164,29 @@ struct sample_state_t {
   uint64_t cur_cpu;
   void* frame_ptr;
   bool sync_use_shadow_frame;
+  /* The frame just visited by pystacks_get_frame_data was a CPython 3.12+
+   * entry ("shim") frame: nothing was read from it and no symbol is to be
+   * emitted for it. */
+  bool frame_skipped;
   char long_file_name[BPF_LIB_FILE_NAME_TRYGET];
   struct pystacks_symbol sym;
   struct pystacks_line_table linetable;
   int32_t lasti;
 };
+
+/*
+ * _PyInterpreterFrame.owner values whose frame carries no code object.
+ * CPython 3.12+ pushes an entry ("shim") frame onto the frame chain for every
+ * entry into _PyEval_EvalFrameDefault, i.e. at every C -> Python re-entry
+ * (PyObject_Call from C, a bound method's __call__, functools.partial, ...):
+ * owner = FRAME_OWNED_BY_CSTACK (3) and f_executable = None. 3.14 renames
+ * that value FRAME_OWNED_BY_INTERPRETER (3) and adds FRAME_OWNED_BY_CSTACK
+ * (4); both are non-Python frames. The values below it (THREAD 0,
+ * GENERATOR 1, FRAME_OBJECT 2) are real frames. Reading the (None) code
+ * object of a shim frame yields a "[Frame Error]" placeholder and spends a
+ * slot of the stack budget, so such frames are skipped instead.
+ */
+#define PYSTACKS_FIRST_NON_PYTHON_FRAME_OWNER 3
 
 struct {
   __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
@@ -678,10 +696,33 @@ __noinline bool pystacks_get_frame_data(int pid) {
     return false;
   }
 
-  void* code_ptr =
-      get_code_ptr(state->frame_ptr, offsets, use_shadow_frame, task);
+  /*
+   * CPython 3.12+ interleaves entry ("shim") frames with the real ones on
+   * the `previous` chain; they carry no code object (see
+   * PYSTACKS_FIRST_NON_PYTHON_FRAME_OWNER). Read the frame's owner first
+   * and step over such a frame without reading names for it.
+   */
+  state->frame_skipped = false;
+  if (!use_shadow_frame && offsets->PyVersion_major >= 3 &&
+      offsets->PyVersion_minor >= 12 &&
+      offsets->PyFrameObject_owner != BPF_LIB_DEFAULT_FIELD_OFFSET) {
+    uint8_t owner = 0;
+    if (bpf_probe_read_user_task(
+            &owner,
+            sizeof(owner),
+            state->frame_ptr + offsets->PyFrameObject_owner,
+            task) == 0 &&
+        owner >= PYSTACKS_FIRST_NON_PYTHON_FRAME_OWNER) {
+      state->frame_skipped = true;
+    }
+  }
 
-  get_names(state, state->frame_ptr, code_ptr, use_shadow_frame, task);
+  if (!state->frame_skipped) {
+    void* code_ptr =
+        get_code_ptr(state->frame_ptr, offsets, use_shadow_frame, task);
+
+    get_names(state, state->frame_ptr, code_ptr, use_shadow_frame, task);
+  }
 
   int ret_code = 0;
 
@@ -869,7 +910,9 @@ __hidden int walk_and_load_py_stack(
   for (; i < stack_max_len && i < BPF_LIB_MAX_STACK_DEPTH &&
        (last_frame_read = pystacks_get_frame_data(pid));
        ++i) {
-    add_symbol_to_buffer(py_msg);
+    if (!state->frame_skipped) {
+      add_symbol_to_buffer(py_msg);
+    }
   }
 
   set_py_stack_status(
--- a/src/pystacks/bpf/pystacks.bpf.c
+++ b/src/pystacks/bpf/pystacks.bpf.c
@@ -907,7 +907,15 @@ __hidden int walk_and_load_py_stack(
   bool last_frame_read = false;
   int pid = task ? BPF_CORE_READ(task, pid) : 0;
 
-  for (; i < stack_max_len && i < BPF_LIB_MAX_STACK_DEPTH &&
+  /*
+   * Bound the walk on emitted symbols rather than on visited frames, so a
+   * stack of N Python frames interleaved with shim frames still yields up
+   * to stack_max_len symbols. The visited-frame bound stays for the
+   * verifier; a shim frame never follows another shim frame on the chain
+   * (each hosts at least one real frame above it), so at most every other
+   * visited frame is skipped and twice the symbol budget covers the walk.
+   */
+  for (; i < 2 * BPF_LIB_MAX_STACK_DEPTH && py_msg->stack_len < stack_max_len &&
        (last_frame_read = pystacks_get_frame_data(pid));
        ++i) {
     if (!state->frame_skipped) {
@@ -919,7 +927,7 @@ __hidden int walk_and_load_py_stack(
       &py_msg->stack_status,
       (long)state->frame_ptr,
       last_frame_read,
-      (i == stack_max_len - 1) /* is_final_iteration */);
+      (py_msg->stack_len == stack_max_len - 1) /* is_final_iteration */);
 
   return py_msg->header.len;
 }


josefbacik:
## Code review

v1.17.8): one BPF source file, two commits, +57/−6, no Rust.

### Summary

Two small BPF-only changes to the Python stack walker. (1) CPython 3.12+
pushes an entry ("shim") frame onto the `previous` chain at every
C → Python re-entry (`owner = FRAME_OWNED_BY_CSTACK`, `f_executable =
None`; 3.14: `FRAME_OWNED_BY_INTERPRETER` = 3 and `FRAME_OWNED_BY_CSTACK`
= 4); the walker read its (None) code object and emitted a
`"[Frame Error]"` symbol per shim frame while spending a slot of the
stack budget. The walker now reads the frame's `owner` first (the offset
already resolved per version as `PyFrameObject_owner`, populated for 3.12,
3.13 and 3.14) and steps over an `owner >= 3` frame without reading
names or emitting a symbol. (2) The walk is bounded on emitted symbols
instead of visited frames, with a visited bound of twice the symbol
budget for the verifier, so deep stacks recover the depth the shim
frames used to consume. Measured on production Python 3.13 workloads
the placeholder was 20–34 % of all Python frames recorded and 1.5–4.6 %
of Python stacks hit the budget with shim frames inside.

### Critical Issues

None.

  changes), so no `SCHEMA_VERSION` bump, `SCHEMA_CHANGES.md` entry, or
  `[bump minor]` title token required. The parquet columns and the frame
  encoding are unchanged; only which frames are emitted changes.
- `Cargo.toml` version: untouched (bump-on-merge workflow owns it).
- Recorder consistency / Python version support criteria: the change is
  gated on `PyVersion_minor >= 12` beside the existing
  `PyFrameObject_owner != BPF_LIB_DEFAULT_FIELD_OFFSET` guard, so Python
  ≤ 3.11 (no shim frames on the chain, `owner` unpopulated) and the
  Cinder shadow-frame path (`use_shadow_frame`) are byte-for-byte on
  their old behaviour.
- Memory safety: the `owner` read is one byte via
  `bpf_probe_read_user_task` into a stack local; a failed read leaves the
  frame unskipped (the old path). `frame_skipped` lives in the per-CPU
  `pystacks_state_heap` entry and is written by `pystacks_get_frame_data`
  before the caller reads it in the same iteration; the leaf-frame path
  (`read_leaf_frame`, compiled out — `STROBELIGHT_READ_LEAF_FRAME` is
  defined nowhere) does not touch it. `get_names` is skipped for a shim
  frame, so `state->sym` / `lasti` keep the previous frame's contents —
  harmless, because `add_symbol_to_buffer` is not called for that frame
  and `get_names` zeroes the symbol at the next real frame.
- The loop bound: `2 * BPF_LIB_MAX_STACK_DEPTH` (254) visited frames is
  a compile-time constant, so the verifier still sees a bounded loop;
  `py_msg->stack_len` (`uint64_t`, zeroed with the message header before
  the walk) is compared against `stack_max_len` as in
  `add_symbol_to_buffer`, so no more than `stack_max_len` symbols are
  ever written. The verifier's cost of the doubled iteration count is
  not readable from the source; it was measured — see Mechanics (34 of
  34 shapes load, the pystacks-on shape in 270.8 ms).
- Maps and program contexts: no map is added or written, no `SEC()` is
  added; the change runs inside the existing walk in the same program
  context and adds one user-memory read per visited frame.

### Improvements

- `is_final_iteration` now reads `py_msg->stack_len == stack_max_len - 1`
  to keep the previous expression's meaning against emitted symbols;
  `stack_status` is not consumed by the Rust side today (no reader of
  `PYSTACKS_STATUS_TRUNCATED` / `stack_status` in `src/pystacks/*.rs`),
  so nothing downstream changes. The expression itself is odd inherited
  logic (a chain that ends exactly one short of the budget reads as
  truncated) — worth revisiting if the status is ever surfaced.
- The version gate `PyVersion_major >= 3 && PyVersion_minor >= 12`
  mirrors the file's existing `>= 11` idiom (it would misread a
  hypothetical 4.0); a helper comparing (major, minor) as a pair would
  fix every site at once — a follow-up, not this change.
- A per-stack count of skipped frames (in `pystacks_message`) would make
  the shim share observable without querying the stored traces;
  deliberately not added here since it changes the message layout.

### Suggestions

None — comment density and style match the surrounding code.

### Positive Notes

- The skip is keyed on the frame's `owner`, the field CPython itself uses
  to tell entry frames apart, rather than on a `None` address or a
  failed read — it stays correct if the placeholder text or the
  qualname read path change.
- The two commits separate the correctness change (skip) from the budget
  change (bound on emitted symbols), so the second can be dropped alone
  if a verifier disagrees with the doubled iteration count.
- The block comment records the enum values per CPython version with the
  3.14 rename, which is exactly what the next person touching this will
  need.

### Mechanics

The x86_64 and aarch64 pystacks objects were compiled with the
repository's build recipe (`-g -O2 -target bpf -mcpu=v3 -fwrapv`, plus
the arch define and `vmlinux_<arch>.h`) using a distro clang 18: both
compile at the merge base and on the branch; the branch's
`pystacks_get_frame_data` shows the one-byte `bpf_probe_read_user` of the
owner and the `< 3` compare feeding the skip, and `walk_and_load_py_stack`
the `stack_len < stack_max_len` test around the call. The release object
is built by a different compiler version, so these are shape reads, not
byte matches with the release artifact.

The load and functional verdicts were observed on the branch's bytes
applied to the v1.17.7 source (the pystacks tree is identical at
v1.17.8) in an x86_64 KVM guest, kernel 6.12.93 with BTF: the release
object built by the repository's own nix package (embedded verbatim in
`bin/systing` and in the `bpf_load_shapes` test binary) loaded 34 of 34
shapes with no verifier refusal, the pystacks-on shape in 270.8 ms
(the unpatched control 431.3 ms on a different host) — so the doubled
visited bound of commit 2 passes the verifier on that kernel at no
visible cost; and a python 3.13.13 workload under
`--collect-pystacks` produced 0 samples carrying a `[Frame Error]` frame
(the unpatched control: every Python-bearing sample), with the `.py`
frame share (96.0 % vs 95.5 %), the hot function's samples (3,004 vs
2,998) and the `with_pystack` / `unknown` counters unchanged within run
noise, and the walker's per-event `stack_len` reading 2 where the control
read 3. The same object built for aarch64 (on top of the separate aarch64
TLS fix), loaded in an emulated aarch64 guest on the same kernel: 34 of
34 shapes, no refusal, the pystacks-on shape in 15.1 s under emulation
(17.3 s with the TLS fix alone), and the same workload with 0 samples
carrying a `[Frame Error]` frame (control: every Python-bearing sample),
`stack_len` 2 vs 3; the Python-bearing share moved −2.5 pt there and
+0.5 pt on x86_64 between runs of different length, read as run noise,
with one real reclassification named in the PR body (a mid-entry sample
whose chain holds only a shim frame now reads as no Python stack rather
than a lone placeholder). Not observed: the production kernels' own
verdict, which comes from loading the release-flag object with the
agent's runtime configuration before the tag ships.

PR #250, opened 2026-09-11
analyze: bound `sched aggregate` — Arrow-batched fold, time-chunked stream, event budget

josefbacik:
## Summary

`systing-analyze sched aggregate` folds one scheduler event stream — every slice start and end, runnable marker, new-task wakeup and migrate event, ordered by `(ts, kind)` — through a single-threaded state machine. Two things made that fold fail on busy hosts once a capture holds tens of millions of slices:

1. **The fetch loop, not the engine, dominated the wall.** Rows came through the driver's row API — `rows.next()` plus six typed `get()` calls per event — at roughly 180 ns per stream row. DuckDB's scan plus global sort was about 12 % of the wall at 24 M and 48 M stream rows; the client side was the other 88 %.
2. **The global sort does not fit and does not spill.** The engine materializes the whole stream to sort it. At 96 M stream rows the fold dies "Out of Memory" inside DuckDB within seconds under a 3–4 GiB memory limit, writes no spill file, and `preserve_insertion_order = false` changes nothing; forcing an external sort completes but the process peaks at about twice the limit in RSS. A host recording ~10 M context switches per second over a 10 s capture produces hundreds of millions of stream rows — more than any bounded process can sort in memory.

This PR bounds the fold in three parts, in order of necessity:

- **(A) Fetch as Arrow record batches.** `query_arrow` hands each batch over as typed column slices; the pass reads them directly. Same rows, same order, same state machine.
- **(B) Time-chunked streaming.** The window is split into K consecutive time chunks; each chunk's stream is sorted by the engine on its own and consumed by the *same* pass, so at most one chunk is materialized at a time. The result is identical to one pass over the whole stream: an ordered stream is the concatenation of its ordered time chunks, and a slice whose start and end fall in different chunks is two events either way. K comes from a cheap pre-count of the window's stream rows against `--chunk-rows` (default 16 M); K = 1 for every capture that fits, so those run exactly the one query they always did.
- **(C) An event budget with a truncation stamp.** When the window holds more stream rows than `--max-rows` (default 80 M, about 10 s of folding at the measured cost), the window is shortened from its **end** so it holds about the budget, and the pass runs over that shorter window. Nothing is sampled or extrapolated — every statistic is exact over a contiguous window — and the row says so: `meta.window_truncated = true`, `meta.window_ns` / `meta.window_end_ns` are the window actually folded, and `meta.slice_rows_capture` is the number of slices the requested window held. A shorter, honest window on the hosts where a full one cannot be folded beats no row at all.

Deliberately not in this PR: `TABLESAMPLE` / `LIMIT` (they break start/end pairing and sample the head of the ordered stream), forced external sorting (its RSS is outside the memory limit), and a longer deadline on the caller's side (the memory bound bites first).

## Output changes

Three new fields in `meta`, all additive; nothing existing changes meaning:

| field | type | meaning |
|---|---|---|
| `slice_rows_capture` | u64 | `sched_slice` rows (idle included) inside the requested window, counted before the fold |
| `window_truncated` | bool | the window was shortened from its end to fit the event budget |
| `stream_chunks` | u32 | time chunks the stream was fetched in (1 = one query over the whole window); diagnostic, never changes a result |

Consumers that pin the JSON shape (a golden file) will see the three new keys; consumers that already rate and filter by `window_ns` need nothing — a truncated row carries the window it was folded over.

The table output prints one line when the window was truncated and one when the stream was chunked.

## CLI

Two new options on `sched aggregate`: `--chunk-rows <N>` (0 = the whole window in one query) and `--max-rows <N>` (0 = no budget). The MCP `sched_aggregate` tool takes the defaults; its description mentions the truncation stamp.

## Test plan

Unit tests (all existing tests run through the new fetch path unchanged, and reproduce their fixed expectations):

- `chunked_fold_matches_the_single_query`: a 1 s two-CPU workload with wakeups, sleeps, preemptions, idle stretches and periodic migrations (1,600 slices, 1,200 markers, 4,400 stream rows) folded in 9 chunks and in 5 chunks (with a migrate table, exact placement) equals the single-query fold field for field.
- `event_budget_truncates_the_window_and_stamps_it`: a budget of half the stream rows cuts the window to exactly half, stamps it, keeps the full pre-count, and equals an explicit `--end-time` run over the same half; a budget the window fits leaves it whole; a budget far below the window keeps a stamped sliver; a budget no span can hold is an error naming the option.
- `chunked_stream_query_bounds_each_arm_on_its_own_timestamp`, `stream_chunks_cover_the_window_exactly`, `stream_count_query_shape`: the query builders (a slice's start arm is bounded at `ts`, its end arm at `ts + dur`, the last chunk closed inclusively; the chunk arithmetic; the pre-count's filters equal the stream's).

Bench (a synthetic 144-CPU / 29K-thread stream — WAKING + SWITCH pairs — written through the production parquet writer, then `parquet_to_duckdb`; DuckDB sized as a container with a 4 GiB analyze pool / 8 threads; release build; one 64-core host; "before" = the unmodified fold on the same data; wall = the run's own `aggregate_ms`, memory = the process's peak RSS):

| stream rows | before: one query | after: one query (`--chunk-rows 0 --max-rows 0`) | after: chunked |
|---|---|---|---|
| 24 M (8 M slices) | 4.79 s, 2.33 GB | 3.28 s, 2.31 GB | 3.02 s, 2.33 GB (3 chunks of 8 M) |
| 48 M (16 M slices) | 9.99 s, 4.60 GB | 6.34 s, 4.61 GB | 6.06 s, 2.48 GB (2 chunks of 32 M) |
| 96 M (32 M slices) | OOM inside DuckDB after 2 s, no spill | OOM after 2 s (`could not allocate block of size 256.0 KiB (4.0 GiB/4.0 GiB used)`) | 12.0 s, 3.48 GB (3 chunks of 32 M); 12.1 s, 2.03 GB (6 chunks of 16 M) |
| 96 M with `--max-rows 48000000` | — | — | 6.2 s, 1.58 GB: the window halved from its end (2.01 s → 1.005 s) and stamped, `slice_rows_capture` = 32.0 M kept |

Every output pair that should be equal is equal field for field (`aggregate_ms` and the three new fields aside): before vs after at 24 M rows, one query vs chunked at 24 M and 48 M, 3 chunks vs 6 chunks at 96 M. The fold now costs about 125 ns per stream row end to end (it was about 200); the remainder is the pass itself (per-event hash-map lookups of thread and CPU state), a possible follow-up, not this PR.

`cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `cargo test --lib analyze::sched_aggregate` (22 tests) pass.

### DIFF
--- a/src/analyze/mod.rs
+++ b/src/analyze/mod.rs
@@ -29,7 +29,8 @@ pub use network_socket_pairs::{
 };
 pub use sched_aggregate::{
     Dist, Imbalance, PerCpu, RqDist, SchedAggregate, SchedAggregateMeta, SchedAggregateParams,
-    SwitchStats, TailContributor, HIST_BASE_LOG2, HIST_OCTAVES,
+    SwitchStats, TailContributor, DEFAULT_CHUNK_ROWS, DEFAULT_MAX_ROWS, HIST_BASE_LOG2,
+    HIST_OCTAVES,
 };
 pub use sched_stats::{
     EndStateCount, ProcessSchedStats, SchedStatsParams, SchedStatsResult, SchedSummary,
--- a/src/analyze/sched_aggregate.rs
+++ b/src/analyze/sched_aggregate.rs
@@ -81,8 +81,41 @@
 //! are exact. The `hist_log2` arrays are the
 //! same histograms folded to one count per octave so rows from different
 //! captures and hosts can be merged.
+//!
+//! # Bounded folding
+//!
+//! The pass consumes one event stream ordered by `(ts, kind)`. The engine
+//! materializes that stream to sort it, and the sort does not spill: on a
+//! host that records tens of millions of context switches per second a
+//! 10-second capture holds hundreds of millions of stream rows, more than a
+//! bounded process can hold. Two guards keep the fold inside a budget, both
+//! set from [`SchedAggregateParams`] and both reported in the metadata:
+//!
+//! - **Chunking** (`chunk_rows`): the stream is fetched in consecutive time
+//!   chunks of the window, each sorted by the engine on its own and consumed
+//!   by the same pass, so at most one chunk is materialized at a time. The
+//!   result is identical to one pass over the whole stream — an ordered
+//!   stream is the concatenation of its ordered time chunks, and a slice
+//!   whose start and end fall in different chunks is two events in the
+//!   stream either way. `meta.stream_chunks` says how many chunks were used;
+//!   1 means the fold ran as a single query.
+//! - **The event budget** (`max_rows`): when the requested window holds more
+//!   stream rows than the budget, the window is shortened from its END so
+//!   that it holds about the budget, and the pass runs over that shorter
+//!   window. Every statistic stays exactly defined over a contiguous,
+//!   un-sampled window — nothing is sampled or extrapolated — and the row
+//!   says so: `meta.window_truncated` is true, `meta.window_ns` is the
+//!   window actually folded, and `meta.slice_rows_capture` is the number of
+//!   slices the requested window held. A shorter window on the hosts where a
+//!   full one cannot be folded beats no row at all.
+//!
+//! The stream rows of a window are counted before the fold (one aggregate
+//! query per source table) to size both guards.
 
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
+use duckdb::arrow::array::{Array, AsArray, PrimitiveArray};
+use duckdb::arrow::datatypes::{Int32Type, Int64Type};
+use duckdb::arrow::record_batch::RecordBatch;
 use serde::Serialize;
 use std::collections::HashMap;
 
@@ -100,6 +133,18 @@ const HIST_SUB: usize = 1 << HIST_SUB_BITS;
 /// Largest runqueue length tracked exactly by the time-weighted histogram;
 /// longer queues are clamped into the last bucket.
 const RQ_HIST_MAX: usize = 4096;
+/// Default stream rows per fetched chunk (see the module docs, "Bounded
+/// folding"): the engine sorts one chunk at a time, so this bounds the
+/// memory the sort holds — measured, a 16 M-row chunk keeps the process
+/// near 2 GB where one query over 96 M rows exhausts a 4 GiB memory limit;
+/// chunking costs no wall time, so the default errs small.
+pub const DEFAULT_CHUNK_ROWS: u64 = 16_000_000;
+/// Default event budget for one fold (stream rows in the window): above it
+/// the window is shortened from its end so the pass stays inside the time
+/// a bounded process allows it — about 10 s at the measured 125 ns per
+/// stream row. A stream row is a slice start, a slice end, a runnable
+/// marker, a new-task wakeup or a migrate event.
+pub const DEFAULT_MAX_ROWS: u64 = 80_000_000;
 
 /// Parameters for [`AnalyzeDb::sched_aggregate`].
 ///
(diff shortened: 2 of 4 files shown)


josefbacik:
# rust-bpf-code-reviewer verdict — systing branch `sched-aggregate-bounded`

Reviewed against the repository's reviewer definition (`.claude/agents/rust-bpf-code-reviewer.md`, blob `3739c8b3…`, unchanged since v1.13.x), applied by the author to the committed diff `6b106c83..fb40bdb0` on v1.17.8 — four files, 586 insertions / 32 deletions: `src/analyze/sched_aggregate.rs` (587 lines changed, about half of them tests), `src/analyze/mod.rs` (3), `src/bin/systing_analyze.rs` (25), `src/mcp.rs` (3). No BPF file moves; no schema, `SCHEMA_VERSION`, `SCHEMA_CHANGES.md` or `Cargo.toml` version edits.

### Summary

A contained change to one analysis pass: the scheduler event stream is fetched as Arrow record batches instead of through the driver's row API, the window is folded in consecutive time chunks so the engine sorts one chunk at a time, and an event budget shortens an over-large window from its end with an explicit stamp in the output. The state machine (`Pass`) is untouched; the equivalence of the chunked fold to the single-query fold is argued in the module docs and pinned by tests at three chunk counts plus a bench equality at 24 M / 48 M / 96 M rows. The code is direct, the docs say what the guarantees are, and the measured effect is the one claimed (96 M-row captures that died OOM now complete in 12 s under a 4 GiB limit).

### Critical Issues

None found.

- Schema & versioning: no table definition, `create_schema()`, `SCHEMA_VERSION` or `SCHEMA_CHANGES.md` touch; the three new fields are in the aggregate's JSON output (`SchedAggregateMeta`), not a DuckDB table. `Cargo.toml` / `Cargo.lock` untouched. No `[bump minor]` needed.
- Unsafe: none added.
- Correctness of the chunk predicate: each arm is bounded on its own event timestamp (start at `ss.ts`, end at `ss.ts + ss.dur`, the others at `ts`), with the outer window filter kept on every arm, and the last chunk closed inclusively — the pinned tests `chunked_stream_query_bounds_each_arm_on_its_own_timestamp` and `stream_chunks_cover_the_window_exactly` read this back. A slice straddling a chunk boundary is a START in one chunk and an END in a later one, exactly as the single query orders them.
- Types across the driver boundary: every column of the stream query is cast explicitly, and `stream_column_i64` / `stream_column_i32` fail with the observed type instead of panicking (`as_primitive_opt`, not `as_primitive`). A NULL in a non-nullable column fails the fold, which is what the row API did (`r.get::<_, i64>` on NULL).

### Improvements

1. `sched_aggregate()` — the bail text for a budget no span of the window fits says "raise `--max-rows`", a CLI spelling inside a library function; the MCP path takes the defaults and cannot raise it. Naming the parameter (`max_rows`) instead would be cleaner. Cosmetic; deferred so the tested bytes ship unchanged — the reviewer may ask for it on the PR.
2. The window-truncation branch re-counts the shortened window to size the chunks exactly (one extra aggregate query, ~1 s at 96 M rows). Acceptable — it runs only on captures that would otherwise fail — and simpler than scaling the first count, which would mis-size chunks on a bursty stream.

### Suggestions

- The measured cost is now ≈125 ns per stream row end to end; the remainder is the pass's per-event `HashMap<i64, ThreadState>` lookups (SipHash). A faster hasher for the thread map is the next lever if one is wanted; out of scope here (the PR body says so).
- `StreamCounts::stream_rows()` weights a slice as two rows (start + end), which matches the query; a one-line comment already says so in the struct docs.
- Test helpers use type aliases (`SliceRow` / `MarkerRow` / `ThreadRow`) to satisfy `clippy::type_complexity`; fine.

### Positive Notes

- The module docs' "Bounded folding" section states the two guarantees a consumer needs (identical results under chunking; exact-over-a-shorter-window under the budget, with the stamp) before any implementation detail.
- Chunking is a no-op by construction when K = 1: the query text is the one the code always ran (no chunk predicate is emitted), so unaffected captures cannot regress.
- The truncation test proves the truncated run equals an explicit `--end-time` run over the same half-window, field for field, rather than asserting a few figures.
- `chunk_rows = 0` / `max_rows = 0` opt out of each guard independently, which is what a bench or a reviewer needs to reproduce the old behaviour.

Gates on these bytes: `rustfmt --check` on the four files clean; `cargo clippy --all-targets --no-default-features -- -D warnings` clean; `cargo test --lib analyze::sched_aggregate` 22 passed / 0 failed (03:23Z 2026-09-07).

Verdict: **approve for push** (READY on open) — no critical issue, one cosmetic improvement left to the PR review.

PR #251, opened 2026-09-12
README: the network-packets tier is a sample on large hosts — document --packet-sample-rate

josefbacik:
Docs only — two places in `README.md`, nothing else.

1. The `network-packets` recorder bullet now says that on a large host the tier loses most of its events to the userspace consumer, so the packet tables are a sample; what `--packet-sample-rate N` keeps (1 in N packets of the data-path event types, every stage of a kept packet kept together so per-stage latencies stay pairable; packet and byte counts scale by N; the diagnostic events — zero-window probes, RTO timeouts, drops, state changes — are never sampled); where the rate ran is recorded (`sysinfo.network_packet_sample_rate`); and that the exit summary prints the events missed beside the events recorded. The memory tier's sampling knobs were already listed on its own bullet; the network one was not.
2. The Network Traffic Recording paragraph said the packets tier's event volume is "bounded by traffic rather than by anything you control". It now says how to bound it: keep a packets capture short or run it with `--packet-sample-rate N`, with the exit summary's missed count as the check.

Follows #242 (the knob, the `sysinfo` column and the exit-time line; v1.18.0). No code, no schema, no behaviour change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ This will display all available recorders and their default states:
 - `cpu-stacks` - CPU perf stack traces (on by default)
 - `network` - Network connection state tracking
 - `network-syscalls` - Network syscall-level tracing (send/recv bytes, retransmits, drops, stalls) without per-packet probes
-- `network-packets` - Network packet-level tracing (sendmsg, recvmsg, qdisc, drops)
+- `network-packets` - Network packet-level tracing (sendmsg, recvmsg, qdisc, drops). On a large host this tier loses most of its events to the userspace consumer, so the packet tables are a sample: `--packet-sample-rate N` keeps 1 in N packets of the data-path event types (every stage of a kept packet is kept together, so per-stage latencies stay pairable; packet and byte counts scale by N; the diagnostic events — zero-window probes, RTO timeouts, drops, state changes — are never sampled), the rate ran is recorded as `sysinfo.network_packet_sample_rate`, and the exit summary prints the events missed beside the events recorded.
 - `memory` - Memory usage tracking (RSS, mmap/munmap/brk, page faults; host-wide THP/compaction counters into `memory_vmstat`). `--memory-vfio` adds VFIO DMA regions and the IOMMU map/unmap run-size histogram (how fragmented the memory behind device mappings is); `--memory-thp-sample-rate N` adds sampled THP-split events with stacks. Both legs turn themselves off (named in `sysinfo`) on hosts without the symbols.
 - `memory-alloc` - Heap allocator uprobes (malloc/calloc/realloc/free) with stacks
 - `markers` - Userspace marker events (faccessat2 with mode=-975)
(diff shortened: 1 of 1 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `README: the network-packets tier is a sample on large hosts — document --packet-sample-rate` (head `217425b8e43c9103ab23b7099bd0b76770ff11fe`)

Mechanics: the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session by the author over the full diff against `v1.17.8` (`6b106c8`; one file, `README.md` +4/−2), in the definition's five passes. The Rust, BPF and recorder-consistency criteria have nothing to read here (no source file changes); the schema-and-versioning pass and the integration pass are the substance, and the integration pass read every claim the new text makes against the released code it documents (`v1.18.0`, `decaea9`, which added `--packet-sample-rate` and did not touch `README.md`): `packet_sample_keep()` and its comment in `src/bpf/systing_system.bpf.c`, the `--packet-sample-rate` argument's help text in `src/main.rs`, the `network_packet_sample_rate` column of `sysinfo` (`SCHEMA_VERSION` 21) and the exit summary's `packet events: … recorded, … missed` line in `src/systing_core.rs`.

### Summary

A documentation-only change that says, in the two places a reader meets the `network-packets` tier, what the code already does: on a large host the tier is a sample because the userspace consumer falls behind, `--packet-sample-rate N` keeps 1 in N packets with every stage of a kept packet kept together, the rate ran is recorded in `sysinfo`, and the exit summary reports the missed count. Every claim in the added text matches the released implementation; the wording follows the `memory` bullet's precedent for per-recorder knobs. No code, schema, version or behaviour change.

### Critical Issues

None. `Cargo.toml` and `Cargo.lock` are untouched; no schema file changes, so no `SCHEMA_CHANGES.md` entry or `[bump minor]` token is owed; the commit carries exactly one `Co-authored-by` trailer and no other trailer.

### Improvements

None owed for this change.

### Suggestions

- The example capture under "Network Traffic Recording" (`--only-recorder network-packets --duration 60`) predates the knob and shows a 60-second packets capture with no rate; a later docs change could show the sampled form beside it, since the new paragraph tells the reader to keep such a capture short or sample it. Not part of this change.
- The exit summary prints the sample line only when the missed count is non-zero; the bullet's "prints the events missed beside the events recorded" reads naturally as that case and needs no change.

### Positive Notes

- The bullet's description of the sampling decision (per packet, every stage of a kept packet kept together, per-stage latencies pairable, counts and bytes scaling by N, the diagnostic event types never sampled) is the same statement the BPF comment and the argument's help text make, so the three readers a user has agree.
- Naming `sysinfo.network_packet_sample_rate` and the exit summary's missed count gives the reader the two checks that tell a sampled capture from a whole one, which is the point of the change.

PR #252, opened 2026-09-12
ci: load every BPF shape on pinned kernels in a guest per kernel

josefbacik:
## Summary

Add a CI workflow, `bpf-load-shapes`, that loads every BPF program of the release-compiler object at every shipping configuration shape on the kernels the gate is expected to pass on — in a guest per kernel, on every push and pull request to `main`, before any tag.

`every_shape_loads` (`tests/bpf_load_shapes.rs`, from #225) is already the load-time positive control for the class of regression where a program that loads at the defaults is rejected once the memory recorder's `.rodata` configuration enables a pruned branch. Until now it ran only by hand: `ci.yml` builds with `SYSTING_BPF_CLANG: clang-21` but runs the unit tests alone on the runner's own kernel, so the gate never ran on a pull request and never on the kernels that matter.

## What the workflow does

**Job `build`** — the same toolchain steps as `ci.yml`'s test job (clang 21 as the BPF compiler, the pinned pre-built libduckdb), then `cargo test --test bpf_load_shapes --no-run --no-default-features` and the binary located from cargo's JSON output exactly as `scripts/run-integration-tests.sh` does; the binary and `libduckdb.so` go up as one artifact.

**Job `load-shapes`** — a matrix, one guest per kernel, `fail-fast: false`:

| row | kernel | image |
|---|---|---|
| `vmtest-6.12` | the repository's vmtest kernel (vanilla v6.12, BPF selftests config) | `bzImage-v6.12-bpf` from the `vmtest-kernels` release |
| `cos-125-6.12.85` | Container-Optimized OS milestone 125, build 19216.395.138, kernel 6.12.85+ | the build's published ELF `vmlinux` |
| `cos-129-6.12.90` | Container-Optimized OS milestone 129, build 19506.224.80, kernel 6.12.90+ | the build's published ELF `vmlinux` |

Each row downloads its image (cached by digest), verifies the sha256 pinned in the matrix — a kernel image is executed, so it is pinned like any other dependency — boots it with virtme-ng 1.41 (KVM when the runner exposes `/dev/kvm`, TCG otherwise), and runs `every_shape_loads` through the new guest script `scripts/ci/vmtest-load-shapes.sh`, which writes the test's output and a `VNG-TEST-EXIT:<rc>` line into the shared workspace. The host step prints that file into the job log and fails unless the line reads `VNG-TEST-EXIT:0`.

The verdict travels in the file on purpose: the Container-Optimized OS kernels have no virtio console, so vng cannot read the guest's exit code there and returns 255 whether the test passed or failed. A COS `vmlinux` boots directly under qemu (`CONFIG_PVH=y`) with 9p built in, which is what makes those rows possible without a rebuilt image.

## Image sources (for a reviewer to verify)

- `https://github.com/josefbacik/systing/releases/download/vmtest-kernels/bzImage-v6.12-bpf` — sha256 `324f2d6cad6763886afc0392a205b27f71176ae5f9357a06219e773177089244` is the release asset's own digest.
- `https://storage.googleapis.com/cos-tools/19216.395.138/gke-amd64-gcp/vmlinux` — sha256 `6c9460560118608c7ca90d7da324fd100d40fb18ecf03c06fa7902cf2b57011a` (376,046,128 bytes); the build → kernel mapping is on the milestone 125 release-notes page.
- `https://storage.googleapis.com/cos-tools/19506.224.80/gke-amd64-gcp/vmlinux` — sha256 `0c39e1964e6dfe2871227d30c4141a01a05b1e7fd1e184e87608b461e64ef69e` (378,632,832 bytes); milestone 129.

## What it does and does not prove

- It proves the object built by CI with the release BPF compiler loads on those kernels at every shape in `shape_table()`, and prints the rejecting verifier log in full when one does not. The 6.12.90 image booted and ran a script on a TCG guest before this change was written; the load-shape run itself on these images is what this workflow's first run will show.
- It does not prove the nix-built release object: that is a separate guard elsewhere. The two together cover both artifacts.
- The coverage read `every_instruction_is_verified_by_some_shape` is deliberately not run here (its level-2 verifier log is unbounded under TCG); the gate alone runs.

## Expectations for the first run (inferred until it runs)

- GitHub's x64 Linux runners expose `/dev/kvm`; with KVM a 6.12 guest boots in seconds and the gate takes tens of seconds per kernel. Under TCG the gate takes roughly three to six minutes per kernel; the per-row timeout is 15 minutes.
- Each COS image is ≈376–379 MB, cached by digest after the first run; the vmtest image is 13 MB.
- `pipx install virtme-ng==1.41` builds virtme-ng's init helper with the runner's Rust toolchain.

## Follow-ups

- A Container-Optimized OS 6.6 row (milestone 121's builds) the same way.
- Further kernel rows are added by the same pattern: an image pinned by its digest, one matrix row.
- Making the check required is a branch-protection setting, not part of this change.

### DIFF
--- a/.github/workflows/bpf-load-shapes.yml
+++ b/.github/workflows/bpf-load-shapes.yml
@@ -1,0 +1,158 @@
+name: bpf-load-shapes
+
+# Load every BPF program of the release-compiler object, at every configuration
+# shape that ships, on the kernels the object runs on — in a guest per kernel,
+# before any tag. `every_shape_loads` (tests/bpf_load_shapes.rs) is the gate;
+# the object is built here with the same BPF compiler the release build uses.
+#
+# Kernel images are executed, so each one is pinned by its sha256 in the matrix
+# below and verified after the download; the download is cached by that digest.
+# The verdict travels in a file the guest writes into the shared workspace, not in
+# the VM tool's exit code: a Container-Optimized OS kernel has no virtio console,
+# so vng cannot read the guest's exit code there and returns 255 either way.
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build the load-shape gate (release BPF compiler)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y linux-tools-common libelf-dev linux-libc-dev clang libbpf-dev make pkg-config jq libjemalloc2
+      - name: Install clang 21 (the release build's BPF compiler)
+        run: |
+          wget -q https://apt.llvm.org/llvm.sh -O /tmp/llvm.sh
+          sudo bash /tmp/llvm.sh 21
+          clang-21 --version
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.9
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: "true"
+      - name: Download the pre-built DuckDB library the crate pins
+        run: |
+          LIBDUCKDB_VERSION="$(scripts/libduckdb-version.sh)"
+          echo "libduckdb v${LIBDUCKDB_VERSION} (the DuckDB version Cargo.lock pins)"
+          wget -nv "https://github.com/duckdb/duckdb/releases/download/v${LIBDUCKDB_VERSION}/libduckdb-linux-amd64.zip" -O /tmp/libduckdb.zip
+          unzip /tmp/libduckdb.zip -d ${{ github.workspace }}/libduckdb
+      - name: Build the bpf_load_shapes test binary
+        env:
+          SCCACHE_GHA_ENABLED: "true"
+          RUSTC_WRAPPER: "sccache"
+          DUCKDB_LIB_DIR: "${{ github.workspace }}/libduckdb"
+          DUCKDB_INCLUDE_DIR: "${{ github.workspace }}/libduckdb"
+          LD_LIBRARY_PATH: "${{ github.workspace }}/libduckdb"
+          SYSTING_BPF_CLANG: "clang-21"
+        run: |
+          cargo test --test bpf_load_shapes --no-run --no-default-features --message-format=json > /tmp/build.json
+          BIN="$(jq -r 'select(.reason == "compiler-artifact") | select(.target.kind[] == "test") | select(.target.name == "bpf_load_shapes") | .executable' /tmp/build.json | tail -1)"
+          test -x "$BIN"
+          mkdir -p gate
+          cp "$BIN" gate/bpf_load_shapes
+          cp libduckdb/libduckdb.so gate/
+          ls -l gate
+      - uses: actions/upload-artifact@v4
+        with:
+          name: load-shape-gate
+          path: gate/
+          retention-days: 1
+          if-no-files-found: error
+
+  load-shapes:
+    name: Load every shape on ${{ matrix.kernel.id }}
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        kernel:
+          # The repository's own vmtest kernel: vanilla v6.12 with the BPF
+          # selftests config (kernel/build.sh, the `vmtest-kernels` release).
+          - id: vmtest-6.12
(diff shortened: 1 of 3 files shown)


josefbacik:

Executed in-session by the author over the full diff of the branch against its base `6b106c8` (v1.17.8): `.github/workflows/bpf-load-shapes.yml` (+158, new), `scripts/ci/vmtest-load-shapes.sh` (+27, new), `README.md` (+6). No Rust, no BPF, no schema: passes 2–4 of the definition have nothing to judge; passes 1 (structure) and 5 (integration) and the cleanliness criteria apply.

### Summary

A CI workflow that runs the existing load-shape gate `every_shape_loads` on three pinned kernels in virtme-ng guests, with the object built by the release BPF compiler. The structure is two jobs (one build, one matrix of guests), the verdict is a guest-written file read by the host step, and every executed image is pinned by digest and verified after download. Nothing blocks; the first CI run is the validation the change itself cannot get locally.

### Critical Issues

None.

- Supply chain: the three kernel images are executed, and each is pinned by sha256 in the matrix and checked with `sha256sum -c` before the boot; the cache key carries the digest, so a cache entry cannot serve a different image. The actions are referenced the way `ci.yml` references them (checkout by commit sha, the others by version tag) — consistent with the repository, not tighter.
- Secrets: none used; `permissions: contents: read`.
- The workflow cannot silently pass: the host step fails unless the guest-written file carries `VNG-TEST-EXIT:0`, so a guest that never booted, a script that never ran, or a test that failed all fail the row; `set +e` around the vng call exists only so vng's meaningless exit status on a console-less kernel does not pre-empt reading the file, and `set -e` is restored before the grep.

### Improvements

- The `build` job repeats the toolchain steps of `ci.yml`'s `test` job (apt list, clang 21, sccache, rust-cache, libduckdb). A composite action or a reusable workflow would remove the duplication; left as is so this change stays a single readable file and does not touch `ci.yml`. Worth doing when a third job needs the same steps.
- `pipx install virtme-ng==1.41` builds virtme-ng's init helper on the runner; if the first run shows that step slow or flaky, a cached wheel (or `--no-virtme-ng-init`) is the lever.

### Suggestions

- The guest script prints `hypervisor flag set` from `/proc/cpuinfo` — under qemu that flag reads set for KVM and TCG alike; it tells the reader a VM ran, not which accelerator. The host step's `accelerator:` line is the authoritative one. A one-word rename of the guest line would avoid the double reading.
- When further kernel rows are added by the same pattern, the download step's `--max-time 900` and the row timeout should be re-read against larger or module-carrying images.

### Positive Notes

- The file-verdict design is stated where it matters (the workflow header comment, the guest script header, the README), with the reason: a console-less guest kernel makes the VM tool's exit status meaningless, so the exit code is written into the shared workspace instead.
- The binary is located from cargo's JSON output with the same filter `scripts/run-integration-tests.sh` uses, so the two paths cannot drift.
- Comments explain why, name no line numbers and no external change numbers.
- `fail-fast: false` keeps every kernel's verdict visible when one fails, which is the point of a matrix over kernels.

### Verified locally
- The workflow parses as YAML (two jobs, three matrix rows); `bash -n` on the guest script; the three pinned digests re-computed from the downloaded images and the release asset's own digest and found to match the matrix and the body (3 of 3 each, no stray digests).
- Not verified locally (no GitHub Actions runner here): the workflow's execution; the first CI run on the pull request is that check.

josefbacik:
## Code review addendum for `605e7422` (rust-bpf-code-reviewer)

Executed in-session by the author over the diff of `605e7422` against `87feeeb6`: `.github/workflows/bpf-load-shapes.yml` (+7/−2). No Rust, no BPF, no schema: passes 2–4 of the definition have nothing to judge; pass 5 (integration) and the cleanliness criteria apply.

### Summary

The first run of the workflow on this branch failed in every matrix job at the "Install qemu and virtme-ng" step: `pipx install virtme-ng==1.41` succeeded and the next line ran `"$HOME/.local/bin/vng" --version`, which exited 127 — pipx places its apps in `PIPX_BIN_DIR`, which the runner image points elsewhere. The step now asks pipx for that directory (`pipx environment --value PIPX_BIN_DIR`), prints it, appends it to `GITHUB_PATH` and runs `vng --version` from it. The boot step already calls `vng` by name from `PATH`, so nothing else moves.

### Critical Issues

None.

### Improvements

- `pipx environment --value` exists from pipx 1.1.0; the step installs the distribution's pipx (1.4.3 on the current image, per the failed run's log) and the image's own pipx is newer still, so the call is available either way. If a future image ships an older pipx the step fails at that line with a clear message rather than at a guessed path — acceptable.

### Suggestions

- None beyond the change: the printed `pipx apps: …` line is the record of where the tool landed, which is what the first run lacked.

### Positive Notes

- The cure asks the tool instead of assuming a second path; a wrong assumption is what failed.
- The `--version` probe stays as the step's own gate, so a missing binary fails the step at install time, before any kernel download.

PR #253, opened 2026-09-12
test: allow the TIME_WAIT programs unselected only where the kernel lacks a hooked function

josefbacik:
## Summary

The load-shape gate `every_shape_loads` requires every program in the object to be selected by some shape, or documented in `SELECTED_ELSEWHERE`. On a 6.6-series kernel it failed naming six programs — the network recorder's TIME_WAIT leg (`tcp_time_wait`, `inet_twsk_hashdance_schedule`, `inet_twsk_deschedule_put`, each as a kprobe and as its fentry twin) — although every selected program loaded: the recorder switches that leg off on a kernel that lacks one of the three hooked functions (`inet_twsk_hashdance_schedule` exists from 6.11, b334b924c9b7), so no shape selects the six there.

This change teaches the gate the recorder's own rule instead of a blanket exemption.

## What changes

- `src/systing_core.rs`: `NETWORK_TW_SYMBOLS` and `kallsyms_has_funcs` become `pub` (a `const` slice and a read-only `/proc/kallsyms` reader; no runtime behaviour changes).
- `tests/bpf_load_shapes.rs`: a new `SELECTED_WHEN_KERNEL_HAS` list names the six programs with the symbols the leg needs. A program in it that no shape selected is allowed — reported under `[selection]` with the missing function named — only when the running kernel lacks one of those functions. On a kernel that has all three, an unselected TIME_WAIT program remains a finding, as before.

A static `SELECTED_ELSEWHERE` row would have reported the six on every kernel and let a regression that dropped the leg where it must load pass silently; the kernel-conditional form keeps that regression visible.

## Verified

- In a guest on a 6.6.97 kernel: 34/34 shapes loaded, 0 verifier rejections, the six programs reported `[selection] … (this kernel lacks inet_twsk_hashdance_schedule, so the recorder keeps the leg off)`, `test result: ok`.
- In a guest on a 6.12.85 kernel: 34/34 shapes loaded, every program selected, the new allowance never consulted, `test result: ok`.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and the lib tests (590 passed) clean.

Before this change the same 6.6.97 guest run loaded 34/34 shapes and failed only on the six unselected programs.

## Notes

- The reader's failure mode is shared with the recorder by construction: an unreadable `/proc/kallsyms` reads as "no symbol present" on both sides (the leg off; the six allowed unselected).
- The change is test-only in effect; it needs no schema or version bump.

### DIFF
--- a/src/systing_core.rs
+++ b/src/systing_core.rs
@@ -342,7 +342,7 @@ const NETWORK_TW_KPROBE_PROGRAMS: &[&str] = &[
 /// (b334b924c9b7) — a 6.6-series kernel lacks it, and before this gate a
 /// network capture on such a node failed its every attach (the whole
 /// capture, CPU lane included) rather than running without the leg.
-const NETWORK_TW_SYMBOLS: &[&str] = &[
+pub const NETWORK_TW_SYMBOLS: &[&str] = &[
     "tcp_time_wait",
     "inet_twsk_hashdance_schedule",
     "inet_twsk_deschedule_put",
@@ -1145,7 +1145,7 @@ fn probe_memory_kernel_legs(opts: &Config, kernel: &KernelSymbols) -> MemoryKern
 /// Which of `names` /proc/kallsyms lists as text symbols (`t`/`T`). An
 /// unreadable kallsyms reads as "none present", which turns the legs off
 /// rather than failing the capture.
-fn kallsyms_has_funcs(names: &[&str]) -> HashSet<String> {
+pub fn kallsyms_has_funcs(names: &[&str]) -> HashSet<String> {
     let mut found = HashSet::new();
     let Ok(contents) = fs::read_to_string("/proc/kallsyms") else {
         return found;
--- a/tests/bpf_load_shapes.rs
+++ b/tests/bpf_load_shapes.rs
@@ -30,7 +30,7 @@
 use std::collections::{BTreeMap, BTreeSet};
 
 use systing::bpf_load_shapes::{coverage_gaps, ranges, shape_table, LoadReport};
-use systing::systing_core::bpf_load_probe;
+use systing::systing_core::{bpf_load_probe, kallsyms_has_funcs, NETWORK_TW_SYMBOLS};
 
 /// Instructions known to be unreachable at every shipping shape, with the
 /// reason. Add a row only with the reason; the entry is `(program, ranges)`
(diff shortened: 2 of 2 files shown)


josefbacik:

Executed in-session by the author over the full diff of the branch against its base `6b106c8` (v1.17.8): `src/systing_core.rs` (+2/−2: two visibility changes), `tests/bpf_load_shapes.rs` (+40/−4). No BPF change, no schema change, no runtime behaviour change: passes 2 and 4 of the definition have nothing to judge; pass 1 (structure), pass 3 (Rust) and pass 5 (integration/tests) apply.

### Summary

The load-shape gate's "every program is selected by some shape" check gains a kernel-conditional allowance for the network recorder's six TIME_WAIT programs, mirroring the recorder's own symbol gate instead of copying it: `NETWORK_TW_SYMBOLS` and `kallsyms_has_funcs` become `pub`, and the six are allowed unselected only when the running kernel lacks one of the leg's three hooked functions, with the missing one named in the `[selection]` line. On a kernel that has all three, an unselected TIME_WAIT program stays a finding. Nothing blocks.

### Critical Issues

None.

- The two `pub` promotions expose a `const` slice and a read-only `/proc/kallsyms` reader; neither carries state or a side effect, and neither changes the crate's runtime paths. The test crate is the only new caller.
- The allowance cannot mask the regression a static row would mask: the branch is taken only when `kallsyms_has_funcs` reports a symbol absent, and on a kernel with all three present the program falls through to `never_selected` exactly as before.
- The failure mode of the reader is inherited and safe here: an unreadable `/proc/kallsyms` reads as "none present" in the recorder (the leg turns off) and in the test (the six are allowed unselected, with all three symbols named as missing) — the same reading on both sides, which is the point of sharing the function.

### Improvements

- `kallsyms_has_funcs` is called once per program in the list (six reads of `/proc/kallsyms` on a kernel where the leg is off). Hoisting one read of the union of symbols above the loop would cost less; at six programs and a file of a few MB it is a fraction of a second beside a 250-s test, so left as is for readability.
- The `[selection]` message names the missing symbol(s) but not the leg; a reader of the test log on an unfamiliar kernel would be helped by the leg's name ("the network recorder's TIME_WAIT leg"). A one-string change.

### Suggestions

- If a second kernel-gated leg appears (a future fentry set), `SELECTED_WHEN_KERNEL_HAS` already takes any symbol list per program; the doc comment could say so.

### Positive Notes

- The allowance's condition is the recorder's own gate (`NETWORK_TW_SYMBOLS` through `kallsyms_has_funcs`), so the test cannot drift from the recorder's selection rule.
- The comment states the boundary with its commit (`inet_twsk_hashdance_schedule` from 6.11, b334b924c9b7) and why a static row would be wrong.
- The control flow keeps `SELECTED_ELSEWHERE` first and unchanged; every path that reaches `never_selected` is the previous behaviour.
- Verified in guests on two kernels: 6.6.97 (the six reported under `[selection]` naming the missing function; 34/34 shapes loaded; the gate passes) and 6.12.85 (every program selected, the allowance never consulted; 34/34; passes) — the positive case and the negative control of the change.

### Verified locally
- `cargo fmt --all -- --check` clean; `cargo clippy --all-targets -- -D warnings` clean; `cargo test --lib` 590 passed, 0 failed.
- The two guest runs above, read from the guest-written files.

PR #254, opened 2026-09-12
symbolize: never hand a gVisor pool-memfd address to the process source

josefbacik:
## Summary

On hosts running gVisor sandboxes, `symbolize_user_addr` handed every
user address to blazesym's process source first — pool-memfd addresses
included. For an address inside one of the Sentry's memory pools
(`runsc-memory` islands inside guest text, `systrap-memory` sysmsg
pages) that call resolves the mapping through `/proc/<pid>/maps`, opens
the mapping's `map_files` link — the pool memfd itself, for
`runsc-memory` the sandbox's whole MemoryFile — and parses that file
from byte 0 as if it were the ELF the address belongs to. A pool is
never an ELF image, so the call costs time and symbol-cache memory and
can never name the address; the bridge (`bridge_for`) and the guest's
own maps, which run right after it, are where such addresses resolve.

This PR routes pool-backed addresses past the process source.

## The cost, as observed

On a gVisor-based Kubernetes cluster, continuous captures of 1–5 K stack
rows that sampled sandbox processes spent 60–192 s in `symbolize and
write parquet` while the fast captures on the same hosts took 0.5–16 s,
with the tracer's anon RSS at the symbolizer's memory-budget valve
(`memory.max / 2`). The slow captures were exactly the ones whose samples
landed in `memfd:runsc` processes, and their time tracked the number of
pool addresses that ended as `[gvisor:*]` labels — roughly 4–13 s per
such address — i.e. a per-address cost sitting on the addresses that go
through `handle_elf_addr` on the island's link before `bridge_for` and
come out unresolved.

## The change

- `ProcessMaps::is_pool_backed(addr)` (new): true iff the maps entry
  containing `addr` is a memfd named `runsc-memory` or `systrap-memory`
  — the `GVISOR_MEMFDS` set the sandbox classification (`is_gvisor`)
  already keys on. The test is the memfd's NAME, never `memfd:` as a
  class: the sandbox binary runsc re-execs from a memfd named `runsc`
  is a real ELF and keeps the ordinary path, and so does any other
  runtime's exec memfd (a JIT's code file).
- `symbolize_user_addr`: for a pool-backed address the
  `symbolize_single(Process, AbsAddr)` call is skipped; the address goes
  straight to the bridge, then the guest view, then the host-side label
  — the path it took after the failed call before.

## What does not change

No frame string changes for any class: bridged islands still render as
the original binary's symbol, guest-view misses as `[gvisor:*]` /
`[jit:*]` / `<module>:unknown`, file-backed and runsc-binary frames
exactly as today. Only the time and memory of the pool parse go away.

## Tests

- `sandbox_maps::tests::test_is_pool_backed_names_the_pools_only`: a
  maps fixture with the usertrap trampolines below the image, a
  patched-syscall island inside guest text, a writable pool mapping, a
  `systrap-memory` page, the runsc binary's own `/memfd:runsc
  (deleted)` mapping, file-backed text, an anon exec page and `[stack]`
  — pool-backed only for the four pool addresses (an island's last byte
  included); plus a non-gVisor process with a `/memfd:jit-code` exec
  mapping, untouched.
- `stack_recorder`'s existing guest fixture test now also asserts that
  the island address `bridge_for` resolves is the pool-backed one and
  its file-backed neighbors are not.

Not covered by a unit test: the skip inside `symbolize_user_addr` itself
needs a symbolizer over a live process; a runsc guest capture in the VM
suite is the end-to-end proof (a before/after on the symbolizer's
`Note: symbolizer caches rebuilt` count and its `symbolization done:
anon RSS` line).

## Runtime evidence

Test Suite rows are green at the head; the runtime effect on a host
running gVisor sandboxes is measured after the change ships, on a
capture that sampled sandbox processes, against the two lines named
above.

### DIFF
--- a/src/sandbox_maps.rs
+++ b/src/sandbox_maps.rs
@@ -227,6 +227,30 @@ impl ProcessMaps {
         self.gvisor_exe || self.gvisor_memfd
     }
 
+    /// Whether `addr` lies in a mapping of one of gVisor's memory pools
+    /// (`runsc-memory`, `systrap-memory` — [`GVISOR_MEMFDS`]).
+    ///
+    /// Such an address must never be handed to a symbolizer as a live
+    /// process address: the mapping's `map_files` link opens the pool memfd
+    /// itself — for `runsc-memory` the sandbox's entire MemoryFile, every
+    /// guest page of every process in it, gigabytes — which is not an ELF.
+    /// A symbolizer that resolves the address through the process's maps
+    /// opens that file and parses it from byte 0; whenever the pool's first
+    /// page happens to hold an ELF header (guest images live in the same
+    /// pool), the "symbol tables" it builds are read out of guest memory,
+    /// bounded only by the pool's size. An island's identity comes from its
+    /// file-backed neighbors ([`Self::bridge_for`]) or the guest's own maps
+    /// instead, so skipping the process source loses nothing.
+    ///
+    /// Only the pools count. The sandbox binary that runsc re-execs from a
+    /// memfd (`/memfd:runsc (deleted)`) is a real ELF and keeps the ordinary
+    /// path — the test is the memfd's name, never `memfd:` as a class.
+    pub fn is_pool_backed(&self, addr: u64) -> bool {
+        self.entry_for(addr).is_some_and(|e| {
+            matches!(&e.backing, Backing::Memfd(name) if GVISOR_MEMFDS.contains(&name.as_str()))
+        })
+    }
+
     /// Executable file-backed ranges of this process — for a systrap stub,
     /// the fragments of guest text it maps directly. Used as the
     /// fingerprint for correlating a stub to the guest process it mirrors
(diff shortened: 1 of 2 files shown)


josefbacik:
## Code review

v1.18.0): two Rust files, one commit, +115/−6, no BPF, no schema.

### Summary

A small, contained symbolizer change: `ProcessMaps` gains `is_pool_backed(addr)`
— true iff the maps entry containing the address is a memfd named
`runsc-memory` or `systrap-memory`, the same `GVISOR_MEMFDS` set the
sandbox classification (`is_gvisor`) already keys on — and
`symbolize_user_addr` skips the blazesym process-source call for such an
address, going straight to the island bridge, the guest view and the
host-side label it fell through to anyway. The discriminator keys on the
memfd's name, never on `memfd:` as a class, so the runsc binary's own
exec memfd and any other runtime's exec memfd (a JIT code file) keep the
ordinary path. No frame string changes for any class; what goes away is
the time and symbol-cache memory of parsing a sandbox's whole memory
file as if it were an ELF.

### Critical Issues

None.

  changes), so no `SCHEMA_VERSION` bump, `SCHEMA_CHANGES.md` entry or
  `[bump minor]` title token required.
- BPF: no BPF file moves (no map, no program, no `SEC()`), so the
  map/context pairing question does not arise.
- ELF parsing: none added; the change REMOVES a parse (the process
  source's open of the pool mapping) rather than adding one.
- Unsafe: none.

### Improvements

- `symbolize_user_addr`: the call-site comment (15 lines) restates most
  of `is_pool_backed`'s doc comment, mechanism and cost included. Cut it
  to the two sentences the call site needs — a pool address never goes
  through the process source, and why in one clause — and let the
  rustdoc on `is_pool_backed` carry the mechanism, so the two do not
  drift apart. (Applied by the author before the commit.)
- `is_pool_backed`: the nested `matches!` with two guards reads better as
  `self.entry_for(addr).is_some_and(|e| matches!(&e.backing,
  Backing::Memfd(name) if GVISOR_MEMFDS.contains(&name.as_str())))` —
  one `matches!`, the same semantics. (Applied by the author before the
  commit.)

### Suggestions

- `GVISOR_MEMFDS.contains(&name.as_str())` beside the existing
  `GVISOR_MEMFDS.iter().any(|m| name == m)` in `parse()`: two spellings
  of one test; either is fine, one would be cleaner. Not changed.
- The skip is silent by design. A per-capture count of pool-backed
  addresses routed past the process source, on the symbolizer's existing
  `symbolization done` stats line, would make the effect measurable
  where it lands (the before/after on a sandbox host) without a debug
  build. Optional; a follow-up if wanted.
- The guard's shape `if !pool_backed { if let Some(sym) = … { return …; } }`
  is the edition-2021 form (let-chains are edition 2024); it is fine as
  written.

### Positive Notes

- The discriminator is the memfd's NAME — the same two names the sandbox
  classification already uses — so the change cannot widen to a JIT's or
  another runtime's exec memfd; the test pins that with a non-gVisor
  process carrying `/memfd:jit-code` and with the runsc binary's own
  `/memfd:runsc` mapping.
- `test_is_pool_backed_names_the_pools_only` covers the boundaries that
  matter for an island inside guest text: the island's last byte, the
  byte before it, the file fragment right after it, an unmapped address,
  `[stack]`, and a writable pool page; and the existing guest-fixture test
  in `stack_recorder` now ties the routing to the bridge — the address
  the bridge resolves is exactly the pool-backed one.
- The skip changes no output: every path a pool address takes after the
  process source is the path it took before, so the behavioural
  guarantee ("no frame string changes") is by construction, not by test.

Test Suite, which runs the two new/extended unit tests — are the gate;
the PR stays a draft until they are green. The runtime effect (the
symbolization-time and anon-RSS drop on a host running gVisor sandboxes)
is measured after the change ships, on a capture that sampled sandbox
processes: the `symbolizer caches rebuilt` count and the
`symbolization done: anon RSS` line before and after.

PR #255, opened 2026-09-12
sched: take the sched_slice and thread_state parquet encode off the shard consumers

josefbacik:
## Summary

On a busy many-CPU host the sched/IRQ recorder silently loses events: the BPF side reports `Missed sched/IRQ events` in the millions per 10 s on 128–208-CPU hosts, at a userspace ceiling of about 2.4–2.9 M events/s per host that does not move with the ring layout (64 × 8 MiB or 8 × 64 MiB). This PR takes the parquet encode of the two hot sched tables (`sched_slice`, `thread_state`) off the shard consumers and onto the encode lane the `network_packet` table already uses, and measures the consumer path before and after: **4.6 → 18.5 M events/s at 64 shards on a 64-core host**, no longer flat with the shard count.

It builds on the network-packets consumer change (the encode lane, `src/parquet/lane.rs`; merged as #248, in `v1.18.4`) and changes nothing on the BPF side or in any schema — one commit, `sched: take the sched_slice and thread_state parquet encode off the shard consumers`.

## Where the time went

The consumer side is already sharded: one `SchedEventRecorder` per ring behind its own mutex, one SPSC channel and one `sched_rec_{i}` consumer thread per ring, the lock taken once per 4,096-event batch. The shards contend on nothing except the collector they all flush into every 10,000 records — one `SharedCollector` over one `StreamingParquetWriter`. There, every 200,000 rows of `sched_slice` and of `thread_state`, the writer sorted the batch by `(cpu, ts)` / `(utid, ts)`, built the arrow batch and ran `ArrowWriter::write` (ZSTD-3 + delta encodings) **inline under that shared lock**, on the thread of whichever shard crossed the threshold. For the length of that encode no other shard could hand off; their channels filled, their pollers blocked in `send`, and the BPF rings overflowed.

Measured in release mode on a 64-core host with an ignored benchmark module (`sched::sched_consumer_bench`): N shard threads each drive their own recorder the way the consumers do (4,096-event batches under the shard's lock) with 1 M synthetic events per shard in the shape of the finding (144 CPUs, ~29 K threads; WAKING + SWITCH pairs, every other switch a sleep — 1.25 rows per event, measured 0.500 `sched_slice` + 0.750 `thread_state`). The absolute figures are this host's; the split is the datum.

| collector | 1 shard | 8 shards | 64 shards |
|---|---|---|---|
| a private `InMemoryCollector` per shard (record build + the utid map; no shared lock, no encode) | 13.8 M/s | 45.6 M/s | 45.8 M/s |
| one `SharedCollector(InMemoryCollector)` (the shared lock per flush, no encode) | 11.1 M/s | — | 32.0 M/s |
| **before:** one `SharedCollector(StreamingParquetWriter)`, the encode inline | 4.82 M/s (70 % of the time in flush stalls, worst 31.6 ms) | 5.30 M/s (91 %, worst 59.6 ms) | **4.62 M/s** (95.3 % of the shard-seconds in 5,442 stalls, worst 812 ms) |
| **after:** the same, the two tables on encode lanes | **13.4 M/s** (no stall over 5 ms, worst 4.2 ms) | **21.6 M/s** (worst 22.5 ms) | **18.5 M/s** (worst 159 ms) |

The before row is flat with the shard count — the signature of serialized time on the shared path, not of contention — and the in-memory rows bound it from above: the encode under the lock is the ceiling, not the channel (per-ring SPSC), not the utid `DashMap` (46 M/s), not the lock itself (32 M/s). The after row rises with the shard count; its remaining waits are the shared lock's 10,000-record flushes plus the lane's two-batch queue, about 3.5× above the new figure.

## The change

- `src/parquet/lane.rs`: `EncodeLane::start` takes a `prepare: Option<fn(&mut [R])>` hook, run over each batch on the lane thread before the build — the per-batch sort the inline writer ran on the producer's thread. Test: `prepare_rewrites_each_batch_before_the_build`.
- `src/parquet/writer.rs`: `sched_slice_lane` / `thread_state_lane` replace the two `TableWriter`s; `flush_sched_slices` / `flush_thread_states` hand the `Vec` to the lane through one `push_to_lane` helper (which also carries the `network_packet` lane now: the lane is started on the table's first flush, with the table's sort as `prepare`) and return with a fresh buffer; `finish` closes the lanes where it closed the writers. `add_sched_batch` is overridden on the writer so a shard's flush appends its whole `Vec`s under the lock (one flush check each) instead of pushing a record at a time; the five small sched tables (`irq_slice`, `softirq_slice`, `wakeup_new`, `sched_migrate`, `process_exit`) keep the per-record inline path — their row rates are a fraction of the two hot tables'. Tests: `sched_lane_tests` — every row a shard hands over lands exactly once and each lane batch is sorted by the table's key; an empty batch writes nothing.
- `src/sched.rs`: the benchmark module (ignored; `cargo test --release --lib sched_consumer_bench -- --ignored --nocapture`).

No schema change: the same columns, writer properties, row-group bound (1 M rows) and `ARROW:schema` metadata; the same file names; every reader is unchanged. File sizes on the same 64-shard stream (32 M slices + 48 M thread states) across three runs: `sched_slice.parquet` 61.70 MB before, 62.23 / 61.67 MB after; `thread_state.parquet` 224.6 MB before, 235.5 / 145.0 MB after. The spread is the shards' interleaving — which rows share a 200,000-row batch is decided by 64 threads racing to the shared lock, and the sort key is per batch — not the lane; the per-batch sort is kept and tested.

## What to look at

- The `prepare` hook runs on the lane thread with the batch owned by the lane, before `build_parallel` borrows it — no aliasing with the producer, which has already returned with a fresh `Vec`.
- `add_sched_batch` on the writer: the `should_flush` check after an append can hand the lane up to `batch_size + 10,000` rows (a shard's flush is at most `STREAMING_SCHED_FLUSH_THRESHOLD` records); the lane accepts any batch size and closes row groups at the configured bound regardless.
- A shard blocked on the lane's queue (`LANE_QUEUE = 2` batches) holds the writer lock exactly as the inline encode did, so a lane that cannot keep up still shows as missed events, never as unbounded memory — the same bound the network-packets PR established.
- Not in this PR: the five small tables (inline by design), a per-shard tid → utid cache (the in-memory legs show the map is not a ceiling), and the kernel-side changes.

## Test plan

- `cargo test --release --lib`: 599 passed, 0 failed, 17 ignored (the benchmark legs).
- `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean.
- The benchmark legs above, before and after, on the same host.
- After a release: the `Missed sched/IRQ events` counter under a loaded host, before/after, then the per-capture missed counts on the many-CPU hosts where the ceiling was measured.
- One consequence to plan for: a host that sat above the old ceiling now records up to ~4× the `sched_slice` / `thread_state` rows per capture it used to (the events it was dropping). Anything that post-processes those tables under a fixed time budget — `systing-analyze sched aggregate` run with a deadline, for one — sees that growth on exactly those hosts, so a budget calibrated on the old, truncated traces wants re-reading alongside this change.

### DIFF
--- a/src/parquet/lane.rs
+++ b/src/parquet/lane.rs
@@ -40,6 +40,11 @@ use parquet::schema::types::SchemaDescriptor;
 /// Builds the arrow batch for a slice of rows.
 pub type BatchBuilder<R> = fn(&[R], &SchemaRef) -> Result<RecordBatch>;
 
+/// Rewrites a flushed batch on the lane thread before it is built — the
+/// per-batch sort a table wants for its delta encodings, which the inline
+/// writer ran on the producing thread.
+pub type BatchPrepare<R> = fn(&mut [R]);
+
 /// Flushed row batches the producing thread may run ahead of the lane by
 /// before its `push` blocks. Two 200,000-row batches of the widest record
 /// are under 200 MB; the backpressure is the point — a lane that cannot
@@ -72,12 +77,15 @@ pub struct EncodeLane<R: Send + Sync + 'static> {
 impl<R: Send + Sync + 'static> EncodeLane<R> {
     /// Start the lane: the file writer, the arrow-to-parquet schema and the
     /// `ARROW:schema` metadata are set up on the lane thread exactly as
-    /// `ArrowWriter::try_new` does.
+    /// `ArrowWriter::try_new` does. `prepare`, when given, runs over every
+    /// batch on the lane thread before `build` sees it.
+    #[allow(clippy::too_many_arguments)]
     pub fn start(
         table: &str,
         out: Box<dyn Write + Send>,
         schema: SchemaRef,
         props: WriterProperties,
+        prepare: Option<BatchPrepare<R>>,
         build: BatchBuilder<R>,
         workers: usize,
         row_group_rows: usize,
@@ -94,6 +102,7 @@ impl<R: Send + Sync + 'static> EncodeLane<R> {
                     out,
                     schema,
                     props,
+                    prepare,
                     build,
                     workers.max(1),
                     row_group_rows.max(1),
@@ -179,6 +188,7 @@ fn run_lane<R: Send + Sync + 'static>(
     out: Box<dyn Write + Send>,
     schema: SchemaRef,
     mut props: WriterProperties,
+    prepare: Option<BatchPrepare<R>>,
     build: BatchBuilder<R>,
     workers: usize,
     row_group_rows: usize,
@@ -193,9 +203,12 @@ fn run_lane<R: Send + Sync + 'static>(
         .with_context(|| format!("opening the parquet file of table {table}"))?;
 
     let mut group: Option<RowGroup> = None;
-    for rows in rx {
+    for mut rows in rx {
         if rows.is_empty() {
             continue;
+        }
+        if let Some(prepare) = prepare {
+            prepare(&mut rows);
         }
         let batches = build_parallel(&rows, &schema, build, workers)?;
         drop(rows);
@@ -433,7 +446,7 @@ mod tests {
         let props = WriterProperties::builder()
             .set_max_row_group_size(250)
             .build();
-        let mut lane = EncodeLane::start("t", out, schema(), props, build, 3, 250).unwrap();
+        let mut lane = EncodeLane::start("t", out, schema(), props, None, build, 3, 250).unwrap();
         for b in 0..7 {
             let rows: Vec<Row> = (0..100)
                 .map(|i| Row {
@@ -468,6 +481,7 @@ mod tests {
             out,
             schema(),
             WriterProperties::default(),
+            None,
             build,
             2,
             1000,
@@ -497,6 +511,7 @@ mod tests {
             out,
             schema(),
             WriterProperties::default(),
+            None,
             build,
             2,
             10,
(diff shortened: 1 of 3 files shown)


josefbacik:

Mechanics: the repository's `rust-bpf-code-reviewer` agent definition was executed by the author in-session over the full diff of this branch against its base — all five passes (structure, Rust quality, BPF, schema & versioning, integration) — before the branch was pushed. The review ran against `ac7c5b43` (the head of the network-packets consumer change, since merged as #248); the branch was then rebased onto main past that merge with the patch's content unchanged — the same hunks line for line, one hunk's offset moved by the lines main added to `src/parquet/writer.rs` in between. The findings below were applied before the commit that is this PR's head; the "Improvements" and "Suggestions" that remain are stated as such.

### Summary

The change moves the parquet encode of the two hot sched tables (`sched_slice`, `thread_state`) from the shard consumer's thread — where it ran under the one writer lock every shard flushes through — onto the encode lane the `network_packet` table already uses, with a `prepare` hook so the per-batch sort stays on the lane thread. The mechanism it fixes is measured (the benchmark module in the same PR: flat 4.6–5.3 M events/s across 1/8/64 shards before, 13.4–21.6 M/s after), the file contract is tested (every row once, per-batch order kept, the columns unchanged), and no schema, version or BPF source is touched. The code follows the existing lane pattern closely and is easy to follow.

### Critical Issues

None.

- No schema change: no table definition, `SCHEMA_VERSION` or `SCHEMA_CHANGES.md` is touched — verified by the diff's file list (`src/parquet/lane.rs`, `src/parquet/writer.rs`, `src/sched.rs`); the parquet files keep their columns, writer properties, row-group bound and `ARROW:schema` metadata, and `sched_lane_tests` reads the column names back.
- `Cargo.toml` / `Cargo.lock` untouched.
- No BPF source touched; no `unsafe` added.
- Recorder consistency: no recorder table is added or changed; `utid` attribution is untouched.

### Improvements (applied before the head)

1. The lane-start-and-push block was written three times (`sched_slice`, `thread_state`, and the pre-existing `network_packet`) — deduplicated into one `push_to_lane` helper on the writer; the three flush functions are now the buffer swap plus one call.
2. The `prepare` hook and the two sort functions took `&mut Vec<R>` where `&mut [R]` suffices (clippy `ptr_arg`) — changed to slices.
3. The benchmark's `% 2 == 0` tests — `is_multiple_of` (clippy `manual_is_multiple_of`).
4. The contract test's read helper downcast every second column as `Int32Array`; `thread_state.utid` is `Int64` — replaced by a helper that widens either integer type to `i64`.

### Suggestions (not applied)

- `StreamingParquetWriter::add_sched_batch` repeats the trait default's five per-record loops for the small tables after handling the two hot ones as whole appends. A shared helper on the trait would remove the duplication, but the default must keep working for every other collector, and the repetition is a plain loop per table; left as is.
- The benchmark finishes each recorder with `end_ts = i64::MAX / 2` — a large sentinel that keeps every open slice's final duration positive. Fine for a benchmark; a production caller passes the real end instant.
- `finish_lane!` is a local macro next to `close_writer!` in `finish()`; a small generic `fn` would do the same, but the macro mirrors the existing style of that function.

### Positive Notes

- The `prepare` hook is the minimal generalisation of the lane: one optional `fn` pointer, run on the lane thread with the batch owned by the lane, before `build_parallel` borrows it — no new lock, no new lifetime, and the packet lane passes `None`.
- `add_sched_batch` keeps the per-table flush checks exactly where the per-record path had them, so the lane sees batches of `batch_size` to `batch_size + 10,000` rows and nothing else changes shape; the PR body says so.
- The benchmark separates the three candidate ceilings with three collector shapes and reads the discriminator the flat-vs-rising shard curve gives, so the fix is argued from a measurement a reviewer can re-run, not from a code read alone.
- The contract test drives the writer through the same `add_sched_batch` entry the shards use, at a small batch size, and checks both the multiset of rows and the per-batch order.

PR #256, opened 2026-09-12
network-packets: buffer headroom, exact ids on a failed append, quieter consumers

josefbacik:
## Summary

Follow-ups on the network-packets consumer path from the review of #248. None of them changes what a successful capture writes; they close three small holes on the consumer's own bookkeeping and add the file-contract test the review asked for.

## The changes

1. **No regrow on the crossing append** (`src/parquet/writer.rs`). `add_network_packet_batch` extends the packet buffer by a whole consumer batch and flushes when the buffer has reached `batch_size`, so the buffer crosses the bound by up to one batch. A buffer sized to the bound alone therefore reallocated — and copied its 200,000 records — on the crossing append of nearly every cycle. The writer now remembers the largest batch it has been handed and sizes the buffer, and the fresh buffer after each flush, to the bound plus that headroom.
2. **Exact ids on a failed append** (`src/network_recorder.rs`). `append_packet_batch` emitted each socket record and assigned each id in one loop before the collector call, so a socket-record error dropped the whole batch with the ids it had consumed, and a collector error dropped it with every id consumed — and `packet_events_recorded()`, the exit summary's denominator, counted them. The socket records now go first, the ids are assigned after, and they are handed back when the collector refuses the batch: a failed batch consumes no ids.
3. **One warning per doubling** (`src/systing_core.rs`). A failed append was one `eprintln!` per 4,096-event batch per ring until the capture ended — on a 64-ring host behind a stopped lane, a stderr flood. The packet consumers now share one failure counter and report on the first failure and then at every power of two, with the count. The consumers keep draining their rings, as before: the batches all end on one encode lane, so a lane that has stopped fails every one of them, and the capture's `finish` returns the lane's own error.
4. **The lane's file contract, tested through the production path** (`packet_lane_tests`). Two collector handles hand batches to one `SharedCollector` over a `StreamingParquetWriter` whose packet table flushes to its `EncodeLane` with the production writer properties; the file is read back with the arrow reader and checked for the table's schema and every record exactly once, in arrival order. A second test pins the buffer sizing. The bench module's record factory is shared with them.

## A correction to #248's description

Its "closes a row group exactly where `ArrowWriter` closed one" is not exact: the lane closes a row group after the batch that reaches `max_row_group_size`, so a row group holds between the bound and the bound plus one lane batch (1,000,000 to about 1,020,000 rows at the default sizes). Readers are unaffected.

## Test plan

- `packet_lane_tests::packet_rows_through_the_lane_come_back_whole_and_in_order` and `::the_packet_buffer_holds_the_crossing_append_without_a_regrow`, new; the existing suite unchanged.
- `cargo fmt --check` passes; clippy and the tests did not run before the push (no local build for this lockfile) — the repository's Clippy / Test Suite rows are the gate and the PR stays a draft until they are green.
- No schema, version or BPF change; no new dependency.

### DIFF
--- a/src/network_recorder.rs
+++ b/src/network_recorder.rs
@@ -1103,7 +1103,10 @@ impl NetworkRecorder {
     /// part of the packet path that needs the recorder: the socket record
     /// on a socket's first sighting, the packet ids in arrival order, the
     /// trace's minimum timestamp, then ONE collector call for the whole
-    /// batch (one collector lock, one buffer extend).
+    /// batch (one collector lock, one buffer extend). The socket records go
+    /// first and the ids are handed back on a collector error, so a batch
+    /// that fails consumes no ids and [`Self::packet_events_recorded`]
+    /// counts only records the collector took.
     pub fn append_packet_batch(&mut self, batch: PreparedBatch) -> Result<()> {
         let PreparedBatch {
             mut records,
@@ -1115,8 +1118,11 @@ impl NetworkRecorder {
         if let Some(min_ts) = sockets.iter().map(|s| s.ts).min() {
             self.track_min_ts(min_ts);
         }
-        for (record, socket) in records.iter_mut().zip(&sockets) {
+        for socket in &sockets {
             self.maybe_emit_socket_record(socket)?;
+        }
+        let first_id = self.next_packet_id;
+        for record in records.iter_mut() {
             record.id = self.next_packet_id;
             self.next_packet_id += 1;
         }
@@ -1124,7 +1130,11 @@ impl NetworkRecorder {
             .streaming_collector
             .as_mut()
             .ok_or_else(|| anyhow::anyhow!("Streaming collector not set in append_packet_batch"))?;
-        collector.add_network_packet_batch(records)
+        if let Err(e) = collector.add_network_packet_batch(records) {
+            self.next_packet_id = first_id;
+            return Err(e);
+        }
+        Ok(())
     }
 
     pub fn handle_epoll_event(&mut self, event: crate::systing_core::types::epoll_event_bpf) {
--- a/src/parquet/writer.rs
+++ b/src/parquet/writer.rs
@@ -174,6 +174,10 @@ pub struct StreamingParquetWriter {
     /// The network_packet table encodes off the flushing thread: see
     /// [`EncodeLane`]. Started on the first flush, finished with the others.
     network_packet_lane: Option<EncodeLane<NetworkPacketRecord>>,
+    /// The largest packet batch a consumer has handed over. A fresh packet
+    /// buffer is sized to `batch_size` plus this, so the append that carries
+    /// the buffer past the flush bound never regrows it.
+    network_packet_headroom: usize,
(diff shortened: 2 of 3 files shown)


josefbacik:
## rust-bpf-code-reviewer verdict — `network-packets: buffer headroom, exact ids on a failed append, quieter consumers` (head `01f8dc060c3c03693e1d372b5d279c18bb7ab4ea`)

Mechanics: the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session by the author over the full diff against `v1.18.9` (`104c2fda`; three files, +182/−9), in the definition's five passes. No BPF source is touched and no schema, table definition, `SCHEMA_VERSION`, `Cargo.toml` or `Cargo.lock` changes, so the BPF and schema-and-versioning passes have nothing to flag; the Rust-quality and integration passes are the substance and were read against the consumer path #248 landed (`EncodeLane::push`, `StreamingParquetWriter::flush_network_packets`, `NetworkRecorder::append_packet_batch`, the `packet_rec_{i}` threads).

### Summary

Three small corrections on the packet consumer's bookkeeping — the buffer's capacity headroom, the id accounting on a failed append, the failure report's rate — and the file-contract test the review of #248 asked for. Each change is local, keeps the existing structure, and is explained where it sits; a successful capture writes exactly what it wrote before.

### Critical Issues

None.

- `Cargo.toml` / `Cargo.lock` untouched; no schema file, no `SCHEMA_CHANGES.md` entry owed, no `[bump minor]` owed.
- No BPF source; no `unsafe` added.
- Recorder consistency: no recorder table added or changed; `utid` attribution untouched (the packet rows carry none by design).

### Improvements

None owed for this change.

### Suggestions

- `the_packet_buffer_holds_the_crossing_append_without_a_regrow` asserts the fresh buffer's capacity is equal to the first buffer's; that equality rests on `Vec::reserve` on an empty vector allocating exactly the request, which the standard library does today and documents only as "at least". A `>=` form would be robust to that but would no longer detect a doubling; left as the exact form, with this note.
- After a lane failure the records already buffered in the writer at that flush are lost with the failed push and stay in `packet_events_recorded()` — at most one flush's worth, and the capture's `finish` reports the failure; making the writer report what it dropped would close that residual but is not worth the interface for an error path.

### Positive Notes

- The headroom is one field, updated with `max` and read in the two places a packet buffer is sized; the flush bound and the lane's batch sizes are unchanged, so the encode lane sees what it saw before.
- Reordering `append_packet_batch` (sockets first, ids after, the ids handed back on refusal) makes the id counter exactly the records the collector accepted, with no change to the ids a successful batch receives.
- The shared failure counter's power-of-two report keeps the first failure's error text and the running count visible without a line per batch, and the consumers still drain, so the ring pollers never block behind a dead lane.
- The contract test runs the production path (`SharedCollector` → `StreamingParquetWriter` → `EncodeLane` with `build_writer_properties()`) end to end and reads the file back with the arrow reader; reusing the bench's record factory keeps the test's rows the shape the recorder emits.

PR #257, opened 2026-09-12
analyze: size the sched-aggregate guards from an event histogram, not from time

josefbacik:
## Summary

The bounded `sched aggregate` fold (#250) keeps two guards — an event
budget that shortens the window from its end, and a chunk size that
bounds what the engine sorts at once — and sized both in TIME: the
budget cut in proportion to the row count, the chunks as equal time
slices. That is exact only for a window of uniform density. A window
whose events crowd one stretch of it (a burst, a busy first second)
keeps more rows than the budget after the proportional cut, and hands
the burst to one chunk whole — so the sort memory the chunking was
introduced to bound comes back on exactly the traces it was for (a
clean engine error under the memory limit, never a wrong row).

This PR sizes both guards from a histogram of the stream's event
timestamps instead.

## The change

- One vectorized scan of the same four arms the stream query reads —
  slice starts, slice ends, runnable markers, new-task wakeups (and
  migrates on newer traces) — into 4096 time buckets of the window,
  laid out so that the running total of buckets `0..k` is an EXACT
  upper bound of the stream rows a window ending at the k-th boundary
  holds: an end, marker, wakeup or migrate sitting on a boundary
  belongs to that window and counts below it; a slice start on a
  boundary does not (its slice ends after the window) and counts above
  it.
- **The budget cut** is the last boundary whose running total fits
  `--max-rows`, so the fold never runs over the budget; the recount
  after the cut guards that bound and refuses the fold (an error naming
  the figures) rather than exceed it. When not even the first bucket
  fits, the proportional sliver cut stays as the fallback (the existing
  `max_rows: 1` behaviour, byte for byte).
- **The chunk edges** are the boundaries where the running total crosses
  each `k/n` of the window's rows, so every chunk holds about
  `--chunk-rows` events whatever the density, and can exceed that only
  by a burst narrower than one bucket (1/4096 of the window), which
  nothing at that resolution can split. `stream_chunks` takes an explicit
  edge list and tiles the window exactly for any input.
- The bucket arithmetic runs in 128-bit integers (HUGEINT in the engine,
  i128 for the boundaries), which also retires the equal-time chunking's
  i64 `span * k` product — it could wrap on a long window with many
  chunks and mis-tile the window with no stamp.
- The CLI's truncation line names what it counts (the requested window's
  `sched_slice` rows, idle included) and the folded span, and says
  "below" of the figures that follow it on stdout.

## What does not change

A window of uniform density cuts and chunks exactly as before: the
existing truncation test's figures (the half window at exactly 500 ms,
the 227 µs sliver, the chunk counts) hold unchanged, and the fold's
result is the single query's for any edge list. `SchedAggregateMeta`'s
JSON is untouched — no new field — and the `--chunk-rows` / `--max-rows`
defaults are unchanged.

## Cost

One histogram scan per fold when more than one chunk is used, and one
more over the requested window when the budget cuts it: each is a scan
of the timestamp columns with a 4096-group aggregate, about what the
count query the fold already runs costs, and nothing is sorted or
materialized by it.

## Tests

- `event_histogram_query_shape`: the query's arms, filters and the two
  boundary conventions.
- `histogram_boundaries_and_edges`: the boundary arithmetic (a span above
  2^63 / 4096 does not wrap), the running totals, the edge placement, a
  burst that cannot be split, an empty stream.
- `budget_and_chunks_hold_under_a_burst`: a one-second workload whose
  first fifth is five times as dense as the rest (7,920 stream rows,
  4,400 in the first 200 ms). With a budget of 3,000 the cut lands
  inside the dense fifth holding 2,998 rows — the proportional cut kept
  the first 378.8 ms, 5,187 rows, 73 % over. With eight chunks of a
  nominal 1,000 rows every chunk holds at most 1,020 — equal time slices
  hand 2,750 to the first — and the chunked fold equals the single
  query.
- `stream_chunks_cover_the_window_exactly` on an explicit edge list
  (unsorted, repeated, on and outside the window); the existing
  chunked-versus-single-query and truncation tests unchanged.

## Runtime evidence

Test Suite rows are green at the head. The histogram's cost on a
hundreds-of-millions-of-rows trace is not measured here; the bench that
sized the chunk default is the place to read it.

### DIFF
--- a/src/analyze/sched_aggregate.rs
+++ b/src/analyze/sched_aggregate.rs
@@ -94,14 +94,23 @@
 //! - **Chunking** (`chunk_rows`): the stream is fetched in consecutive time
 //!   chunks of the window, each sorted by the engine on its own and consumed
 //!   by the same pass, so at most one chunk is materialized at a time. The
-//!   result is identical to one pass over the whole stream — an ordered
-//!   stream is the concatenation of its ordered time chunks, and a slice
-//!   whose start and end fall in different chunks is two events in the
-//!   stream either way. `meta.stream_chunks` says how many chunks were used;
-//!   1 means the fold ran as a single query.
+//!   chunk edges come from a histogram of the stream's event timestamps
+//!   (one vectorized scan into [`EVENT_HIST_BUCKETS`] time buckets): an edge
+//!   sits where the cumulative event count crosses each `k/n` of the
+//!   window's rows, so every chunk holds about `chunk_rows` events whatever
+//!   the density inside the window — equal time slices would hand a burst
+//!   to one chunk whole — and a chunk can exceed that only by a burst
+//!   narrower than one bucket. The result is identical to one pass over the
+//!   whole stream — an ordered stream is the concatenation of its ordered
+//!   time chunks, and a slice whose start and end fall in different chunks
+//!   is two events in the stream either way. `meta.stream_chunks` says how
+//!   many chunks were used; 1 means the fold ran as a single query.
 //! - **The event budget** (`max_rows`): when the requested window holds more
-//!   stream rows than the budget, the window is shortened from its END so
-//!   that it holds about the budget, and the pass runs over that shorter
+//!   stream rows than the budget, the window is shortened from its END to
+//!   the last bucket boundary at which the cumulative event count still
+//!   fits the budget — an exact bound, since every stream row of the shorter
+//!   window is an event of the requested window's stream with a timestamp
+//!   at or before that boundary — and the pass runs over that shorter
 //!   window. Every statistic stays exactly defined over a contiguous,
 //!   un-sampled window — nothing is sampled or extrapolated — and the row
 //!   says so: `meta.window_truncated` is true, `meta.window_ns` is the
@@ -110,7 +119,9 @@
 //!   full one cannot be folded beats no row at all.
 //!
 //! The stream rows of a window are counted before the fold (one aggregate
-//! query per source table) to size both guards.
+//! query per source table) to size both guards, and counted again after a
+//! budget cut so the chunk count and the metadata describe the window
+//! actually folded.
 
 use anyhow::{bail, Context, Result};
 use duckdb::arrow::array::{Array, AsArray, PrimitiveArray};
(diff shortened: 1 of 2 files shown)


josefbacik:
## Code review

v1.18.9): two Rust files, one commit, `src/analyze/sched_aggregate.rs`
(+471/−34 with tests) and `src/bin/systing_analyze.rs` (+5/−3); no BPF,
no schema, no output-shape change.

### Summary

The bounded `sched aggregate` fold sized both of its guards in time — the
event-budget cut in proportion to the row count, the chunks as equal time
slices — which holds only for a window of uniform density: a burst kept
more rows than the budget after the cut and landed in one chunk whole,
bringing back the sort memory the chunking was introduced to bound. The
change sizes both guards from one histogram of the stream's event
timestamps (4096 time buckets, one vectorized scan of the arms the stream
query already reads), laid out so its running total is an exact upper
bound of the rows a window ending at each boundary holds: the budget cut
is the last boundary whose total fits (never over budget, with a recount
that refuses the fold rather than exceed it), the chunk edges are the
boundaries where the total crosses each `k/n` of the rows (balanced to
within a bucket). Uniform windows cut and chunk exactly as before; the
existing tests pass unchanged; the equal-time chunking's i64 `span * k`
product, which could wrap on a long window with many chunks, is retired
with it.

### Critical Issues

None.

  changes); no `SCHEMA_VERSION` bump, `SCHEMA_CHANGES.md` entry or
  `[bump minor]` token required. `SchedAggregateMeta`'s serde output is
  unchanged, so a consumer comparing the JSON against a stored expected
  output sees no change from this PR.
- BPF: none. ELF parsing: none. Unsafe: none.

### Improvements

- The histogram's exactness argument rests on two boundary conventions
  (an end, marker, wakeup or migrate at a boundary counts below it; a
  slice start counts above it). Both are stated in the query builder's
  doc comment and pinned by the shape test; the fold-level proof is the
  truncation test's exact half-window figures (unchanged) and the burst
  test's `folded <= budget`. Nothing further owed.
- On a truncated window the histogram runs twice — over the requested
  window for the cut and over the folded window for the chunk edges. The
  requested window's histogram covers the folded one at coarser
  resolution and could serve both; the second scan costs about what one
  chunk's scan costs and keeps the chunk edges at full resolution, so the
  simpler form is kept. Noted, not changed.

### Suggestions

- `event_histogram` reads the bucket rows through the row API; at most
  4096 rows, so the per-row cost that moved the fold to Arrow batches
  does not apply. Fine as written.
- The CLI's truncation line now says "the requested window holds N
  sched_slice rows (idle included); every figure below is exact over the
  first X s of it" — the two "slices" figures that shared a word on
  adjacent lines are named apart, and "below" matches where the figures
  print (stdout, after the stderr header).

### Positive Notes

- The bound is exact by construction, not by tolerance: the docs say why
  the running total bounds the shorter window's rows, and the recount
  after the cut is a guard on that argument rather than a second guess.
- `stream_chunks` takes an explicit edge list and tiles the window
  exactly for any input (unsorted, repeated, on or outside the window),
  and the test says so.
- The burst test is built from the same step as the existing workload at
  two rates, so its figures (7,920 rows, 4,400 in the first fifth; the
  proportional cut's 5,187; equal-time chunking's 2,750) are derived from
  the fixture, not asserted from a run.

Test Suite, which runs the four new tests beside the existing
chunked-versus-single-query and truncation tests — are the gate; the PR
stays a draft until they are green. The histogram query's cost on a
hundreds-of-millions-of-rows trace (one scan of the four tables' timestamp
columns with a 4096-group aggregate, twice on a truncated window) is not
measured here; it is bounded by the count query's scan the fold already
runs, and the bench that sized the chunk default is the place to read it.

PR #258, opened 2026-09-12
analyze: read escape strings and refuse non-ASCII as DuckDB's lexer does

josefbacik:
## Why

The statement classifier in `src/analyze/query.rs` stands in front of the MCP `query` tool because the binding's `prepare` is not parse-only: it executes every statement but the last of the text it is handed, and a `PIVOT` without a literal `IN (...)` list is expanded by DuckDB's parser into a `CREATE TYPE __pivot_enum_... AS ENUM (...)` statement that `prepare` would then run outside the row cap. #243 made that guard refuse every pivot shape but one literal top-level statement — but the guard is only as good as the classifier's agreement with DuckDB's own lexer about where a string ends and where a statement ends, and a read of the lexer (`third_party/libpg_query/scan.l`) and of `Parser::ParseQuery` against the three scanners found two disagreements, each letting a refused shape back through behind a literal:

- **Escape-string literals.** DuckDB lexes `E'\''` as ONE string: `xestart` is `[eE]{quote}`, and inside the `xe` state `xeescape [\\][^0-7]` consumes a backslash and the character after it. The scanners had no escape rule, so they ended the string at the `\'` and read everything from there to the next lone `'` as string — code to DuckDB. `SELECT E'\'' AS q FROM (PIVOT t ON bucket USING count(id)) WHERE '' = ''` classified as a single plain statement (a nested unlisted pivot reached `prepare`), and `SELECT E'\''; SET memory_limit = '100GB'; SELECT E'\''` classified as one statement with the first two executed.
- **Stripped Unicode spaces.** `Parser::ParseQuery` runs `StripUnicodeSpaces` first, replacing U+00A0, U+2000–U+200B, U+202F, U+205F, U+2060, U+3000 and U+FEFF outside quotes with an ASCII space; Rust's `char::is_whitespace` excludes U+200B, U+2060 and U+FEFF, so `PIVOT t ON bucket IN (\u{200B}SELECT DISTINCT bucket FROM t) USING count(id)` read as a fully listed literal list here and as `IN (<subquery>)` to DuckDB, which builds the enum by running the subquery at prepare.

Both are reachable only from the MCP `query` tool's text — SQL an analysis session authors — so they are an adversarial or prompt-steered exposure, not one a normal query hits; they are fixed the same day for the same reason the guard exists.

## What

- One helper, `quoted_run_end`, now walks every quoted run for all three scanners (`statement_shape`, `pivot_keyword_count`, `pivot_is_fully_listed`): a doubled quote is an escaped quote, and inside an escape-string literal a backslash escapes the character after it, exactly as the lexer reads it. `opens_escape_string` decides whether a `'` opens such a literal the way the lexer does: the `E` must begin a token of its own — an `E` that ends a longer identifier is the identifier's (`name'\'` is the identifier `name` and an ordinary string), while the `E` after a numeric literal IS the prefix (`1E'\''` is the number 1 and an escape string, because a number's exponent needs digits after its `E`).
- Any non-ASCII character outside a quoted run or a comment is refused as a new `StatementShape::NonAscii(char)` with a message naming the character; a trace database's schemas are ASCII, and the alternative is tracking DuckDB's replacement set version by version. Non-ASCII text inside a literal, a quoted identifier or a comment is untouched.
- `PIVOT_LONGER`, DuckDB's other spelling of `UNPIVOT`, joins `QUERY_KEYWORD` (a list opening with it is a subquery), `NOT_A_PLAIN_ELEMENT`, and the must-wrap statements beside `UNPIVOT`.
- The module doc's claim that the escape-string mismatch could only make the scan see MORE separators is replaced by what the scan models and what it refuses; the `query` tool's description names the non-ASCII refusal.

## Tests

- `statement_shape` unit tests: the two hiding shapes classify as `Multiple` / `PivotWithoutIn` (upper- and lower-case `E`, and after a numeric literal); an escape string in a single statement stays one statement with its `\'`, `;` and `--` intact; the prefix rule is pinned on identifier and numeric cases (`name'`, `x1E'`, `_e'`, `1ee'` are not the prefix; `E'`, `1E'`, `1e5E'`, `1e'` are) and `quoted_run_end`'s ends on each string form; the three stripped zero-width spaces, U+00A0, a non-ASCII identifier and an em dash are refused outside literals and kept inside literals, quoted identifiers and comments; `PIVOT_LONGER` as a list is refused and as a statement must wrap.
- Through `AnalyzeDb::query` on a real database: the escape-string `;` shapes are refused as several statements and the hidden `SET` never runs (`memory_limit` unchanged), `SELECT E'\'' AS q` runs and yields `'`, the zero-width-space and non-breaking-space texts are refused as non-ASCII, and the nested pivot behind an escape string, the `PIVOT_LONGER` list and the zero-width-space subquery list each leave the `__pivot_enum%` type count at 0.

## Runtime evidence

`cargo fmt --all -- --check` is clean. The classifier's unit tests (the nine `statement_shape` tests, the new ones included) were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs`, 9 passed, and the same file is clean under clippy's default lint set with `-D warnings`. The crate's clippy and full test suite — the `AnalyzeDb::query` binding tests above among them — were NOT run before this push: no cargo build resolves on the host this was written on, so this PR's own CI rows are that proof, and the PR stays a draft until they are green at the head.

## Not changed

`cgroup_id(path)` and `File::open(path)` in `systing_core.rs` are still two separate resolutions of the `--cgroup` path (the readback later reports a disagreement as the generic mismatch); reading the id from the held descriptor's inode is a small follow-up, not this change. The `query` tool's over-refusals (the SQL-standard `FROM t PIVOT (...)` form, `IN some_enum_type`, a CTE-prefixed literal pivot) are unchanged and safe as refused.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of this PR's head at the time of the run (one commit on v1.18.9: `src/analyze/query.rs`, `src/analyze/mod.rs`, `src/mcp.rs`; no BPF file, no schema, no `Cargo.*`); verdict as it produced it.

### Summary

The change makes the three text scanners share one quoted-run walker that reads DuckDB's escape-string literal the way its lexer does, refuses non-ASCII outside literals instead of modelling DuckDB's pre-lex space replacement, and closes the `PIVOT_LONGER` gap. It is small, direct, and every rule it adds is pinned by a test at the classifier and, where DuckDB's behaviour is the claim, at the binding. The prefix rule's numeric-literal case is the one subtle part and is documented and tested.

### Critical Issues

None.

### Improvements

- Applied before the commit: the three call sites first passed the escape decision into `quoted_run_end` as a flag computed identically at each (`c == '\'' && opens_escape_string(&chars, i)`); the helper now decides it itself from the opening quote, so a fourth scanner cannot get it wrong by omitting the flag.
- `opens_escape_string` reads the identifier run ending at the `E` as the lexer would tokenize it (digits → an optional exponent → the rest an identifier). The loop is the one non-obvious piece of logic in the change; its doc comment states the lexer facts it rests on and the tests cover the identifier, underscore, digit, exponent and non-ASCII-preceded cases. No change requested.

### Suggestions

- `StatementShape::NonAscii(char)` carries the first offending character only; the refusal message names it with its code point, which is enough to locate it in a query of ordinary length. If refusals of this kind turn out to be common in practice, an offset would help more than the character — not needed now.
- The `query` tool's description is already long; the added clause is one phrase. Fine as is.

### Positive Notes

- The doc on `statement_shape` now says exactly what the scan models (ordinary and escape-string runs) and what it refuses (dollar quoting, non-ASCII outside literals) instead of the earlier claim that any mismatch could only err toward more separators, which the escape-string case had falsified.
- The binding tests assert the property that matters — the hidden `SET` never runs and no `__pivot_enum` type is created — rather than only the classifier's verdict.
- `test_statement_shape_reads_escape_strings_as_duckdb_does` documents both directions: the shapes an escape string hides from a naive scanner, and the shapes where the escape string legitimately contains a `;` or a `PIVOT` as text.

### Runtime evidence

The classifier's unit tests were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs` (9 passed) and the file is clean under clippy's default lint set with `-D warnings`; `cargo fmt --all -- --check` is clean on the tree. The crate's own clippy and test run (the `AnalyzeDb::query` binding tests included) did not run before the push — no cargo build resolves on the host this was written on — so this PR's CI rows are the gate, and the PR stays a draft until they are green at the head.

PR #259, opened 2026-09-12
recorder: write a one-row manifest (version, schema, finish time) and record it on _traces [bump minor]

josefbacik:
The recorder records who it is. A parquet directory carried no writer version: `_traces.systing_version` names the systing that CONVERTED the directory (schema 12), so a consumer could not tell "this column is NULL because the recorder predates it" from "the reader predates the recorder and dropped it at import" (the import-skew guard of #244), and the guard's warning could not name the writer whose columns it dropped.

## What changes

- **`systing_manifest`, a one-row parquet table** written by the streaming writer's `finish()` after every table's last flush — so its presence means the writer completed — through the same sink as every table: `systing_manifest.parquet` in a directory, the `systing_manifest` header on `--stream` (one `TABLE_NAMES` entry; the receiver writes it as a file like any table). Columns: `systing_version` (the recorder's `CARGO_PKG_VERSION`), `schema_version` (the `SCHEMA_VERSION` it wrote against), `recorded_at_unix_ns` (the wall clock at finish). Its shape is fixed at its introduction so every reader from schema 22 on can read a manifest written by any later systing.
- **`_traces` gains `recorder_version`, `recorder_schema_version`, `recorded_at_unix_ns`** (nullable), lifted from the manifest at import (`parquet_to_duckdb`, `systing-util convert`; a DuckDB→DuckDB merge copies them when the source has them). A directory without a manifest — written before schema 22, or by a writer that never reached `finish()` — imports with the three NULL and no line at all; a manifest that is present but unreadable is reported once and read as absent, never a failure. `systing_version` keeps its converter meaning. `systing-analyze trace info` / the MCP `trace_info` tool report the pair per trace under `trace_versions` (absent when NULL); the profile export's header gains `source_recorder_version`.
- **The guard's messages name the writer** when the manifest is present ("the trace was written by systing 1.N (schema K)") and say so when it is not.

## Follow-ups from a second review of #244, in the same PR

- `stack.parquet` is under the guard: its import selects its columns by name, so an extra column never failed it and `--strict-schema` did not refuse it; its columns are now read first, an extra one is reported under table `stack` or refused under `--strict-schema`, and the known columns import as before.
- A `.parquet` file in the directory this systing has no table for is reported by name (`ImportReport::unknown_files`, one warning line) and refused under `--strict-schema`; the tables around it import.
- A refused import leaves no database behind: `parquet_to_duckdb` removes the `.duckdb` it created (and DuckDB's `.wal` beside it) on any error, so a caller never finds a hollow-but-valid database where a refusal happened.
- `systing-util convert` prints one `Schema note:` line at the end of a run that left anything out; its `--strict-schema` help says what the flag covers (every table of a parquet directory, `stack.parquet` included) and what it does not (a `.duckdb` input merges by column intersection).
- The `import_column_list` unit test's projected INSERT selects its columns out of the table's order, so the row proves the mapping is `BY NAME`.

## Schema

`SCHEMA_VERSION` 21 → 22 (`[bump minor]`); the SCHEMA_CHANGES entry is under "Schema Version 22 (unreleased)". Additive: no table or column changes shape; a reader older than this release imports a directory that carries `systing_manifest.parquet` with that file ignored (it is not in its `ParquetPaths`) and, from v1.18.2, any newer table column projected away with a warning — the first skew window the #244 guard absorbs.

## How to read the pair

`recorder_schema_version` above the database's `_schema_version` means the reader predated the recorder and dropped the columns it did not know at import; a NULL column in a trace whose `recorder_schema_version` is below the column's schema is one the recorder predates. On the first release after this one, traces written by older recorders read NULL for the pair (the era boundary); the release after that is the first where every fresh trace carries it.

## Tests

`test_manifest_written_at_finish_and_imported` (the writer writes it last; the import records it; a writer dropped without `finish()` leaves none), `test_import_records_the_recorder_from_the_manifest` (the pre-manifest directory reads NULL with no line; a newer-schema manifest beside an unknown column is recorded and the guard's messages name the writer), `test_stack_parquet_extra_column_is_reported_or_refused`, `test_strict_refusal_removes_the_database`, `test_unknown_parquet_files_are_reported`, `test_import_preserves_systing_version` (the pair copied by the DuckDB→DuckDB merge) and `test_import_from_a_source_without_recorder_columns` (a pre-22 source merges with the pair NULL); the ignored `--stream` round-trip expects `systing_manifest.parquet` among the receiver's files. The tree was not built on the author's host; this PR's CI is the compile and test gate.


josefbacik:

Executed in-session by the author over the full diff of the branch against its base `104c2fda` (v1.18.9): `src/duckdb.rs` (+≈640/−≈100 incl. tests), `src/parquet/writer.rs` (+≈180), `src/bin/systing_util.rs` (+≈90/−≈20), `src/analyze/mod.rs`, `src/profile_export.rs`, `src/parquet_paths.rs`, `src/stream.rs`, `src/trace/{models,schema}.rs`, `src/mcp.rs`, `tests/stream_unix.rs`, `SCHEMA_CHANGES.md`, the repository's contributor guide, `docs/PROFILE_EXPORT_FORMAT.md`, `skills/systing-analyze/SKILL.md`. No BPF and no ELF parsing: passes 3 and the binary-parsing criteria have nothing to judge; passes 1, 2, 4 (schema and versioning) and 5 and the cleanliness criteria apply. The tree was not compiled or tested on the author's host; the pull request's own CI (Clippy, Rustfmt, Test Suite) is the compile and test gate, and `cargo fmt --all -- --check` passed locally.

### Summary

The recorder writes a one-row `systing_manifest` table (its version, its `SCHEMA_VERSION`, the wall clock at finish) through the same sink as every table, and the import lifts it onto three nullable `_traces` columns — the provenance the import-skew guard lacked. The guard is extended to `stack.parquet` (whose import selects by name and never failed on an extra column) and to whole parquet files the reader has no table for; a refused import removes the database it created; the guard's messages name the writer. Schema 22, additive: a directory without a manifest imports exactly as before, with the three columns NULL.

### Critical Issues

None. `SCHEMA_CHANGES.md` carries the schema 22 entry, `SCHEMA_VERSION` is 21 → 22, `Cargo.toml` is untouched (the title carries `[bump minor]`), and the new table attributes to no thread (the `utid` rule does not apply).

### Improvements

- Applied before the commit: the "unknown columns → warn and report, or refuse" tail was written twice (the `BY NAME` tables' guard and the new `stack.parquet` check) and the "unknown files" tail twice (the library import and `systing-util convert`); both now go through one helper each (`note_unknown_columns`, `note_unknown_files`), so the messages and the strict/lax split cannot drift apart.

### Suggestions

- `import_tables` keeps one closure for every table and branches on the name `stack` inside it, because a second closure borrowing the same report would not compile beside it; a `report` parameter on the closure would let the stack import stay a separate call at the cost of touching every call site. Kept for the smaller diff.
- `recorded_at_unix_ns` is a `BIGINT` of Unix nanoseconds rather than a `TIMESTAMP`, the tree's clock idiom (`timestamp_ns`, `ts`); a consumer that wants a timestamp converts with `to_timestamp(recorded_at_unix_ns / 1e9)`.
- A manifest that is present but unreadable is warned about and read as absent, on the argument that provenance must not cost the tables; a reader that would rather refuse can check `ImportReport::recorder` against the file's presence.

### Positive Notes

- The manifest is one `TABLE_NAMES` entry and one `ParquetPaths` field, no second mechanism: the directory writer, the `--stream` receiver and the DuckDB import all handle it as a table, and DuckDB can `SELECT` it directly.
- The missing-manifest era is silent by design (no line on thousands of existing traces), the guard's warning gains the writer only when it can name one, and the tests cover both directions of the pair (`recorder_schema_version` above and below the reader), the strict refusal's cleanup, the unknown-file report and the DuckDB→DuckDB merge with and without the columns.

josefbacik:
## Code review addendum (rust-bpf-code-reviewer) — the Clippy fix commit

Executed in-session by the author over the fix commit's own diff (`src/duckdb.rs`, +9/−8, one hunk in `import_directory`), on top of the first commit's review above. The first CI run at the PR's original head: Rustfmt, Test Suite (every new test included) and the three load-shape kernels green; Clippy red on one lint, `field_reassign_with_default`, denied under `-D warnings` — `ImportReport::default()` followed by two field assignments.

### Summary

The two field assignments become a struct literal with `..Default::default()`; the value, the evaluation order (the manifest read, then the directory scan) and every message are unchanged. The comment above the block is folded into one paragraph.

### Critical Issues

None. No schema, version or BPF change.

### Improvements

None.

### Suggestions

None.

### Positive Notes

- The rewrite is the lint's own suggestion, verbatim in effect, and the tests that exercise this path (the pre-manifest directory, the newer-schema manifest, the unknown-file report, the strict cleanup) passed on the first run.

PR #260, opened 2026-09-12
analyze: refuse a string literal continued across a newline

josefbacik:
## Why

#258 made the statement classifier in front of the MCP `query` tool read DuckDB's escape-string literals the way the lexer does, because the binding's `prepare` executes every statement but the last of the text it is handed, and a string form the classifier reads differently from the lexer is a place for a `;` to hide. One string form was still read differently.

DuckDB's lexer (`third_party/libpg_query/scan.l`: `quotecontinue` is a quote, whitespace containing at least one newline, then a quote, ignored in the `xq`, `xe` and `xus` states) **continues a string literal across a newline**: `'a'` + newline + `'b'` is the one literal `ab`, and the lexer stays in the state the first part opened, so an escape-string literal's backslash rule reaches into the second part. The classifier closed every `'` run at its quote and read the second part as a fresh ordinary string. So for

```
SELECT E''
'\'') AS q LIMIT 1; SET memory_limit = '100GB'; SELECT 1 FROM (SELECT 1 --'
```

the classifier read `E''`, then an ordinary run `'\'') AS q LIMIT 1; SET memory_limit = '` whose `\` is just a character, and so on — one statement, nothing stripped — while DuckDB read `E''⏎'\''` as one escape string closing at the `'` before the `)`, then two more statements. Wrapped as the tool wraps it, DuckDB extracts three statements and `prepare` executes the second. Observed, not only traced: on DuckDB 1.5.5 through the Python package, `extract_statements` on the wrapped text returns 3 and executing it changes `memory_limit`; `'a'⏎'b'` yields `ab` with `\n`, with `\r` and with mixed whitespace; `E'a'⏎'\'x'` yields `a'x` (the escape mode carries); `'a' 'b'` on one line is a syntax error (no newline, no join); `'a' -- c⏎'b'` yields `ab` (a `--` comment before the newline is part of the continuation's whitespace); `'a'/*x*/⏎'b'` is a syntax error in the original text but `ab` once the comment is a space.

That last case matters for where the check runs: the classifier strips comments before the text reaches DuckDB, so it can create a continuation the original text did not have. The check is therefore made on the stripped text — the text DuckDB receives.

Reachable only from the `query` tool's text, so an adversarial or prompt-steered exposure rather than one an ordinary query hits; fixed the same way and for the same reason as #258.

## What

- A `'` run whose closing quote is followed, across at least one newline and nothing but whitespace, by another `'` is refused as `StatementShape::QuoteContinuation` (`has_quote_continuation`, one pass over the stripped text through the same quoted-run walker every scanner uses). The message says what to write instead: the literal on one line, or the parts joined with `||`. Refusing is the smaller and safer change: modelling the join would mean re-deriving it on the stripped text for all three lexer states, and the shape has no use through this tool.
- A `--` comment ends at `\r` as well as `\n`, as in the lexer (before, a `\r`-terminated line comment swallowed the statement after it here while DuckDB ran it).
- An integer literal with digit separators (`1_0`) is a number ahead of an escape-string prefix (`integer_literal_len`), so `1_0E'` opens an escape string as the lexer reads it. DuckDB's grammar rejects a number directly followed by a string, so this closes no reachable path; it keeps the prefix rule in step with the lexer.
- The module doc and the `query` tool's description name the new refusal.

Not changed: `\v` is whitespace to the classifier and a syntax error to DuckDB — the safe direction, left as is.

## Tests

- `statement_shape` unit tests: the hiding shape and its `\r`, mixed-whitespace, block-comment and line-comment variants classify as `QuoteContinuation`, a plain two-line literal is refused on its own and beside other statements, and the shapes that are not continuations — a separator between the parts, no newline between them, double-quoted identifiers, a mixed pair, a closed escape string nothing continues — still classify as one statement; the `\r` comment end; the digit-separator prefix cases and `integer_literal_len`'s bounds.
- Through `AnalyzeDb::query` on a real database: the hiding shape, its `\r` twin, the comment-created continuation and a plain two-line literal are refused with the new message and `memory_limit` is unchanged, and `'a' ||⏎'b'` runs and yields `ab`.

## Runtime evidence

`cargo fmt --all -- --check` is clean. The classifier's unit tests (the ten `statement_shape` tests, the new one included) were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs`, 10 passed, and the same file is clean under clippy's default lint set with `-D warnings`. DuckDB's behaviour on every case above was observed on DuckDB 1.5.5 through the Python package, as described. The crate's clippy and full test suite — the `AnalyzeDb::query` binding tests among them — were NOT run before this push: no cargo build resolves on the host this was written on, so this PR's own CI rows are that proof, and the PR stays a draft until they are green at the head.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session by the author over the full diff of this PR's head at the time of the run (one commit on v1.18.12: `src/analyze/query.rs`, `src/analyze/mod.rs`, `src/mcp.rs`; no BPF file, no schema, no `Cargo.*`); verdict as it produced it.

### Summary

The change closes the last string form on which the statement classifier read a boundary differently from DuckDB's lexer: a literal continued across a newline. It refuses the shape instead of modelling it, judges it on the text DuckDB actually receives (the comment-stripped text), and aligns two smaller lexer details (`--` at `\r`, digit separators in the numeric prefix). Every rule is pinned at the classifier and, where DuckDB's behaviour is the claim, at the binding, and the lexer claims were observed on a DuckDB build rather than only traced.

### Critical Issues

None.

### Improvements

- `has_quote_continuation` is a second pass over the stripped text rather than a hook in the main loop. That is the right shape here — the main loop walks the original text with its comments, and the join has to be judged after stripping — so the extra pass is the honest implementation, not an oversight. No change requested.
- The escape-mode decision inside `quoted_run_end` is evaluated on the original text by the main loop and on the stripped text by the continuation pass. They agree, because the decision reads only the identifier run directly before the quote and stripping never touches identifier characters (a comment becomes a space, and `/` and `-` were not identifier characters either). Worth a one-line note in the doc if this module grows another pass; not needed now.

### Suggestions

- The refusal message's "join the parts with `||`" is exercised by a binding test, so the advice is known to run. Good.
- `\v` remains whitespace to the classifier and a syntax error to DuckDB; the PR says so. Fine to leave.

### Positive Notes

- The tests carry the negative cases (the non-continuations) beside the positives, so the refusal cannot quietly widen to `'a',⏎'b'` or to double-quoted identifiers.
- The doc on `StatementShape::QuoteContinuation` states why the shape is refused rather than joined, including the stripping subtlety that makes "judge the stripped text" the correct rule.
- `integer_literal_len` matches the lexer's `{digit}("_"?{digit})*` exactly, and its edge cases (doubled and trailing `_`) are pinned.

### Runtime evidence

The classifier's unit tests were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs` (10 passed) and the file is clean under clippy's default lint set with `-D warnings`; `cargo fmt --all -- --check` is clean on the tree; DuckDB's lexer behaviour on each case in the body was observed on DuckDB 1.5.5 through the Python package. The crate's own clippy and test run (the `AnalyzeDb::query` binding tests included) did not run before the push — no cargo build resolves on the host this was written on — so this PR's CI rows are the gate, and the PR stays a draft until they are green at the head.

PR #261, opened 2026-09-12
duckdb: build the trace database in one transaction

josefbacik:
## Why

`parquet_to_duckdb` builds a trace database by creating the schema and importing each present parquet file, and it did all of that in autocommit: every statement — the table DDL, the two metadata rows, one `INSERT ... SELECT ... FROM read_parquet()` per present table — was its own DuckDB commit, a WAL write and an fsync-class wait, forty to seventy of them per trace, then the close checkpoint. On an idle disk that is well under a second in all. On a host whose workload holds the disk saturated each of those waits takes seconds and the generation runs for minutes, while the sibling step that writes more bytes to the same disk without fsync (the parquet write) stays sub-second. A caller that builds a small database after every short capture — one process traced every few minutes on a busy host — pays that tail on exactly the hosts that are busiest, and it pays it in wall time between captures.

## What

The schema and every import now run inside one transaction, committed once at the end: one durable write plus the close-time checkpoint instead of one per statement. `configure_for_bulk_io`'s settings are applied before the transaction opens; the transaction guard commits after `import_tables` returns and rolls back when it is dropped on any error path, so the import is atomic as a side effect — a failure part-way through leaves no half-built database with some tables filled and the rest missing, only the caller's error. The column projection for a newer trace (`import_column_list`, the `ImportReport`) is unchanged and runs inside the transaction like the imports it serves.

## Tests

`test_parquet_to_duckdb_failed_import_leaves_no_half_built_database`: a trace exported to parquet, its `sched_slice.parquet` overwritten with bytes that are not parquet, so the import fails at that table after `process` and `thread` were already imported; the error names the table, and the database reopened afterwards holds no table in `main`.

## Runtime evidence

`cargo fmt --all -- --check` is clean. The transaction's behaviour was observed on DuckDB 1.5.5 through the Python package on a file database: `CREATE TABLE`, a view and an `INSERT ... FROM read_parquet()` inside one explicit transaction, then the read of a file that is not parquet failing (`No magic bytes found at end of file`), `ROLLBACK`, and the database reopened with zero tables in `main`; the same statements committed survive a reopen. The crate's clippy and test suite — this test among them — were NOT run before this push: no cargo build resolves on the host this was written on, so this PR's own CI rows are that proof, and the PR stays a draft until they are green at the head. Not measured here: the wall-time gain on a saturated disk (the motivating tail is read from production traces, not reproduced in CI) and the memory profile of a very large import inside one transaction — DuckDB writes large appends to the file ahead of the commit, so the import's memory bound is unchanged by design, and the first traces built at this version are where both are read.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session by the author over the full diff of this PR's head at the time of the run (one commit on v1.18.12: `src/duckdb.rs`; no BPF file, no schema change, no `Cargo.*`); verdict as it produced it.

### Summary

The change wraps `parquet_to_duckdb_with_options`'s schema creation, metadata rows and table imports in one `Transaction`, committing once at the end. It is small, every statement it moves is one the function already ran, and the only behavioural change beyond the commit count is atomicity on the error path, which the new test pins. The column-projection path the import gained recently (`import_column_list` / `ImportReport`) is untouched and runs inside the transaction through the same `&Connection` the imports use.

### Critical Issues

None.

### Improvements

- `create_schema(&tx)` and `import_tables(&tx, ...)` rely on `Transaction`'s `Deref<Target = Connection>` at the call sites, which is the crate's intended use and keeps every helper's signature as `&Connection`. No change requested.
- The doc comment on `parquet_to_duckdb` states the transaction and its two consequences (one commit, empty-on-failure); `parquet_to_duckdb_with_options` is where the work happens and inherits that doc by delegation. Fine as is.

### Suggestions

- The test's failing table is `sched_slice`, chosen because it imports after `process` and `thread`; the comment says so. If the import order ever moves it first, the test still passes (zero tables either way) but stops exercising the "earlier imports rolled back" half — a reader changing the order should re-pick the table. Not a change for this PR.
- `configure_for_bulk_io` runs before the transaction opens, so its `SET` statements are outside it; that is the right order (session settings are not transactional) and worth keeping if the function is reordered later.

### Positive Notes

- The comment above the transaction states the cost model it removes (forty to seventy fsync-class commits per trace) in plain terms, so the next reader knows why the statements are grouped.
- The test asserts the observable property — no table survives — rather than counting rollbacks, and checks that the error still names the failing table.

### Runtime evidence

`cargo fmt --all -- --check` is clean on the tree. The transaction semantics the change rests on were observed on DuckDB 1.5.5 through the Python package (DDL plus an insert inside one explicit transaction, a later read of a non-parquet file failing, `ROLLBACK`, zero tables on reopen; the committed form surviving a reopen). The crate's own clippy and test run did not run before the push — no cargo build resolves on the host this was written on — so this PR's CI rows are the gate, and the PR stays a draft until they are green at the head.

josefbacik:
Rebased onto `main` at v1.19.0 (`3ab6068a`) after #259 merged; one commit, `fc117f8e`, the change unchanged in substance.

The one conflict was in `src/duckdb.rs`: #259 split `parquet_to_duckdb_with_options` into a caller (opens the connection, removes the partial database and its WAL on a failed import) and `import_directory`, the body that creates the schema, writes the metadata rows, reads the recorder's manifest, imports the tables and records the manifest on `_traces`. The resolution keeps that shape whole and moves the transaction inside `import_directory`: `configure_for_bulk_io` before the transaction as before, then `conn.transaction()`, `create_schema`, the two metadata rows, `read_manifest`, `note_unknown_files`, `import_tables` and `record_manifest` on the transaction, `tx.commit()` before `Ok(report)`. `import_directory` takes `&mut Connection` (the transaction needs it) and the caller passes `&mut conn`. On a failed import the guard rolls back on drop and the caller's removal of the file runs as on `main`, so the new test's assertion (no table in `main` after a failed import) holds by both mechanisms; its comment says so. The code is otherwise byte-identical to the previous head; rustfmt is clean.

PR #262, opened 2026-09-14
analyze: state the histogram's rounding and clamp the chunk count to its buckets

josefbacik:
## Summary

Two follow-ups to the histogram-sized `sched aggregate` guards (#257),
neither changing a number the fold reports.

## The change

- **The bound's rounding, stated.** The bucket boundaries are floored
  fractions of the window (`B_k = start + floor(span * k / n)`), while the
  histogram query cuts a slice start's bucket at the exact fraction. So a
  slice start sitting exactly on a boundary whose fraction is inexact lies
  below the exact cut and is counted one bucket LOW — it lands in the
  running total of a window that ends at that boundary although the
  slice ends after it. That is an over-count of the shorter window's
  rows, never an under-count: the budget cut still never runs over the
  budget, it only errs toward keeping a few rows fewer. The docs claimed
  an exact bound; the query builder's comment and the module overview now
  say "an upper bound, exact up to that rounding" and name the one case
  (an integer timestamp can sit in `[B_k, span·k/n)` only at `B_k`
  itself).
- **The chunk count clamped to the buckets.** The chunk count is derived
  from the row count (`rows / --chunk-rows`, rounded up), so a small
  `--chunk-rows` on a large stream asked for millions of chunks, and
  `chunk_edges_from_hist` built the edge list at that size before
  `stream_chunks` collapsed the coinciding edges: two vectors of the
  chunk count and a sort of the same length, transient, on a path that
  cannot resolve more than one edge per bucket boundary anyway. The count
  is now clamped to the bucket count (4096) before the edges are built,
  which yields the same chunks as asking for one chunk per bucket; the
  `--chunk-rows` docs say what a value below a bucket's worth of rows
  buys (chunks of about a bucket's events).

## What does not change

Every input the previous code handled cuts and chunks as before: the
clamp is reached only when the requested chunk count exceeds the bucket
count, where the histogram's resolution already bounded the chunking.
`SchedAggregateMeta`'s JSON is untouched — no new field — and the
`--chunk-rows` / `--max-rows` defaults are unchanged.

## Tests

- `histogram_boundaries_and_edges`, extended: a request of `u32::MAX`
  chunks on a dense histogram yields exactly one edge per bucket boundary,
  the same list as a request of one chunk per bucket, and those edges tile
  the window into as many chunks as there are buckets; the existing
  boundary, running-total, burst and empty-stream cases unchanged.

## Runtime evidence

Test Suite rows are green at the head.


josefbacik (commented):
## Code review

v1.19.0): one Rust file, one commit, `src/analyze/sched_aggregate.rs`
(+48/−16, docs, one clamp and one test); no BPF, no schema, no
output-shape change.

### Summary

Two follow-ups to the histogram-sized `sched aggregate` guards. The docs
claimed the histogram's running total is an exact bound of the rows a
window ending at each bucket boundary holds; the boundaries are floored
fractions of the window, so a slice start sitting exactly on a boundary
whose fraction is inexact is counted one bucket low — an over-count of
the shorter window's rows, never an under-count — and the docs now state
the bound as an upper bound exact up to that rounding, in the query
builder's comment and the module overview. The chunk count is derived
from the row count, so a small `--chunk-rows` on a large stream asked for
millions of chunks and the edge list was allocated and sorted at that
size before the coinciding edges collapsed (two vectors of the chunk
count and a sort of the same length, transient); the histogram cannot
resolve more than one edge per bucket boundary, so the count is clamped
to the bucket count before the edges are built, which yields the same
chunks as one chunk per bucket. No number the fold reports changes on any
input the previous code handled.

### Critical Issues

None.

  changes); no `SCHEMA_VERSION` bump, `SCHEMA_CHANGES.md` entry or
  `[bump minor]` token required. `SchedAggregateMeta`'s serde output is
  unchanged.
- BPF: none. ELF parsing: none. Unsafe: none.

### Improvements

- The clamp changes the chunk sizes only when `chunk_rows` is below the
  stream's rows per bucket (the count requested exceeds the buckets):
  there the previous code also could not honour the request — every
  bucket boundary with events became an edge — and the clamped form
  yields the same chunks whenever every non-empty bucket holds at least
  one mark's worth of rows (the stream's rows over the bucket count), and
  slightly coarser ones otherwise (a bucket under a mark merged with a
  neighbour; each chunk still within one mark's rows plus one bucket's).
  The parameter docs state the floor. Nothing further owed.
- The rounding sentence names exactly the case that differs: an integer
  timestamp can sit in `[B_k, span·k/n)` only at `B_k` itself, so the
  over-count is at most the slice starts on that one instant per
  boundary.

### Suggestions

- The clamp lives in `chunk_edges_from_hist`, the function that
  allocates by `n`, rather than at its caller, so the pure-function test
  covers it without a database fixture. Fine as written.

### Positive Notes

- The test pins both halves of the clamp's contract: a request of
  `u32::MAX` chunks on a dense histogram yields exactly one edge per
  bucket boundary and the same list as a request of one chunk per bucket,
  and those edges tile the window into as many chunks as there are
  buckets.

Test Suite, which runs the extended `histogram_boundaries_and_edges` test
beside the unchanged chunked-versus-single-query and truncation tests —
are the gate; the PR stays a draft until they are green.

PR #263, opened 2026-09-14
analyze: read digit separators in a real's exponent ahead of an escape-string prefix

josefbacik:
## Why

The escape-string prefix rule in `src/analyze/query.rs` decides whether an `E` before a `'` begins a token of its own by reading the identifier-character run that ends at the `E` the way DuckDB's lexer tokenizes it: leading digits are a number, an exponent when `[eE]` is followed by digits, the rest an identifier. #260 taught the integer part digit separators (`1_0E'`), but the exponent still took plain digits only, while the lexer admits separators there too (`real` is `({integer}|{decimal})[Ee][-+]?{integer}`, and `integer` is `{digit}("_"?{digit})*`): `1e1_0` is one number to DuckDB — observed on DuckDB 1.5.5 through the Python package, `SELECT 1e1_0` yields `1e10` — so `1e1_0E'x'` is the real `1e1_0` and an escape string to the lexer, while the classifier read the `E` as part of an identifier and the string as an ordinary one.

That is the same root as the `1_0E'` case: a walker whose escape string starts one character late reads a `\'` as the string's end and everything after it as code. Nothing reaches `prepare` through it — the grammar rejects a number followed directly by a string, so `duckdb_extract_statements` fails the whole text (`SELECT 1e1_0E'x'` is a syntax error at the string, observed) — but the prefix rule should match the lexer on its own rather than lean on that.

## What

- The exponent's digits are read with the same separator-aware helper as the integer part (`integer_literal_len`), so `1e1_0E'` opens an escape string as the lexer reads it; a doubled or trailing `_` still ends the number, and the rest of the run is an identifier, as before. A sign between the `e` and the digits is not an identifier character, so a run never contains one and the `[-+]?` branch needs nothing.
- A new test hands text to the pinned DuckDB with no classifier in front (`AnalyzeDb::query_strings`) and pins the lexer facts the refusals in `statement_shape` rest on, which the refusals themselves never let reach the engine: two `'` literals join across a newline (`\n` and `\r`) in the first literal's state — an escape string's backslash rule carries into the second part, an ordinary string's does not; no join without a newline; a number followed by an escape string is rejected (`1 E'x'`, `1e1_0E'x'`, `1E'x'`); a `--` comment ends at `\r`. Until now the crate's own tests proved the refusals and not the premise, which had been observed only outside the crate.

One correction to #260's description: it said that before that change a `\r`-terminated line comment let DuckDB run the statement after it. It did not — the classifier had already replaced the comment through the next `\n` or the end of the text with a space, and DuckDB received that stripped text, so the swallowed statement was silently dropped: a fidelity defect, not an exposure. The change itself was right; the description overstated it.

## Tests

- `opens_escape_string`: `1e1_0E'`, `1_0e1_0E'` and `5_0e1_0E'` open an escape string; `1e1__0E'` and `1e1_E'` do not. `statement_shape` reads `SELECT 1e1_0E'\'') AS q LIMIT 1; SET memory_limit = '100GB'; SELECT 1 FROM (SELECT 1 --'` as several statements, and through `AnalyzeDb::query` that text is refused as several statements with `memory_limit` unchanged.
- `test_query_strings_pins_the_lexer_facts_the_classifier_rests_on`, as described above.

## Runtime evidence

`cargo fmt --all -- --check` is clean. The classifier's unit tests (the ten `statement_shape` tests, the new assertions included) were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs`, 10 passed, and the same file is clean under clippy's default lint set with `-D warnings`. Every DuckDB behaviour the new binding test asserts was observed on DuckDB 1.5.5 through the Python package first (`1e1_0` as one number, the joins and the carried escape mode, the rejected number-then-string forms, the `\r` comment end). The crate's clippy and full test suite — the two binding tests among them — were NOT run before this push: no cargo build resolves on the host this was written on, so this PR's own CI rows are that proof, and the PR stays a draft until they are green at the head.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session by the author over the full diff of this PR's head at the time of the run (one commit on v1.18.15: `src/analyze/query.rs`, `src/analyze/mod.rs`; no BPF file, no schema, no `Cargo.*`); verdict as it produced it.

### Summary

A one-line alignment of the escape-string prefix rule with the lexer's numeric grammar (digit separators in a real's exponent), pinned at the classifier, plus a binding test that reads the pinned DuckDB's own lexer for the facts the classifier's refusals rest on. The code change is minimal and reuses the helper the integer part already uses; the new test is the part that earns its keep, since before it the crate proved its refusals and not their premise.

### Critical Issues

None.

### Improvements

- `integer_literal_len` on the exponent matches `real`'s `{integer}` exactly, including the doubled-`_` and trailing-`_` ends; the tests pin both. The `[-+]?` branch is handled by construction (a sign is not an identifier character), and the comment says so. No change requested.
- The new test compares a `DOUBLE` through a `CAST(... AS BIGINT)` rather than through the double's string form, which keeps it independent of float formatting. Good.

### Suggestions

- The binding test is version-pinned by nature: if a future DuckDB changes the continuation rule or the number-then-string rejection, this test is what says so before the classifier's refusals go stale. Worth keeping in mind when the crate's DuckDB is bumped; not a change for this PR.

### Positive Notes

- The description corrects a claim in the previous change's description (the `\r` comment case) instead of leaving it; the correction is accurate — the stripped text is what the engine received, so the old behaviour dropped text rather than ran it.
- Every assertion the new test makes was observed on a DuckDB build before it was written, per the description.

### Runtime evidence

The classifier's unit tests were run standalone under `rustc --test` on the exact bytes of `src/analyze/query.rs` (10 passed) and the file is clean under clippy's default lint set with `-D warnings`; `cargo fmt --all -- --check` is clean on the tree; the DuckDB facts the binding test asserts were observed on DuckDB 1.5.5 through the Python package. The crate's own clippy and test run (the two binding tests included) did not run before the push — no cargo build resolves on the host this was written on — so this PR's CI rows are the gate, and the PR stays a draft until they are green at the head.

PR #264, opened 2026-09-14
parquet: refuse appends on a stopped lane, close row groups at the bound, size the buffers for the crossing append

josefbacik:
## Summary

Follow-ups on the encode lane from the reviews of #248 / #256 (the packet consumer) and #255 (the sched consumer). Seven small corrections to the lane and its two writers' bookkeeping; a successful capture writes the same tables, schema and encodings as before, and the one file-level change is that a row group now holds exactly `max_row_group_size` rows.

## The changes

1. **A stopped lane refuses every later append** (`src/parquet/lane.rs`, `src/parquet/writer.rs`). `EncodeLane::stopped` reads true once a push found the lane thread gone, and as soon as that thread has exited on its own (an encoder, builder or sink error ends it before any push notices); `stop_error` returns the lane's own outcome. The writer asks before it buffers rows for any of the three lane tables (`network_packet`, `sched_slice`, `thread_state`), so an append after a lane death fails at once with that error. Until now a stopped lane failed only the append that crossed the flush bound: the batches before it were accepted, counted as recorded, and dropped whole with the buffer at the failed push — an id gap of one buffer per cycle in the packet file, a failure count off by the batches per cycle, and an exit summary that counted rows the file never got. This also makes #256's description true as written: "a lane that has stopped fails every one of them".
2. **Row groups close at the bound** (`src/parquet/lane.rs`). The lane closed a row group after the batch that reached `max_row_group_size`, so a group held between the bound and the bound plus one lane batch — up to about 1,210,000 rows at the default sizes (#256's description noted the looseness but put the figure at 1,020,000, a batch-size slip). The batch that would carry the open group past the bound is now split there, as `ArrowWriter::write` split it, and the rest opens the next group. Readers are unaffected either way; the two PR descriptions' "the same 1M bound" holds now.
3. **Failures counted in records** (`src/parquet/writer.rs`). `add_sched_batch` counted one failure per dropped batch where the trait's default and the five small tables count per record, so a shard's bail said "1 record" for a buffer of up to 210,000. It now counts every row of a refused batch and every row the lane dropped with a failed flush — the whole buffer, this batch and what earlier ones had left in it.
4. **No regrow on the crossing append, from the first cycle** (`src/parquet/writer.rs`, `src/systing_core.rs`, `src/sched.rs`). The packet buffer's headroom was the largest batch seen, which left the first crossing append one regrow short when the first batch was smaller than the later ones (the consumer's first `recv` plus whatever `try_iter` found). The headroom is now the consumers' own batch bound, `CONSUME_BATCH`, read from where the consumers are written. The two sched buffers had no headroom at all: they were swapped in at `batch_size` while a shard's flush carries them up to `STREAMING_SCHED_FLUSH_THRESHOLD` past it, so nearly every crossing append reallocated — a copy of the whole buffer under the shared writer lock every other shard waits on. They are sized for the bound plus one shard flush on the reserve and on the fresh buffer after each flush.
5. **A panic's message survives** (`src/parquet/lane.rs`). A lane-thread, builder or encoder panic was reported as fixed text; the payload's message is carried into the outcome. `build_parallel` joins every builder before it returns the first error, so a builder panic reaches the outcome as its own message rather than as the scope's.
6. **`total_records` after the flush** (`src/parquet/writer.rs`). A batch is counted after the flush it triggered succeeded, not before, on the three lane tables' batch and per-record appends.
7. **The collector resolved before the ids** (`src/network_recorder.rs`). `append_packet_batch` took the collector after the id loop, the one early return that could leave ids consumed; it is taken first.

## Test plan

- `lane::tests`: `round_trip_in_order_with_row_groups` now asserts the row groups' row counts (250, 250, 200); new `a_batch_past_the_bound_is_split_into_exact_row_groups` (100, 100, 100, 100, 20), `a_dead_lane_reads_stopped_and_names_its_error`, `a_builder_panic_carries_its_message` (the inline and the scoped builder shape).
- `packet_lane_tests`: new `a_small_first_batch_costs_no_regrow_on_the_first_crossing` and `a_stopped_packet_lane_refuses_every_later_append` (a lane made to die, then five batches and one record, every append refused with the lane's error, nothing buffered, nothing counted).
- `sched_lane_tests`: new `a_stopped_sched_lane_refuses_its_batches_and_counts_the_rows` (the refused batch counted as 10 records while the other table's rows land) and `a_million_lane_written_sched_rows_read_back_through_duckdb` (1,000,001 rows through the production writer properties: row groups of 1,000,000 and 1, every row back through `parquet_to_duckdb`).
- `cargo fmt --check` passes; clippy and the tests did not run before the push (no local build for this checkout) — the repository's Clippy / Test Suite rows are the gate and the PR stays a draft until they are green. The million-row test's wall time in the test row is read on the draft; if it runs past the suite's tolerance it gets `#[ignore]` with a smaller non-ignored twin.
- No schema, version or BPF change; no new dependency.


josefbacik:
## rust-bpf-code-reviewer verdict — `parquet: refuse appends on a stopped lane, close row groups at the bound, size the buffers for the crossing append` (head `1b7808dcf3c4eba49a5cb897e41d5380c5d17f38`)

Mechanics: the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition executed in-session by the author over the full diff against `v1.19.0` (`3ab6068a`; five files, +641/−94), in the definition's five passes. No BPF source is touched and no schema, table definition, `SCHEMA_VERSION`, `Cargo.toml` or `Cargo.lock` changes, so the BPF and schema-and-versioning passes have nothing to flag; the Rust-quality and integration passes are the substance and were read against the lane (`EncodeLane::push` / `finish` / `join`, `run_lane`, `RowGroup`), the writer's batch appends (`add_network_packet_batch`, `add_sched_batch`, the per-record `add_*` on the three lane tables), the sched shard's `flush_pending` bail, the packet consumers' failure report in `systing_core.rs`, and the collector trait's `add_sched_batch` contract. Each claim below names the line it was read at; the tests did not run before the push (no local build for this checkout), so every "the test pins" sentence is a reading of the test's code, and the CI rows are the gate.

### Summary

Seven small corrections on the encode lane and its two writers' bookkeeping, each local and explained where it sits, plus seven tests. A successful capture writes the same tables, schema and encodings as before; the one file-level change is that a row group now holds exactly `max_row_group_size` rows where it held up to one lane batch more.

### Critical Issues

None.

- `Cargo.toml` / `Cargo.lock` untouched; no schema file, no `SCHEMA_CHANGES.md` entry owed, no `[bump minor]` owed.
- No BPF source; no `unsafe` added.
- Recorder consistency: no recorder table added or changed; `utid` attribution untouched.

### Improvements

None owed for this change. Two readings the reviewer checked because they are where a change of this shape goes wrong:

- **The refusal cannot deadlock or double-join.** `refuse_if_stopped` (`writer.rs:364`) asks `stopped()` (`lane.rs:148`: `tx.is_none() || thread.is_finished()`, an atomic read, no join) and calls `stop_error()` only on a stopped lane; `stop_error` sets `tx = None` and calls `join()`, which takes the `JoinHandle` once and repeats the cached outcome afterwards (`lane.rs:174–190`), so a writer that refuses five appends joins once and reports the same error five times. `push` now ends in `stop_error()` on a failed send (`lane.rs:139`) — the same text it produced before.
- **`is_finished()` is the right early read.** A lane thread that returned an error has dropped its receiver, so a later `push` would fail anyway; asking the handle first means the refusal lands on the first append after the death rather than on the next crossing flush, which is the whole point of the change. A lane that is merely slow reads live (its thread runs), so backpressure is unchanged.

### Suggestions

- `run_lane`'s split loop takes `room.max(1)` (`lane.rs:268`) so a group that somehow reads full at the top of the loop still advances; with the close at `current.rows >= row_group_rows` right after each write that state is unreachable, and the `max(1)` is a guard against a zero `row_group_rows`, which `start` already clamps to 1. Harmless; noted so the next reader does not look for the case.
- `a_million_lane_written_sched_rows_read_back_through_duckdb` builds 1,000,001 records, encodes six lane batches and runs the DuckDB import in one test. Its wall time in the CI test row is a datum to read on the draft; if it runs past the suite's tolerance it should be `#[ignore]`d with a smaller non-ignored twin at a lowered bound, as the sched-lane tests already do for the file contract.
- `assert_eq!(capacity, 25 + PACKET_BATCH_HEADROOM)` in `a_small_first_batch_costs_no_regrow_on_the_first_crossing` rests on `Vec::reserve` on an empty vector allocating exactly the request (the standard library does that today and documents "at least"), the same reliance the earlier packet-buffer test carried; left exact so a doubling is caught.

### Positive Notes

- The stopped-lane refusal is one helper used at the top of the five appends that touch a lane (`writer.rs:1410, 1424, 1453, 1478, 1699, 1716`), and the writer's own tests make a lane die (a builder that fails) rather than relying on a sink that only fails at close — the death is deterministic and the refusal is asserted five times over, on the packet lane and on one of the two sched lanes while the other keeps landing rows.
- `add_sched_batch` now counts records on every path (`writer.rs:1447–1450`: the refused batch's rows, the whole buffer a failed flush dropped, one per small-table record), which is what the trait's default returns and what `sched.rs:380`'s "failed to stream {failed} sched/IRQ records" prints; `total_records` is added after the flush on all three lane tables and in the per-record adds.
- The row-group split (`lane.rs:255–284`) slices the arrow batch at the remaining room — a zero-copy `RecordBatch::slice` — and keeps the error path exactly as it was (the encoder's own error through `abort`); the two lane tests read the row-group row counts back from the file metadata, and the million-row writer test reads `[1_000_000, 1]` through the production properties.
- The headroom constants are the producers' own bounds — `CONSUME_BATCH` (`systing_core.rs:1921`) and `STREAMING_SCHED_FLUSH_THRESHOLD` (`sched.rs:27`), both now `pub(crate)` with their doc saying who reads them — so there is one number per bound and the observed-max field is gone.
- The panic text helper (`lane.rs:195–203`) reads `&str` and `String` payloads and says so otherwise; `build_parallel` joins every builder before returning the first error (`lane.rs:310–323`), which is what keeps a scoped-thread panic from turning into the scope's own panic and losing the message — the test runs both the inline and the scoped shape.
- `append_packet_batch` resolves the collector before the id loop (`network_recorder.rs:1124–1129`): the doc comment's "a batch that fails consumes no ids" now holds on every return path.

PR #265, opened 2026-09-15
pystacks: shape a recovered qualname and filename the way BPF hands them over

josefbacik:
## What

The pystacks symbol resolver's page-fault recovery path handed a recovered qualname over as the whole window of process memory it read, and `get_symbol_name` rendered the bytes after the string's NUL as a function name: `bytes_print.<non-printable bytes>`, `_worker.<non-printable bytes>`, cached under the symbol's id for the rest of the capture. This shapes a recovered qualname (and filename) the way the BPF read hands them over.

## Why

`get_names()` in `pystacks.bpf.c` copies `co_qualname` into the first 128-byte partition of the symbol's qualname buffer as a NUL-terminated string and then clears the second partition, so user space renders the first partition alone (`get_symbol_name`: the second partition is a function name only when it is non-empty). When that read faults in the sampler's context, BPF records the fault address and `ingest_record` recovers the string from `/proc/<pid>/mem` — but it read all 224 bytes and stored the buffer verbatim, so the heap bytes after the string landed in the second partition. Every recovered symbol rendered as `<qualname>.<garbage>` and, because `ingest_record` caches a symbol by id on first ingest, stayed that way for the capture. On a large CPython 3.13 estate such names sat on roughly a tenth to a fifth of the Python-bearing samples (the recovery path runs on every faulted read, which is common on big cold heaps).

The filename recovery had the sibling flaw for long paths: BPF keeps the tail of a path longer than the 192-byte field (the module name lives at the end), the recovery read kept the head.

## How

- `shape_recovered_qualname`: keep the string up to its NUL, cut at the partition boundary as BPF does, zero the rest.
- `shape_recovered_filename`: read a 4096-byte window instead of 192 bytes and keep the last `BPF_LIB_PYSTACKS_FILE_NAME_LEN - 1` bytes of the string (one fewer than the field, as BPF's NUL-terminating copy leaves).
- Three unit tests with heap-shaped buffers (`0xa5` fill): the recovered qualname renders as the string alone (with a negative control asserting the unshaped buffer did not), a qualname longer than the partition is cut where BPF cuts it, and a long path keeps its tail so the module name survives.

No BPF change, no schema change. Frame strings for symbols the BPF read resolved are unchanged; only recovered names lose the appended bytes.

## Review

The rust-bpf-code-reviewer pass, executed in-session by the author over the full diff, is posted as a review comment on this PR: no Critical issues; one Improvement (the discarded-`Result` map shape) taken before the commit. `cargo fmt --check` clean; clippy and tests in CI.


josefbacik (commented):

Executed in-session by the author over the full diff (`src/pystacks/symbols.rs`, 1 file, +118/−7) following the repository's `rust-bpf-code-reviewer` agent definition; the review passes (structure, Rust quality, BPF quality, schema & versioning, integration) applied in order. Local gates: `cargo fmt --check` clean on the changed file; `cargo clippy` and `cargo test` run in CI on this PR (the author's host cannot resolve the crate graph at the moment).

### Summary
A small, well-scoped correctness fix in the user-space half of the pystacks symbol resolver: the page-fault recovery path stored the whole window of process memory it read as the symbol's qualname, so a recovered name rendered as `<qualname>.<heap bytes>` through `get_symbol_name`'s two-partition rule and was cached that way for the capture. The change shapes the recovered qualname exactly as the BPF read hands one over (string to NUL, cut at the first partition, the rest zero) and makes the filename recovery keep the tail of a long path as BPF does. Three unit tests pin the behaviour, one of them asserting the pre-fix rendering was wrong.

### Critical Issues
None. No BPF file, schema, `SCHEMA_VERSION`, `Cargo.toml` or `Cargo.lock` change; no `unsafe`; no hand-rolled ELF parsing; no recorder table.

### Improvements
- The filename recovery's `let _ = read(...).map(|n| { side effect })` shape was replaced with `if let Ok(n) = read(...) { ... }` before the commit — a discarded `Result` whose `map` closure carries the only effect reads as an accident; taken before commit.

### Suggestions
- `FILENAME_RECOVERY_WINDOW` (4096) is a `PATH_MAX`-sized read of `/proc/<pid>/mem` per recovered filename; recoveries happen once per symbol per capture, so the cost is negligible, and a short read across an unmapped page is handled through the returned length. A future change could size it from `PATH_MAX` explicitly if the crate ever gains that constant; not needed now.
- `shape_recovered_filename` keeps `BPF_LIB_PYSTACKS_FILE_NAME_LEN - 1` bytes because the BPF side's NUL-terminating copy holds one byte fewer than the field; the comment says so. If the BPF copy ever changes to keep the full field, this constant follows it — the test `test_recovered_filename_keeps_the_tail_of_a_long_path` asserts the exact length and would catch the drift.

### Positive Notes
- The helpers are pure functions over byte slices, so the tests exercise the real code path with heap-shaped buffers (`0xa5` fill) rather than a mocked process, and the negative control (`assert_ne!` on the unshaped buffer) documents the bug the fix removes.
- The doc comments explain *why* the shaping exists (the two-partition contract `get_symbol_name` relies on) by naming the BPF function, not by line numbers.
- No frame string changes for symbols the BPF read resolved; only recovered names lose their appended bytes — a bounded blast radius stated in the commit message.

Verdict: APPROVE — no Critical issues; the one Improvement was taken before the commit.

PR #266, opened 2026-09-15
memory recorder: attach each syscall hook's exit before its enter; tag scratch entries by leg

josefbacik:
## Summary

The memory recorder pairs each mmap/munmap/brk syscall's enter and exit through one per-thread scratch map shared by the three legs (`memory_syscall_scratch`, keyed by tgidpid). Two defects in that pairing, one of them visible in recorded data:

1. **Attach order.** The six classic programs (and the six trampoline ones) attached enter-before-exit, one program at a time (`attach_program_set` walks the list by name; each attach is a perf_event_open + ioctl round trip, and the first syscall tracepoint also walks every task to flag it, so consecutive attaches are 100 µs to milliseconds apart on a busy host). A syscall that entered after its enter program was live and returned before its exit program was live left a scratch entry nobody consumed; the thread's next brk, if its own enter fell before `sys_enter_brk` was live and its exit after `sys_exit_brk` was, paired with that entry and recorded `new_brk − <the munmap's address>` as the brk delta: page-aligned negative deltas of 13–46 TB (the mmap region minus the PIE heap) under an ordinary glibc `free → systrim → brk` stack — a handful of rows a week across thousands of hosts. Each set now lists the exit before the enter (`MEMORY_SYSCALL_TP_PROGS`, `MEMORY_SYSCALL_FENTRY_PROGS`; the raw pair kept uniform), so a syscall spanning the attach gap can only exit without an enter, which the exit programs already ignore. Detach order needs no care: the map dies with the capture.
2. **Untagged entries.** The scratch entry carried no leg, so an exit paired with whatever it found. `memory_syscall_args` gains a `leg` tag set by each enter (`enum memory_syscall_leg`); `memory_syscall_scratch_take()` drops an entry of another leg and counts it as a new missed-events class, `MISSED_MEMORY_XLEG_EVENT` (index 9, label `memory-xleg`): a counter track beside the memory loss track in the trace, and an exit-summary line printed unconditionally so a zero is a positive reading — `Memory syscall scratch cross-leg drops: N`. Expected 0 after (1); a count says an enter ran without its exit some other way and was already neutralized.

Also fixed on the way: `missed_events` was re-sized to the CPU count at open (`set_max_entries(num_cpus)`), which on a host with fewer CPUs than classes left the high classes unallocated — their lookups returned NULL in the programs and 0 in the summary, so a 4-CPU machine reported "Missed memory events: 0" whatever it lost. The map keeps its declared size now (a per-CPU array already holds one value per CPU behind each slot); the load-shape probe does the same.

No schema change: no table or column changes (the counter track is one more class in the existing missed-events tracks; the `SCHEMA_VERSION` is untouched). No behaviour change for a correctly paired syscall; the `−EINTR` and `old_brk == 0` guards stay.

## How it was found

Page-aligned negative brk rows of 7–46 TB, about one a day across many memory-recorded hosts, survived the earlier brk fixes (#199's address-as-delta guard, #226's `−EINTR` guard). The rows with a live `libc.so.6:brk` leaf carry `brk ← __default_morecore ← (systrim) ← free ← <the program's allocator>` — the same frames as the ordinary −100 MB … −1 GB trims recorded from the same processes. A glibc `systrim` trims the free top chunk; a 30–41 TB top chunk does not exist (the main arena grows below the mmap threshold per step, and TB requests go to mmap), so the new break was real and the previous break was not the heap: it was whatever the thread's scratch entry held, and a munmap's address (the mmap region, above the PIE heap) gives exactly a negative delta of that size and never a positive one. The source then shows the untagged shared entry and the enter-first attach order.

## Testing

- `cargo fmt --check`; `cargo clippy --all-targets -- -D warnings`; the lib suite, with `test_memory_syscall_program_sets_attach_exit_before_enter` added (every set lists each leg's exit before its enter, pairwise by name).
- The BPF object's load-shape gate in a vanilla 6.12.0 guest (`bpf_load_shapes -f every_shape_loads`): every shape loads with the changed scratch struct and the new class.
- Map/context pairing: `memory_syscall_scratch` (a preallocated HASH) is written only by the syscall-context programs; `missed_events` is a per-CPU array; nothing new is written from the `perf_event` program.
- Prepared on top of #247 (the opt-in raw-tracepoint form of the same hooks) and rebased onto `main` after that merge; the patch is unchanged.


josefbacik:

Reviewed as commit `5ae6008b` on its then-base `d9d545e5`; the same patch, byte-identical (`git range-diff` reads one `=` row), is `94e5ba2a` on `main` at v1.20.4 after #247 merged.

Executed in-session by the author over the full diff of commit `5ae6008b` against its base `d9d545e5` (the opt-in raw-tracepoint form of the same hooks): `src/bpf/systing_system.bpf.c` (+62/−4), `src/systing_core.rs` (+91/−22), `src/session_recorder.rs` (+4/−0). Five passes as the definition lays them out: structure, Rust quality, BPF quality, schema & versioning, integration.

### Summary

A small, well-bounded correctness fix to the memory recorder's syscall pairing: each leg's exit program now attaches before its enter (closing the attach-gap window that created stale scratch entries), the scratch entry carries the leg that wrote it (so an exit never pairs with another leg's entry), and the cross-leg drop is counted as a missed-events class. The change also removes a latent sizing bug that left high missed-events classes unallocated on small hosts. No schema change; no behaviour change for a correctly paired syscall. Nothing blocks.

### Critical Issues

None.

- Schema & versioning: no table, column or `SCHEMA_VERSION` change; `Cargo.toml` untouched; `SCHEMA_CHANGES.md` needs no entry (the new counter track is one more class of the existing missed-events tracks, keyed by label).
- Recorder consistency: no new per-event table; the missed-events counter tracks carry no thread identity by design.
- BPF map/context pairing: `memory_syscall_scratch` (preallocated HASH) is written only from syscall-context programs; `missed_events` is a per-CPU array; nothing new is written from the `perf_event` program.

### Improvements

- `memory_syscall_scratch_take()` returns the map value pointer and leaves the copy-then-delete to the callers, as the three exit commons already did. A version that copies into a caller-provided struct and deletes inside the helper would remove the repeated `struct memory_syscall_args args = *saved; bpf_map_delete_elem(...)` pair from the three callers, at the cost of a memcpy the mismatch path does not need. Optional; the current shape keeps the diff minimal and the verifier's view unchanged.

### Suggestions

- The exit-summary line for the new class prints under `opts.memory` while the `Missed memory events` line prints under `opts.memory || opts.memory_alloc`. Correct — the syscall legs exist only under `opts.memory` — but a one-line comment at the print site would spare the next reader the question.
- A deterministic e2e of the cross-leg drop (a thread that munmaps then holds a brk across the capture's attach) is not reliably reproducible in the VM test runner; the attach-order unit test plus the always-printed counter line are the practical guards, which the change provides. Nothing further asked.

### Positive Notes

- The attach-order invariant is pinned by `test_memory_syscall_program_sets_attach_exit_before_enter`, pairwise by name over all three sets, so a future reorder fails a test rather than silently reopening the window.
- The BPF change is three tagged initializers and one small `__always_inline` helper replacing three bare lookups — the deduplication the definition asks for — with a compare on a `u32` and no new map or ringbuf data; the release-compiler object shows the expected shape in `sys_exit_brk` (lookup → NULL check → leg compare → delete + counter on mismatch), and every load shape still loads on the 6.12.0 guest.
- The `missed_events` sizing fix is the right reading of `BPF_MAP_TYPE_PERCPU_ARRAY` (`max_entries` counts keys; each key already holds one value per possible CPU) and is applied at both open sites so the load-shape probe still mirrors the real open path.
- Comments explain why (the attach gap, the counter's meaning), name functions and fields rather than line numbers, and the doc comments on the three program-set constants carry the invariant where the next editor will see it.

### Verified locally
- `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean; the lib suite 591 passed / 0 failed (the new test included).
- The BPF object's load-shape gate in a vanilla 6.12.0 guest (`bpf_load_shapes -f every_shape_loads`, built with the release compiler): 35 of 35 shapes load, `1 passed; 0 failed`.

PR #267, opened 2026-09-15
duckdb: read the recorder's manifest before the import transaction

josefbacik:
## Why

`import_directory` reads the recorder's manifest (`systing_manifest.parquet`) inside the one transaction that builds the trace database (#261). `read_manifest`'s contract is that a manifest which is present but unreadable is provenance that must not cost the tables: the error is reported on stderr and the manifest reads `None`. Inside an explicit transaction that contract no longer holds. DuckDB invalidates a transaction on the reader's error class — an invalid-input or IO error, unlike a binder, catalog, parser or permission error — so once `read_parquet` has failed on the damaged file, every later statement in the transaction fails with `Current transaction is aborted (please ROLLBACK)`, `import_tables` fails at its first statement, the import fails whole, and `parquet_to_duckdb_with_options` removes the database it created. A damaged manifest — the shape a writer killed between the tables and `finish()` leaves behind — turned into a lost trace database, where in autocommit the same directory imported with the warning.

## What

The manifest is read before the transaction begins, after `configure_for_bulk_io`: one read-only statement in autocommit, on a file at most a few hundred bytes long. The guard's warnings still name the writer (the manifest is read first, as before), and the manifest's values are still written to the `_traces` row inside the transaction. Nothing else moves: the schema, the metadata rows and every table import stay inside the one transaction, and the caller's removal of a refused import's database is unchanged.

## Tests

- `test_import_keeps_the_tables_when_the_manifest_is_damaged` (new): a valid `cpu_info.parquet` beside a `systing_manifest.parquet` of non-parquet bytes. The import succeeds, `recorder` reads `None`, the `_traces` row carries no recorder fields, and the table's row is there. At the parent commit this test fails with the transaction error above.
- `test_parquet_to_duckdb_failed_import_leaves_no_half_built_database` now reads the rollback itself. Before, it reopened a path the caller had already removed, which finds an empty database whether or not anything was committed. It now pins the removal of the file and its write-ahead log on the caller's path, then calls `import_directory` directly against an open connection and counts `information_schema.tables` on that connection after the error: 0, because the transaction rolled back.

## Runtime evidence

`rustfmt --check` (edition 2021) is clean on the file. The transaction's behaviour was observed on DuckDB 1.5.5 through the Python package (the crate pins libduckdb 1.5.4): in autocommit a failed `read_parquet` on a file of non-parquet bytes is its own statement and the next statements run; inside `BEGIN … COMMIT` the same failure (`No magic bytes found at end of file`) is followed by `Current transaction is aborted (please ROLLBACK)` on every later statement and the table created before it does not exist afterwards; the read moved before `BEGIN`, the transaction's imports succeed; a binder-class error (a missing column) inside a transaction does not abort it; a truncated parquet aborts it the same way. The crate's clippy and test suite — the two tests among them — were NOT run before this push: no cargo build resolves on the host this was written on, so this PR's own CI rows are that proof, and the PR stays a draft until they are green at the head.


josefbacik:
Code review — the repository's `rust-bpf-code-reviewer` definition executed in-session over the full diff of `ee89a3e7c34214c876eaa8bb6fd8c0f918df4f9c` (one file, `src/duckdb.rs`, +94/−11 on `7c32b51e`); verdict as it produced it. A first pass over the same change at its previous commit raised two comment-wording suggestions, both folded before this pass.

### Summary

A small, targeted fix: `import_directory` now reads the recorder's manifest before it opens the import transaction, so a manifest that is present but unreadable keeps the documented warn-and-continue contract instead of aborting the transaction and losing the database. The two tests pin the behaviour the change is about (a damaged manifest beside a valid table imports; a failed table import rolls back on the import's own connection). No schema, recorder, BPF or unsafe code is touched.

### Critical Issues

None.

### Improvements

None required. The move is the smallest change that restores the contract: the manifest read is a single read-only statement, so running it in autocommit costs no write-ahead-log write, and the manifest's values still reach the `_traces` row inside the transaction through `record_manifest`.

### Suggestions

- `test_parquet_to_duckdb_failed_import_leaves_no_half_built_database` now asserts the error text twice (once per path); a shared local closure would remove the duplication, at the cost of a few lines — a matter of taste, left as is.

### Positive Notes

- The rollback test now reads the rollback on the connection the import ran on, where the previous form reopened a path the caller had already removed and would have passed with or without the transaction.
- The damaged-manifest test names the failure shape it guards against and asserts both halves of the contract: the manifest reads as absent (`recorder` `None`, the `_traces` recorder fields `NULL`) and the table's data is present.
- The comment explains the DuckDB behaviour the placement depends on rather than the placement itself, so a later reader will not move the read back inside the transaction to "tidy it up".

PR #269, opened 2026-09-16
analyze: name the shape of the sched loss inside the folded window

josefbacik:
## What

`systing-analyze sched aggregate` reports the BPF side's dropped scheduler events as one number for the whole run (`meta.missed_sched_events`, the drop counter track's last value). A row that says "7 million events missed" cannot say whether the ring filled once for a second or steadily for the whole window, nor whether the recorder's own threads were starved of CPU while it happened — and those are different problems with different fixes (a ring-size or poll-latency question versus a scheduling one).

This adds eight `meta` fields that say so, read from data the trace already carries:

- `missed_in_window`, `missed_seconds`, `missed_burst_max`, `missed_span_ns` — the same drop counter track (the recorder's once-a-second poll of the BPF counter), fetched ordered by time and folded over the window the aggregate used: the drops inside the window, how many one-second samples carried drops, the largest single-sample delta, and the time from the first to the last dropping sample. `null` when the trace carries no track or no sample of it inside the window. A window that starts mid-run takes the last sample at or before its start as the baseline, so the run's earlier drops are not counted; a value below its predecessor is a counter reset and counts as no drop.
- `self_poller_max_wait_ns` / `self_poller_waits` and `self_consumer_max_wait_ns` / `self_consumer_waits` — the longest runnable wait the pass closed for one of the recorder's own ring pollers (the `events_<N>` threads) and sched consumers (`sched_rec_<N>`), and how many it closed. The recorder's threads are found once by name; utids are dense small integers, so each closed wait costs one indexed byte, no hashing. A kernel workqueue thread is `events/<N>`, with a slash, and matches neither.

Reading a row: most of `missed_in_window` in one sample with short poller waits is a burst inside the poll latency; steady deltas across the window are throughput; drops in the few seconds a poller waited longer than its ring takes to fill are the recorder starved of CPU, not its consumers' throughput.

## What does not change

- No BPF change and no new cost on the traced host: the counter track and the recorder's threads are already in every trace.
- `missed_sched_events` and every existing field keep their meaning and value; the workload's own distributions are untouched (the recorder's threads' waits stay in them as before).
- The fold's fast path: one extra ordered fetch of the counter track before the fold (a few hundred rows for a continuous capture) and one bounds-checked index per closed wait. Nothing else on the per-event path moves.
- The JSON shape gains eight keys and changes no existing value, so a consumer that pins the shape sees additions only.

## How it was verified

- Unit tests (four new): the fold over cumulative samples (a whole window, a mid-run window, a window with no sample, a window whose samples carry no drop, a counter reset); the thread-name match (`events_0`, `events_63`, `sched_rec_7` in; `events/3`, `events_`, `events_1x`, `kworker/0:1`, `sched_recorder` out, and a utid past the table is nobody's); the fields reaching the meta through a small in-memory trace with a drop counter track and one poller / one consumer wait, the workload's own statistics unchanged; and a trace with no counter track leaving the loss fields `null` and the self fields zero.
- First contact after a release: on a many-CPU host whose captures report `missed_sched_events > 0`, the four `missed_*` fields and the two `self_*` pairs in the aggregate's JSON, read beside the per-capture drop tally — the row's shape (burst / steady / starved) is the point of the change.

Base: `main` at v1.20.6 (`a1f76bc`). One file, `src/analyze/sched_aggregate.rs`.

## Risk review

Run on the final diff (one commit, one file). Each answer is tagged verified (read in the diff, the source or a CI log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** one, stated: the aggregate's JSON output gains eight `meta` keys — an add-only change to an output format, and a consumer that types those keys in the meantime reads them as absent again after a revert. The tool's own output reverts exactly (a revert stops producing the keys and changes nothing else; rows written with the keys stay readable, since readers of the aggregate ignore unknown keys) — verified in the diff; no data is deleted or migrated, no existing key is renamed or re-typed, and the add step is the reversible shape (add → readers adopt → nothing to remove) — verified. The consumer side is outside this repository (inferred).
- **Compat changes:** yes, everything existing behaves as before: `missed_sched_events` and every existing field keep their value, and the workload's own distributions are untouched — verified by the unit test that folds a small trace with recorder waits and asserts the workload's statistics unchanged; nothing on the traced host or in BPF moves — verified in the diff (one file, the analyzer). No default changes; there is no flag.
- **Tradeoffs:** one extra ordered fetch of the counter track per fold and one bounds-checked byte index per closed wait — verified in the diff; the cost on a real trace is a few hundred rows of fetch (inferred from the recorder's once-a-second poll, not measured). The recorder's threads are found by their names (`events_<N>`, `sched_rec_<N>`): a rename of those threads in the recorder would silently zero the `self_*` fields — inferred; the name match is pinned by a unit test so a rename fails a test before it fails a row.
- **Blast radius:** `systing-analyze sched aggregate` only — its JSON and table output; nothing that records, nothing in BPF, nothing on a traced host — verified in the diff.
- **Rollback:** revert the pull request; the next aggregate row carries the previous keys only — verified: no state carries across runs.
- **How verified:** the four unit tests and Clippy in CI's Test Suite (green at the head) and the review pass posted as a comment — verified in the CI runs; the drop-counter fold's baseline and reset rules by the tests' cases — verified.
- **Not checked:** a real trace with drops (the first aggregate over a capture that reports `missed_sched_events > 0` after a release is the check, its bar in "How it was verified"); the consumer that reads the new keys (outside this repository — assumed to ignore unknown keys as JSON readers do).


josefbacik:
### rust-bpf-code-reviewer verdict — `analyze: name the shape of the sched loss inside the folded window`

Executed in-session by the author over the full diff (one file, `src/analyze/sched_aggregate.rs`, +345 lines; the final diff's sha256 `69162d02147f6e40c6e6fd58a9bf81f9245a10fdaa870c091147351296b1ff65`) against the repository's `rust-bpf-code-reviewer` definition, all five passes; the one Improvement it raised was taken before the commit, so the diff reviewed is the diff pushed.

### Summary
Eight `meta` fields describe the shape of the dropped-event loss inside the folded window (from the drop counter track the trace already carries) and the recorder's own threads' worst runnable waits (from the fold's existing wait bookkeeping). No BPF change, no schema change, no per-event cost beyond one indexed byte per closed wait; the existing fields and the workload's distributions are unchanged. Straightforward and well-tested.

### Critical Issues
None. No BPF code; no DuckDB schema change (`SCHEMA_CHANGES.md` / `SCHEMA_VERSION` untouched by design — the aggregate's JSON is not the trace schema, and the change adds keys without changing a value); no `Cargo.toml` version edit; no recorder or Python-version change; no ELF parsing.

### Improvements
1. **(Taken before the commit)** The first draft looked each closed wait up in two `HashSet<i64>`s — two SipHash lookups on the fold's hot path, where the per-event cost is the fold's whole bound. utids are dense small integers, so `self_threads` now builds a `Vec<SelfThread>` indexed by utid (as long as the largest recorder utid; every utid past it is the workload's) and `record_wait` does one bounds-checked index — no hashing, no per-wait allocation.
2. `missed_in_window` swallows a `prepare` / `query` error as `Ok(None)`, mirroring the sibling `missed_sched_events` exactly; the table-existence check above it already covers the one expected failure. Consistent with the file; a future change that propagates both siblings' errors would be welcome but is not this PR's.

### Suggestions
- `missed_shape` charges a delta to its sample's own timestamp; the doc comment says so, and the mid-run baseline rule keeps a window from inheriting the run's earlier drops. Fine as written.
- The `(b - a).max(0)` in `span_ns` is defensive against an unordered fetch; the query orders by `ts`, so it never triggers. Harmless.

### Positive Notes
- The fold over cumulative samples is one linear pass with a clear baseline rule and an explicit counter-reset rule, and the unit test exercises all four cases (a whole window, a mid-run window, a window with no sample, a counter reset).
- The thread-name match is exact (`events_<digits>` / `sched_rec_<digits>`) and the negative cases in the test (`events/3`, `events_`, `events_1x`, `sched_recorder`) pin the boundary.
- The end-to-end test drives the fields through a small in-memory trace with the counter track and one poller / one consumer wait, and asserts the workload's own statistics are unchanged.
- Comments explain *why* (the baseline rule, the reset rule, the indexed table's bound) and reference functions and fields by name, never by line.

PR #270, opened 2026-09-16
symbolize: answer the pool-backed check in O(1) for processes without a pool

josefbacik:
## What

`ProcessMaps::is_pool_backed` decides, per sampled address, whether the address lies in one of gVisor's memory pools (`runsc-memory`, `systrap-memory`) — such an address is never handed to the symbolizer as a live process address, because its `map_files` link opens the pool memfd itself. It is asked once per distinct sampled address of every process, and it answered by walking the process's whole mapping list (`iter().find` over the entries), so the symbolization phase paid (addresses × mappings) per process, most of it on processes that have no pool mapping at all and can only ever answer "no".

Two changes, no change in any answer:

- `is_pool_backed` returns false at once when the process has no pool memfd mapping — the `gvisor_memfd` flag `parse` already sets — so every process outside a sandbox never touches its entries.
- The mapping lookup behind it, and behind `bridge_for`, `pool_image_base` and `label_for`, is a binary search (`partition_point`) over the address-ordered entries instead of a linear scan. `parse` now sorts the entries by start so the order holds even for hand-written maps text; the kernel already emits `/proc/<pid>/maps` in that order and mappings never overlap, so the order by start is total.

## What does not change

- Every answer: the pool-name test, the bridge's neighbour rule, the labels. The existing `test_is_pool_backed_names_the_pools_only` and bridge tests are untouched and keep the semantics.
- No BPF change, no schema change, no new dependency, nothing on the recording side.
- The sort is one stable `sort_by_key` per process at parse time (linear on the already-ordered kernel output), against a linear scan per sampled address it removes.

## How it was verified

- A new unit test parses three mappings written out of order, checks `parse` ordered them, and walks every boundary of the search — an entry's first and last byte, the gap between entries, below the first entry, past the last entry's end, `u64::MAX`, an empty map — then checks a process without a pool mapping (its own re-exec'd binary's memfd included) still answers `is_pool_backed` false for every address.
- First contact after a release: the symbolization phase's wall time (`symbolize and write parquet done in … ms`) at p50 / p90 on a host class with many mappings per process, read in an hour of comparable load before and after; the per-address cost this removes is largest where processes map the most files.

Base: `main` at v1.20.6 (`a1f76bc`). One file, `src/sandbox_maps.rs`, +70 / −11.


josefbacik:
### rust-bpf-code-reviewer verdict — `symbolize: answer the pool-backed check in O(1) for processes without a pool`

Executed in-session by the author over the full diff (one file, `src/sandbox_maps.rs`, +70 / −11; the final diff's sha256 `0c4126d86a4c377ee79d6359ab6d5ca76598d61adb319ddb5e4cd6375a095f2d`) against the repository's `rust-bpf-code-reviewer` definition, all five passes.

### Summary
A cost-only change with no change in any answer: `is_pool_backed` short-circuits on the `gvisor_memfd` flag `parse` already sets, and the mapping lookup behind it, `bridge_for`, `pool_image_base` and `label_for` becomes a binary search over entries that `parse` now keeps in address order. No BPF, no schema, no new dependency; the per-address symbolization path loses a linear scan per process.

### Critical Issues
None. No BPF code; no DuckDB schema change; no `Cargo.toml` version edit; no recorder, Python-version or ELF-parsing change.

### Improvements
None required. Two things checked and found sound: (1) the binary search's invariant — entries ordered by `start`, non-overlapping — holds for kernel-written maps by construction and is now enforced by the stable `sort_by_key` in `parse` for any text, at a one-time O(n log n) per process (O(n) on already-sorted input) against the per-address cost it removes; (2) `bridge_for` keeps its index arithmetic on neighbours (`idx − 1`, `idx + 1`), which the same address order makes correct.

### Suggestions
- `pool_candidates` / `virt_candidates` still filter the whole entry list, but they answer a per-offset question for the guest correlation, not the per-address hot path; left alone on purpose.

### Positive Notes
- `index_for` is four lines and reads exactly as the doc comment states it: the last entry starting at or below the address, if the address is below its end.
- The new test walks every boundary of the search (first byte, last byte, the gap, below the first, past the last, an empty map) on out-of-order text, so the sort and the search are pinned together, and checks that a process without a pool still answers false everywhere, its own re-exec'd binary's memfd included.
- The existing `test_is_pool_backed_names_the_pools_only` and bridge tests are untouched and keep the semantics.

PR #271, opened 2026-09-16
analyze: cut the sched aggregate's window to a wall-clock budget

josefbacik:
## What

`systing-analyze sched aggregate` bounds its fold with an event budget (`--max-rows`, a row count): a window holding more rows is shortened from its end to what fits, exactly, and the row says so. A row count is the wrong unit for a caller with a deadline. The fold's cost per row differs by host — twice over between two machines of one class is ordinary — so one row budget either wastes the deadline on fast hosts or misses it on slow ones, and a fold that misses its deadline leaves no row at all.

This adds `--time-budget <seconds>` (`SchedAggregateParams::time_budget`), a wall-clock budget for the whole call:

- The first chunk of a multi-chunk fold doubles as a probe of this host's fold rate. When the time it took, times the chunks still to fold, would run past the budget, the rows that rate folds in four fifths of what is left (`TIME_CUT_FILL`; the fifth held back covers the rate's error and the restart's own sizing queries) become an event budget, the window is cut to them by the same bucket-boundary cut the event budget uses, and the fold restarts over the shorter window.
- Every statistic stays exactly defined over a contiguous, un-sampled window, as with the event budget — the same cut, the same whole-slice rule. The restart costs the first chunk's fold twice; a missed deadline costs the whole row.
- A single-chunk fold has nothing to probe and is never cut by time; a zero or absent budget is no budget; the restart happens at most once.
- The row says which budget cut it: `meta.truncation_reason` is `"rows"`, `"time"`, or `null` when the window was folded whole. `meta.stream_chunks` counts the chunks of the fold that produced the row. The table output names the time budget in its truncation line.

## What does not change

- Without a time budget, nothing: the chunk plan, the event-budget cut, the stamps and every statistic are as before (the `"rows"` reason is the old `window_truncated`).
- No BPF change, no schema change, nothing on the recording side; no new dependency.
- The JSON gains one key (`truncation_reason`) and changes no existing value.

## How it was verified

- The existing chunking test is unchanged (a chunked fold equals the single query). The existing event-budget test's comparison of the cut run against the explicit half-window run clears the new `truncation_reason` stamp beside `window_truncated` before comparing, and asserts the cut run read `rows` — one existing test's normalisation extended, its values unchanged (the event budget still cuts to exactly the explicit half-window run).
- A new test drives a nine-chunk fold under a budget nothing fits — the window collapses to the same sliver an event budget of one row keeps, stamped `time`, one chunk, exact over that sliver — then a budget everything fits (the result identical to the unbudgeted fold, stamped nothing, nine chunks), a single-chunk fold under the same tiny budget (never cut), a zero budget (no budget), an event-budget cut keeping its own name, and the stamp's JSON spelling (`"time"`, `null`).
- First contact after a release, on hosts whose captures reach a caller's deadline today: the caller's timeout count going to about zero while rows stamped `time` appear where the timeouts were, read per host class beside the fold's own `aggregate_ms`.

Stacked on #269 (the loss-shape fields, the first commit of this branch); the second commit is this change. Base: `main` at v1.20.6 (`a1f76bc`). Three files: `src/analyze/sched_aggregate.rs`, `src/analyze/mod.rs`, `src/bin/systing_analyze.rs`.

## Risk review

Run on the final diff (this change's commit: three files). Each answer is tagged verified (read in the diff, the source or a CI log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** none. A new optional flag and one add-only JSON key (`truncation_reason`, `null` when the window was folded whole); a revert removes both and changes nothing else — verified in the diff; no data is deleted or migrated, no existing key is renamed or re-typed.
- **Compat changes:** yes, everything existing behaves as before: without the flag the chunk plan, the event-budget cut, the stamps and every statistic are unchanged — verified by the existing chunking and event-budget tests, whose values are untouched (the `"rows"` reason names the old `window_truncated` cut beside it, not instead of it); a caller opts in by passing the flag; no default changes — verified in the diff.
- **Tradeoffs:** a time cut costs the first chunk's fold twice (the probe, then the restart over the shorter window) and leaves a fifth of the remaining budget unspent by design (`TIME_CUT_FILL`, the rate's error margin) — verified in the diff; a single-chunk fold has nothing to probe and is never cut by time, so a caller whose whole window is one chunk keeps the deadline risk it had — verified in the diff; the probe reads this host's rate from one chunk, so a rate that changes mid-fold can still run past the budget (inferred; the restart happens at most once by design). Why a restart and not a stop at a chunk boundary: the window rule is whole-slice containment and a slice is charged whole at its start, so a fold stopped mid-way would not equal the window an explicit end time gives — the restart keeps one definition of a truncated window — verified at the source.
- **Blast radius:** `systing-analyze sched aggregate` callers that pass `--time-budget`; nothing that records, nothing in BPF, nothing on a traced host — verified in the diff.
- **Rollback:** revert the pull request (its own commit; #269 beneath it is independent); the next aggregate run folds as before — verified: no state carries across runs.
- **How verified:** the new nine-chunk test and the existing tests, Clippy, in CI's Test Suite (green at the head), and the review pass posted as a comment — verified in the CI runs.
- **Not checked:** the probe's accuracy on real hosts of different fold rates (the first contact after a release — a caller's timeouts going to about zero while rows stamped `time` appear — is the check, its bar in "How it was verified"); the budget's fit under a caller's own timeout is the caller's to set (assumed: the caller passes a budget under its deadline).


josefbacik:
### rust-bpf-code-reviewer verdict — `analyze: cut the sched aggregate's window to a wall-clock budget`

Executed in-session by the author over the full diff (three files — `src/analyze/sched_aggregate.rs` +246/−25, `src/analyze/mod.rs` +2/−2, `src/bin/systing_analyze.rs` +19/−3, git's own counts; the final diff's sha256 `b1edb5bf43b845e97753c48784f4f95869a591cc541f7637c7b16021a0b3e6d2`, the second commit of the branch, on top of #269) against the repository's `rust-bpf-code-reviewer` definition, all five passes; the one Improvement it raised was taken before the commit, and the one CI finding on the first push (an existing test's comparison not clearing the new stamp) was folded into the same commit.

### Summary
A third guard on the bounded fold: a wall-clock budget that probes the host's fold rate with the first chunk, re-cuts the window with the existing bucket-boundary event-budget cut when the rest would overrun, and restarts the fold over the shorter window — so the truncated row keeps the exact, whole-slice semantics the event budget already has, and the row names which budget cut it. No BPF, no schema, no new dependency; without a budget the code path is the old one.

### Critical Issues
None. No BPF code; no DuckDB schema change; no `Cargo.toml` version edit; no recorder, Python-version or ELF-parsing change.

### Improvements
1. **(Taken before the commit)** The fill fraction was a bare `0.8` in `time_cut`; it is now the documented constant `TIME_CUT_FILL` beside `DEFAULT_MAX_ROWS`, stating what the held-back fifth pays for.
2. The restart's rate estimate uses the chunk plan's nominal rows per chunk (`stream_rows / chunk_count`) rather than the rows the first chunk actually returned. The plan cuts by row count on a 4,096-bucket histogram, so the two agree to within one bucket's events; counting the fetched rows would need a return value from `fold_stream` for a precision the fifth already covers. Left as is, and said in the doc comment.

### Suggestions
- The projection runs once, after the first chunk, and the restart at most once (`restarted`); a fold whose rate changes sharply after the probe is not re-cut — the caller's own timeout still bounds it, as today. Stated in the module docs.
- `plan_chunks` is a closure because the same sizing runs before the fold and again after a restart; a method would read the same and cost a longer signature.

### Positive Notes
- The time cut reuses `budget_cut` rather than inventing a second cut, so the exactness argument in the module docs ("every stream row of the shorter window … at or before that boundary") covers both reasons with one proof.
- `restart_seed` clones the thread tables only when a budget is set; a fold without one pays nothing new.
- The test pins the deterministic corners — a budget nothing fits collapses to exactly the event budget's one-row sliver (the same `window_end_ns` the event-budget test asserts), a roomy budget reproduces the unbudgeted fold bit for bit, a single chunk is never cut, zero is no budget, and the stamp's serialization — without depending on wall-clock timing.
- The table output's truncation line now says which budget cut the window and by how much, in the same shape as the event-budget line.

PR #272, opened 2026-09-16
pystacks: gate on the main interpreter's GIL, not the newest one's

josefbacik:
## What this changes

On CPython 3.13 and later the GIL lives in the interpreter state, and discovery derives the addresses of its `locked` and `last_holder` words by following a pointer out of `_PyRuntime.interpreters`. It followed `head`. That list is newest first, and since 3.12 a subinterpreter can own its own GIL, so a process with a subinterpreter alive had its GIL gate keyed on the subinterpreter, not on the main interpreter running the workload. The read predates #268, which left it alone as a pre-existing assumption of discovery while its own thread-name reader walked to the main interpreter. Discovery now follows `interpreters.main`, the field after `head`, the way that reader does: the bindings for the version carry both offsets, and their distance is applied to the head offset the process data holds. Versions without an object table (before 3.13) keep reading `head`, the one interpreter they have.

On a process with a single interpreter `main` and `head` point at the same interpreter, so nothing changes there.

## What it does NOT change

- No BPF change, no frame-grammar change: no BPF program or map, no shared struct (`OffsetConfig` keeps `py_runtime_state_interpreters_head` as it was), no frame format, no output column.
- The read, the null check and the GIL-field derivation are the previous code; only the address read moves by the compiled head→main distance.

## How it was verified

- A new unit test pins `main_interpreter_offset` against the generated 3.13 and 3.14 bindings (640 and 816), the step applied to a process-reported head offset (1000 → 1008), and the fall-through for a version without a table (3.12 keeps head).
- The existing discovery tests are unchanged.
- Runtime first contact, where a workload runs a subinterpreter beside its main one on 3.13+: the Python frames that were gated out by the wrong GIL come back — the share of samples in that process with Python frames, before and after.

Base: `main` at v1.21.0 (`f8574ae8`). One file: `src/pystacks/discovery.rs`.

## Risk review

Run on the final diff (one commit, one file). Each answer is tagged verified (read in the diff, the source or a CI log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** none. User-space discovery only: no BPF program or map, no shared struct, no frame format, no output column, no flag — verified in the diff; a revert restores the `head` read whole, and nothing the walker derives is persisted beyond a capture — verified.
- **Compat changes:** NO, in one respect — on a CPython 3.13+ process with a subinterpreter alive, the GIL gate keys on the main interpreter's lock instead of the newest interpreter's, so the main interpreter's threads walk where they were gated out, and a subinterpreter's threads holding only their own GIL are gated out where they walked — verified at the source (`interpreters.head` is the newest interpreter; `main` the field after it; a subinterpreter may own its own GIL since 3.12); the effect on a real process is inferred (no such process in the tests). On a single-interpreter process `main` and `head` are the same interpreter and nothing changes — verified at the source; before 3.13 the read is unchanged — verified in the diff. WHY: the main interpreter runs the workload, and a gate keyed on the newest interpreter was an accident of the list's order, not a choice; no default, format or stored data changes, and versions coexist (every capture is self-contained).
- **Tradeoffs:** none in cost — the same one read, at an offset moved by a compile-time distance — verified in the diff; the distance is taken from the version's generated bindings, so a version whose bindings lack `main` keeps `head` (the fall-through) — verified by the unit test.
- **Blast radius:** discovery on CPython 3.13+ processes that run more than one interpreter; every other process reads the same words as before — verified at the source; how common such processes are among a user's Python workloads is not known here (inferred: rare).
- **Rollback:** revert the pull request; the next capture discovers as before — verified: no state carries across captures.
- **How verified:** the new unit test (the 3.13 / 3.14 offsets 640 / 816, the step 1000 → 1008, the 3.12 fall-through) and the existing discovery tests, Clippy, in CI's Test Suite (green at the head), and the review pass posted as a comment — verified in the CI runs; the interpreter list's order and the per-interpreter GIL at CPython's `pycore_runtime_structs.h` — verified at the source.
- **Not checked:** a live 3.13+ process with a subinterpreter (the first capture of one after the change is the check, its bar in "How it was verified"); a free-threaded build's interpreter layout (assumed the same fields at the bindings' offsets).


josefbacik:
### rust-bpf-code-reviewer verdict — `pystacks: gate on the main interpreter's GIL, not the newest one's`

Executed in-session by the author over the full diff (one file — `src/pystacks/discovery.rs` +43/−4, git's own count; the diff's sha256 `7f45a67dfdd058c63040d99015af2405cb8200a3bc8330f0f3775274023891e0`) against the repository's `rust-bpf-code-reviewer` definition, all five passes; the one Improvement it raised was taken before the commit.

### Summary
A one-hunk correctness fix in discovery's 3.13+ GIL-address derivation: the interpreter whose GIL the walker gates on becomes `_PyRuntime.interpreters.main` instead of `interpreters.head` (the newest interpreter of a newest-first list), derived from the head offset the process data holds plus the head→main distance the version's bindings carry — the same rule the thread-name reader added in the previous change uses. No BPF, no schema, no new dependency; versions without an object table keep the old read.

### Critical Issues
None. No schema change, no `Cargo.toml` version edit, no BPF object change, no ELF parsing.

### Improvements
- Taken before commit: the helper's doc comment asserted a source for the head offset (a process-reported table) that discovery does not read today; re-cut to state only the invariant it relies on — the distance between the two fields is the step, applied to whatever head offset the process data holds.

### Suggestions
- None. A `match` on the `Option` is the right shape for the two arms; the helper is small and named for what it answers; the unit test pins the derived offsets against the generated bindings rather than literals, so a regenerated binding moves the expectation with it.

### Positive Notes
- The change touches only the address arithmetic and leaves the read, the null check and the GIL field derivation as they were, so the behaviour on a process with one interpreter is byte-for-byte the previous one (the same pointer sits at `main` and `head` there).
- The fall-through for versions without a table keeps `head`, matching the single interpreter those versions have, rather than refusing the GIL read.

PR #273, opened 2026-09-17
memory: skip 32-bit compat syscalls in the raw-tracepoint pair

josefbacik:
## Description for Humans

**Origin:** a review finding on the opt-in raw-tracepoint form of the memory recorder's syscall hooks (#247): under that form, a 32-bit compat task's syscalls were recorded as if they were 64-bit mmap / munmap / brk calls. This change fixes that finding.

**What.** This change makes systing's memory recorder, under its opt-in `--kernel-hooks raw-tracepoint` form, skip the syscalls of 32-bit compat tasks — what the default `classic` form already does — and adds one row to the BPF load-shape gate in CI so the raw form is also loaded under the sampled-map setting. It touches three source files (the BPF object, the load-shape table, one test) and two documents.

**Why.** The raw `sys_enter` / `sys_exit` tracepoints fire for a compat task with the 32-bit syscall table's number, so a 32-bit `link`, `execve` or `chdir` (numbers 9, 11 and 12) was recorded as an mmap, munmap or brk row with meaningless addresses and sizes. The classic form never recorded them, because the kernel filters compat syscalls out of the classic events.

**If it is not done.** Any deployment that opts into the raw form on a host running 32-bit compat programs writes bogus mmap / munmap / brk rows into its memory tables, and its figures for those hosts disagree with the classic form's.

**What it touches.** Only the two raw-tracepoint BPF programs, which load only when a configuration selects the raw form, plus the CI gate's table and one test. The default `classic` form, its programs, every existing row, table and file format are unchanged. No shipped configuration selects the raw form today.

**Is it safe.** Backward compatibility: yes for every existing deployment, because the default form is untouched; under the raw form, compat tasks' rows disappear, which is the intended change. One-way door: none; a revert restores the five files and nothing persists. Verified: CI green at this head on all three kernel rows, and an end-to-end test on a KVM guest that passes with the fix and fails without it (201 bogus rows with the unfixed pair, 0 with the fix). Not checked: the arm64 and riscv compat paths end to end, and a run on a production host. Therefore it is safe to land at the maintainer's discretion; nothing is exposed until a deployment opts into the raw form.

## What

The opt-in `--kernel-hooks raw-tracepoint` form (#247) dispatches on the syscall number at the raw `sys_enter` / `sys_exit` tracepoints. Those tracepoints fire for a 32-bit compat task with the COMPAT table's number, while the classic `syscalls/sys_{enter,exit}_*` events skip compat syscalls in the kernel (`trace_get_syscall_nr` returns -1 under `ARCH_TRACE_IGNORE_COMPAT_SYSCALLS` — x86's `arch_trace_is_compat_syscall()` is `in_32bit_syscall()`, arm64's and riscv's `is_compat_task()`). So under the raw form an ia32 `link` (9) / `execve` (11) / `chdir` (12) was recorded as an mmap / munmap / brk row with the x86-64 argument registers read as addr / size, and on arm64 a compat `setgid32` (214) / `setfsuid32` (215) as a brk / munmap row. No shipped configuration selects the raw form, so no capture produced by the default form carried such rows.

Three changes, no schema change:

1. **The compat early-return.** One helper, `systing_in_compat_syscall(task, regs)`, called in both programs right after the syscall-number test (the entry cost of every other syscall is unchanged) and before the task filter:
   - x86: `TS_COMPAT` (0x0002) in `thread_info.status` — set by `syscall_32_enter()` before `sys_enter` fires and cleared by `arch_exit_to_user_mode_prepare()` after `sys_exit` has fired (arch/x86/entry/common.c, arch/x86/include/asm/entry-common.h, kernel/entry/common.c at v6.12), so both programs read the same bit; a kernel built without `IA32_EMULATION` never sets it. x32 needs no test: its numbers carry `__X32_SYSCALL_BIT` and never pass the number test.
   - arm64: `TIF_32BIT` (bit 22) in `thread_info.flags`, the flag `is_compat_task()` reads.
   - riscv: the ISA's `sstatus.UXL` field of the saved `regs->status`, which `start_thread()` sets to 32 for a compat task (arch/riscv/kernel/process.c) — used instead of riscv's `TIF_32BIT`, whose bit number moved between kernel series (11 at v6.6 and v6.12, 16 at v6.18). `TS_COMPAT` and arm64's `TIF_32BIT` are the same at v6.6, v6.12 and v6.18.
   The compiled x86 programs read, after the number test and the task pointer: one CO-RE direct load of `thread_info.status`, `w1 &= 0x2`, one branch — before the `trace_task` call.
2. **The `memory-raw-tracepoint-97` load-shape row.** The gate loaded the pair only at `memory_map_sample_rate = 1`, so the `> 1` branch the exit helpers inline into `systing_sys_exit` was dead rodata on that row; a tp_btf program the verifier rejects fails the whole object at load with no attach-time fallback. The twin row (memory + sample rates 97 + the raw form) loads that branch, as `memory-sample-rates-97` does for the classic exit programs; the shape-table test asserts the row.
3. **Wording.** "with no grace period waited on the stop path" was exact only when the pair is the tracepoint's last probe: `tracepoint_remove_func`'s 2->1 branch may run `tp_rcu_cond_sync` (a conditional classic-RCU + SRCU wait, never tasks-trace; kernel/tracepoint.c, the same at v6.6, v6.12 and v6.18). Re-worded in the BPF block comment, the two Rust docs and SCHEMA_CHANGES.md; the "rows byte-identical whichever form" sentences now say compat tasks are skipped under every form. SCHEMA_CHANGES.md also stamps schema 24 as 1.21.0 and the `raw_tracepoint` value as 1.20.0.

## Test

`test_memory_syscall_hooks_raw_tracepoint_skips_compat_tasks` (`tests/memory_recorder.rs`; x86-64 only, `#[ignore]`, needs root and BPF). The workload is a 64-bit python3 issuing ia32 syscalls through `int 0x80` — the kernel sets `TS_COMPAT` for that entry exactly as for a 32-bit binary, so no 32-bit toolchain is needed — from a 9-byte stub in a `MAP_32BIT` executable mapping called through ctypes. One ia32 `chdir(".")` (12 = x86-64 `brk`) + `link(src, fresh name)` (9 = x86-64 `mmap`) pair writes the probe file, then 200 rounds of the same pair, every call SUCCEEDING: the exit helpers already drop a failed call, so only a compat call returning 0 makes the unfixed pair emit — a brk row with addr 0 (and a delta of minus the heap address) and an mmap row with addr 0, which no real process produces. Per form (the raw pair, then the classic set): `memory_syscall_leg` is the form's, the pid's real mmap / munmap rows are present (the positive twin), and the compat-shaped rows are 0. It skips, printing why, on a kernel without ia32 emulation (the untraced probe run: SIGSEGV on `int 0x80`, or the calls not returning 0/0) or whose BTF lacks the pair's typedefs.

## Risk review

- **One-way doors:** none. A revert restores the three source files and the two docs; the change adds no map, no program section, no flag, no schema field and no data path (verified: the diff, 5 files, +353/-13).
- **Compat changes:** with the fix unused — the default `classic` form, which every shipped configuration selects — every existing caller, config and row is exactly as before: the compat test lives only in the two `tp_btf` programs, which load only under `--kernel-hooks raw-tracepoint` (verified: `Config::default()` keeps `KernelHooks::Classic`; the classic programs are untouched by the diff). Under the raw form, a compat task's syscalls stop producing mmap / munmap / brk rows — the intended change, and the classic form's existing behaviour, so a reader of both forms now sees the same rows (verified: the e2e below, 0 / 0 on both forms at the head). Old and new binaries coexist: each carries its own object; no wire, file or table format moves.
- **Tradeoffs:** one CO-RE load + one AND + one branch per matching syscall entry / exit in the raw programs, after the number test, so no cost to any other syscall (verified: the x86 disassembly of `tp_btf/sys_enter` / `tp_btf/sys_exit`). The riscv test reads an ISA register field instead of a kernel flag whose bit moved between series — one line of arch-specific code more, no runtime cost (verified: the diff). The load-shape gate gains one row (about a second per kernel row in CI; verified: the CI timings below).
- **Blast radius:** captures under the raw form only — selected by no shipped configuration (inferred: the deployments this author knows pass `--kernel-hooks` as `classic` or `trampoline`; a future opt-in of the raw form is where this fix matters). The CI gate's new row runs on every push to every PR.
- **Rollback:** revert the one commit; nothing written anywhere persists.
- **How verified:**
  - CI at this head (`4e51db9f`): Rustfmt, Clippy (`-D warnings`) and the Test Suite green (run 35183768758); `bpf-load-shapes.yml` green on all three kernel rows — vmtest-6.12, cos-125 6.12.85, cos-129 6.12.90 (run 35183768745): `every_shape_loads` 37 loaded / 0 rejected with the release-compiler object, against 36 / 0 at the base v1.21.0 (run 35131283436) — the one added row is `memory-raw-tracepoint-97`, `loaded=true programs selected=21` on each kernel (verified: the job logs).
  - The end-to-end test, RUN on an x86-64 KVM guest (nixpkgs kernel 6.12.104 with BTF, `CONFIG_IA32_EMULATION=y`, `CONFIG_COMPAT=y`, no `ia32_emulation=` on the cmdline; glibc 2.42; CPython 3.13.15) from a build of the v1.21.0 tree plus this PR's diff applied byte for byte: the probe `returned 0/0`; `every_shape_loads` 37 / 37; the test PASSES — `raw-compat: memory_syscall_leg = raw_tracepoint; 53 big mmap + 52 munmap rows; compat-shaped rows: 0 brk at 0, 0 mmap at 0`, `classic-compat: memory_syscall_leg = tracepoint; 53 big mmap + 52 munmap rows; 0 brk at 0, 0 mmap at 0` (verified: the run's whole output, exit 0).
  - The NEGATIVE CONTROL, run the same way with this PR's diff MINUS the BPF hunk (the test and the load-shape row on the unfixed pair): `every_shape_loads` 37 / 37, and the test FAILS at `raw-compat` — `254 big mmap + 52 munmap rows; compat-shaped rows: 201 brk at 0, 201 mmap at 0` (the probe pair plus the 200 rounds), the assertion `left: (201, 201) right: (0, 0)` — so the unfixed pair records a compat task's `chdir` / `link` as brk / mmap rows on that kernel and the fixed pair does not (verified: the run's whole output, exit 101).
  - A datum from the same guest, not a claim of this PR: `stop-phase: detach bpf programs` read 14 ms under the raw pair and 158 ms under the classic set on the 2-CPU guest at the head.
  - The aarch64 object (the `__TARGET_ARCH_arm64` branch of the compat test, `thread_info.flags & (1UL << 22)`) LOADS at all 37 shapes on an aarch64 6.12.104 + BTF guest under TCG, built from the same tree (verified: the run's `every_shape_loads` output).
  - The BPF object compiled standalone for x86, arm64 and riscv at `build.rs`'s pinned flags (`-mcpu=v3 -fwrapv`), rc 0 each; the x86 disassembly of both programs shows the compat test as described above (verified).
  - The reviewer pass (the repository's `rust-bpf-code-reviewer` definition over the full diff): no Critical issues; its one Improvement (a clippy `manual_range_contains` shape in the test's probe assertion) applied before the commit; the verdict is the first comment on this PR (verified).
  - Pairing: no map or program added; `systing_in_compat_syscall` writes nothing (direct reads of `task_struct` / `pt_regs` fields on BTF-typed pointers, the shape `systing_syscall_nr` already uses); `LRU_HASH` is used by no program of `systing_system.bpf.c` (verified).
- **Not checked:** the arm64 and riscv compat branches END TO END — no compat-capable arm64 or riscv guest ran the workload; their constants are read at the kernel headers of v6.6, v6.12 and v6.18 and the arm64 object's load at every shape is the proof available; the production kernels' `CONFIG_IA32_EMULATION` / `CONFIG_COMPAT` (COS 6.12.85's `CONFIG_IA32_EMULATION` is y by its published config; the arm64 kernels' `CONFIG_COMPAT` unread); a run on a production host — the raw form is selected by no shipped configuration, so no host exercises the fix until one opts in.
- **Since the plan:** the plan named the riscv test as `TIF_32BIT`; the diff reads `sstatus.UXL` instead, because riscv's bit number differs between v6.12 and v6.18 (found at the header read). The plan expected the e2e to be run later on a rig; it ran on a KVM guest before this READY, head and negative control both.

## Checks

- `cargo fmt --all -- --check` clean.
- The BPF object compiled standalone for x86, arm64 and riscv at `build.rs`'s pinned flags (`-mcpu=v3 -fwrapv`), rc 0 each; the x86 disassembly of `tp_btf/sys_enter` / `tp_btf/sys_exit` shows the compat test as described above.
- The reviewer pass (the repository's `rust-bpf-code-reviewer` definition, executed over the full diff): no Critical issues; its one Improvement (a clippy `manual_range_contains` shape in the test's probe assertion) applied before the commit; the verdict is the first comment on this PR.
- Pairing: no map or program added; `systing_in_compat_syscall` writes nothing (direct reads of `task_struct` / `pt_regs` fields on BTF-typed pointers, the shape `systing_syscall_nr` already uses); `LRU_HASH` is used by no program of `systing_system.bpf.c`.
- This PR's CI is the compile, unit-test and load proof: `ci.yml` (fmt, clippy with `-D warnings`, `cargo test --workspace --no-default-features`, with clang 21 as the BPF compiler) and the three `bpf-load-shapes.yml` kernel rows (`every_shape_loads` at the head — the new row loads the pair's sampled branch on each), all green at `4e51db9f`.


josefbacik:
## rust-bpf-code-reviewer pass — the raw-tracepoint pair's compat early-return, the `memory-raw-tracepoint-97` load-shape row, the detach wording

Mechanics: the repository's `.claude/agents/rust-bpf-code-reviewer.md` definition was executed by the author in-session over the full diff (five files, +353/−13), passes 1–5 in the definition's order, before the commit. The one Improvement it produced was applied before the commit (below). No clippy or `cargo test` ran where the review was written; the pull request's own CI carries them (`cargo fmt --all -- --check`, `cargo clippy --workspace --no-default-features -- -D warnings`, `cargo test --workspace --no-default-features`, and the three `bpf-load-shapes.yml` kernel rows).

### Summary

The change makes the opt-in `--kernel-hooks raw-tracepoint` pair skip a 32-bit compat task's syscalls the way the classic `syscalls/*` events already do in the kernel, adds the load-shape row that loads the pair's sampled branch, and corrects the detach wording in four places. The BPF change is one inlined helper (one direct BTF-pointer field load, one mask, one branch) placed after the syscall-number test in both programs, so every other syscall's entry cost is unchanged; no map, program or ring-buffer write is added. The new e2e is `#[ignore]` and x86-64 only, with explicit skip conditions and a documented negative control.

### Critical Issues

None. Schema: no table or column changes, `SCHEMA_VERSION` untouched, `Cargo.toml` untouched; the two `SCHEMA_CHANGES.md` edits stamp a release on an existing heading and an inline note and add prose to an existing entry. Recorder identity: no new per-event table; the test's queries join through `thread` / `process` (`UTIDS_FOR_PID`), never raw pid columns. Binary parsing: none.

### Improvements

1. Applied: `tests/memory_recorder.rs`, the probe's base-address check was written `base > 0 && base < (1 << 31)`, the shape clippy's `manual_range_contains` rejects under `-D warnings`; now `(1..(1i64 << 31)).contains(&base)`.

### Suggestions

1. `systing_in_compat_syscall` takes both `task` and `regs` so one signature serves the three architectures (x86 and arm64 read `thread_info`, riscv reads the saved `sstatus`); the unused parameter per architecture is intentional and costs nothing after inlining. If a fourth architecture ever joins the raw pair, the `#else` arm returns `false` — a new architecture must add its own test rather than inherit "never compat".
2. The riscv test reads the ISA-defined `sstatus.UXL` field instead of `TIF_32BIT` because that flag's bit number moved between kernel series (11 at v6.6 and v6.12, 16 at v6.18); the x86 (`TS_COMPAT` = 0x0002) and arm64 (`TIF_32BIT` = 22) values are the same at v6.6, v6.12 and v6.18, so they are spelled as constants with their header named. A future kernel that moves either is a change to those two lines; the comment says where the values come from.

### Positive Notes

- The compat test sits after the syscall-number test and before `trace_task`, so the pair's hot path for the other ~440 syscalls is byte-identical to the previous build, and a compat call of one of the three numbers returns before any map lookup.
- The x86 timing (`TS_COMPAT` set in `syscall_32_enter()` before `sys_enter` fires, cleared in `arch_exit_to_user_mode_prepare()` after `sys_exit` has fired) is stated from the entry code, not assumed, so the enter and exit programs read the same bit.
- The e2e's workload needs no 32-bit toolchain: a 64-bit python3 issuing ia32 syscalls through `int 0x80` is a compat task for the duration of each call, and the test requires those calls to SUCCEED (return 0) because the recorder's exit helpers already drop a failed call — the rows a misread would leave are exactly brk-at-0 / mmap-at-0, which no real process produces; the positive twin (the pid's real mmap/munmap rows) is asserted first, and the skip conditions (no ia32 emulation, no BTF typedefs) print their reason instead of passing vacuously.
- The `memory-raw-tracepoint-97` row closes a real gap: a tp_btf program the verifier rejects fails the whole object at load with no attach-time fallback, so the pair's sampled branch is now loaded by the gate rather than first on a host.

PR #274, opened 2026-09-17
pystacks: skip entry frames by the interpreter's own mark and count them

josefbacik:
## What this changes

From CPython 3.12 the interpreter pushes a frame of its own onto the frame chain at every entry into `_PyEval_EvalFrameDefault` (every C → Python re-entry); it carries no code object. Since #246 the BPF walker visited those entry frames like any other — a code-object read, a name read and a slot of the 127-symbol budget each — and user space told them apart by NAME (`<interpreter trampoline>` on 3.12; `[Frame Error]` with a computed instruction index on 3.13+), a rule that depends on the word after `_Py_NoneStruct` at the qualname's offset not being a pointer, which nothing in CPython promises.

The walker now reads the interpreter's own mark, `_PyInterpreterFrame.owner`, before the code object and steps over an entry frame (`owner >= 3`: `FRAME_OWNED_BY_CSTACK` on 3.12/3.13, `FRAME_OWNED_BY_INTERPRETER` / `FRAME_OWNED_BY_CSTACK` on 3.14) without reading names or spending a slot; it counts each one on the symbol emitted just inward of it, in that symbol's spare word (`stack_walker_frame.pad_`, zero until now): a frame's `pad_` is the number of entry frames between it and the next emitted frame outward. The walk is bounded on emitted symbols again (a visited-frame bound of twice the budget stays for the verifier), the shape #249 had before #246 took it back.

User space puts the markers back from the counts — in the root-first list a frame's entry frames stand in front of it — so `interleave_python_frames` and everything that reads `PythonFrame::entry` are unchanged; the name rule is gone. The count rides with the frame through the skeleton conversion, the stack spill (16 bytes per Python frame instead of 12) and the stack's identity (`PyAddr` compares and hashes it, so the same Python frames reached through different C re-entries are different stacks, as they interleave differently).

## What consumers see

The frame strings and the output columns are unchanged; the markers never appear in a stack. Three things move at the margins of `stack.frame_names`:

- Deep Python stacks lose fewer frames to the budget: an entry frame no longer costs a slot, so a stack that was cut at the root end where many re-entries stacked up (a workload heavy in callbacks from C) now keeps its outermost frames.
- A real frame whose qualname read failed but whose instruction index was computed renders as `[Frame Error] (python) [unknown]` where the name rule used to take it for an entry frame and drop it: such frames become visible, and a stack that was cut into runs by that misreading is one run again.
- On a 3.13/3.14 build where the word after `_Py_NoneStruct` is a pointer, an entry frame rendered as a Python frame with a name read from wherever it pointed; it now renders as nothing, as on every other build.

An entry frame ahead of the first emitted symbol (the interpreter is on the entry frame alone for a few instructions as the outermost frame of an entry returns) is not counted: before, it produced a marker that opened an empty run and erased the innermost loop frame from the stack; now that stack keeps its native frames whole. A walk that meets only entry frames emits no Python frames, as before.

## How it was verified

- Unit tests on the reconstruction (`with_entry_markers`: markers in front of the frame that counted them, a count of zero, two back to back, a walk that stopped short of its outermost entry frame, a negative word, the empty list, a count past the walk's bound), on the stack spill round trip with a non-zero count, and on the count being part of a frame's identity; the interleave's own tests are unchanged and still pass on the same `entry` marks.
- The BPF object: the repository's load-shape gate (`bpf-load-shapes.yml`) builds the release object and loads every shape on the CI kernels; the walker diff is the #249 shape plus a bounded write into the message buffer (the count's index bounded on the register that indexes, the verifier's requirement).
- Runtime first contact, on a 3.12+ workload heavy in C → Python re-entries: the share of samples whose Python walk hit the 127-symbol budget falls back to what it was before #246, the `[Frame Error]` share stays at the level of real read failures (it may rise slightly where the name rule was hiding them), and no frame named from `_Py_NoneStruct`'s neighbour appears.

Base: `main` at v1.21.0 (`f8574ae8`). Five files: `src/pystacks/bpf/pystacks.bpf.c`, `src/pystacks/bpf/include/stack_walker.h`, `src/pystacks/stack_walker.rs`, `src/pystacks/types.rs`, `src/stack_recorder.rs`.

## Risk review

Run on the final diff (five commits). Each answer is tagged verified (read in the diff, the source or a CI log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** none. A revert restores the previous walk and the name-based filter whole — verified: the diff changes no schema, table, column, file format, wire format or command-line flag; the emitted frame record keeps its layout (`stack_walker_frame`'s `pad_` word existed at the same offset and width and was always zero), and nothing the walker writes is persisted beyond a capture's own output.
- **Compat changes:** NO, in one respect — the rendered Python stacks are not exactly as before. The frame record's spare word now carries a count: a spill written before this change reads as zero markers (the previous behaviour) — verified by the unit tests on `with_entry_markers`; a reader without this change ignores the word, as it did — verified in the previous code. But the stacks move at the margins listed under "What consumers see" — deep stacks keep more frames, a real frame's `[Frame Error]` is no longer hidden by the name rule, and an entry frame never renders under a borrowed name — so a consumer that counts `[Frame Error]` frames or measures Python stack depth sees its figures move — verified at the source; how far each share moves on a given host depends on which of its reads fault, which is not a property of this change (inferred). WHY: the previous behaviour is the defect this change removes — entry frames were told apart by a name rule that rests on an unpromised memory layout and that hides real read failures — and the interpreter's own mark is the discriminator CPython provides; no stored data, file format, default or flag changes, and old and new versions coexist (every capture is self-contained), so the change is right as cut.
- **Tradeoffs:** one `owner` byte read per visited frame in BPF (the #249 cost) in exchange for no code-object or name read on an entry frame and no budget slot spent on it — verified in the diff; the count's clamp (`2 × BPF_LIB_MAX_STACK_DEPTH − 1`) bounds what a corrupt spill record can expand to — verified by the unit test. No new dependency, map, program or allocation.
- **Blast radius:** every capture with Python stack collection on a CPython 3.12+ process; the `owner` read is gated on the interpreter version, so 3.11 and older walk exactly as before — verified in the diff. The BPF object must still pass the verifier on every kernel the object is loaded on — verified on the three CI kernels by the load-shape gate (36 of 36 shapes at the head, after two earlier heads were rejected for an unbounded index and fixed).
- **Rollback:** revert the pull request; the next capture runs the previous walker — verified: no state carries across captures.
- **How verified:** the unit tests and Clippy in CI's Test Suite, the load-shape gate on cos-125 / cos-129 / vmtest-6.12 (green at the head), the review pass posted as comments — verified in the CI runs. The two design claims a run cannot show — that `owner` is `FRAME_OWNED_BY_CSTACK` on every entry frame and that a walk visits at most twice the symbol budget — rest on CPython's `ceval.c` and the walker's loop bound (verified at the source).
- **Not checked:** a live capture with this object on a running interpreter (the first capture after the change is the check, its bars in "How it was verified"); a free-threaded 3.14 build, whose `owner` sits at another offset the offsets table does not carry — such a build walks as before this change (assumed, from the table).


josefbacik:
### rust-bpf-code-reviewer verdict — `pystacks: skip entry frames by the interpreter's own mark and count them`

Executed in-session by the author over the full diff (five files — `src/pystacks/bpf/pystacks.bpf.c` +87/−6, `src/pystacks/bpf/include/stack_walker.h` +3/−0, `src/pystacks/stack_walker.rs` +146/−72, `src/pystacks/types.rs` +4/−1, `src/stack_recorder.rs` +23/−9, git's own numstat; the diff's sha256 `336d89667fc2000191294118af12f44cc0f9b1de41dc193f1f1d31180d821a26`) against the repository's `rust-bpf-code-reviewer` definition, all five passes; the one Improvement it raised was taken before the commit.

### Summary
The BPF walker reads each CPython 3.12+ frame's `_PyInterpreterFrame.owner` byte before its code object and steps over the interpreter's own entry frames (`owner >= 3`) without reading names or spending a slot of the stack budget, counting each on the symbol emitted just inward of it in that symbol's spare word (`stack_walker_frame.pad_`); the walk is bounded on emitted symbols again, with a visited-frame bound of twice the budget for the verifier. User space rebuilds the entry markers from the counts (a frame's markers stand in front of it in the root-first list), so `interleave_python_frames` and every consumer of `PythonFrame::entry` are unchanged; the name-based entry rule (`<interpreter trampoline>`, `[Frame Error]` with an instruction index) and its dependence on what follows `_Py_NoneStruct` go away. The count travels through the frame's `From` conversion, the spill format (16 bytes per Python frame) and the stack's identity (`PyAddr` compares and hashes it).

### Critical Issues
None. No map, program, section or attach form is added or removed; no schema change; no `Cargo.toml` edit; no new dependency. The BPF diff is the shape that shipped in v1.18.5–v1.19.0 with one new helper of three lines (`note_entry_frame`: a bounded index into the message buffer, the same `< BPF_LIB_MAX_STACK_DEPTH` check `add_symbol_to_buffer` uses).

### Improvements
- Taken before commit: the marker expansion pushed a struct literal inside a counted loop; re-cut as `repeat_with(entry_marker).take(n)` over a `usize::try_from(...).unwrap_or(0)` count, so a negative word reads as no entry frames and the lint surface is clean.

### Suggestions
- None. The reconstruction is a pure function with its own tests (in front of the frame, zero, two back to back, the stopped-short walk, the negative word, the empty list); the spill round trip carries a frame with a non-zero count; the identity test pins that the count distinguishes two otherwise equal frames.

### Positive Notes
- The `owner` read is gated the way the walker already gates every version-dependent offset (`PyVersion_major/minor` and the offset not being the default), so 3.11 and older read nothing new and carry a count of 0.
- The one behavioural change at the leaf end is stated in the code: an entry frame ahead of the first emitted symbol (the interpreter between two Python calls) is dropped rather than counted, where the name rule used to emit a marker that opened an empty run — a stack in that window now keeps its innermost loop frame instead of erasing it.
- Public text carries no environment names; the commit carries the `Co-authored-by` trailer only.

josefbacik:
### rust-bpf-code-reviewer verdict — second commit (the count taken in the frame step)

Executed in-session by the author over the full diff against `main` (both commits, five files); the first verdict above covers the first commit and stands.

**What changed since the first verdict.** `note_entry_frame` moves out of the walk loop into `pystacks_get_frame_data`, called where `frame_skipped` is set; it looks the message up itself (`pystacks_get_msg`, the per-CPU heap the walker already writes from the same programs). The loop in `walk_and_load_py_stack` is byte for byte the one that shipped in v1.18.5–v1.19.0 (`if (!state->frame_skipped) add_symbol_to_buffer(py_msg);`); the emitted-symbol bound and the visited-frame bound of twice the budget are unchanged. The count's meaning is unchanged: the frame step runs before the loop body decides whether to emit, so the symbol counted on is still the one emitted last, the frame just inward of the stepped-over entry frame.

**Critical:** none. No map, program, section or attach form is added or removed; no schema change; no `Cargo.toml` edit. The message-buffer write keeps its `< BPF_LIB_MAX_STACK_DEPTH` bound (the same check `add_symbol_to_buffer` uses); the lookup's NULL return is handled before the write. The user-space side and the tests are untouched by this commit.

**Improvements:** none taken.

**Nits:** none.

**Mechanics:** the diff read as `git diff main..HEAD` in the branch's worktree; `rustfmt --check` bare on the three Rust files (unchanged by this commit); clippy and the crate's tests are the repository's CI runs on this head, as is the load-shape gate for the BPF object.

josefbacik:
### rust-bpf-code-reviewer verdict — third commit (the count's index bounded on the register that indexes)

Executed in-session by the author over the full diff against `main` (three commits, five files); the two verdicts above cover the first two commits and stand.

**What changed since the second verdict.** `note_entry_frame` computes `last = stack_len - 1`, makes it opaque with `barrier_var` and bounds it (`last >= BPF_LIB_MAX_STACK_DEPTH` returns) before the write, so the check and the buffer offset share one register; the separate zero test is gone, since a `stack_len` of zero wraps `last` to the largest value and fails the same check. Nothing else moves.

**Why.** The load-shape gate's verifier log at the previous head read `math between map_value pointer and register with unbounded min value is not allowed` at this write: the bound had been proved on `stack_len - 1` while the compiler formed the offset from `stack_len` (`r1 <<= 4; r0 += r1` on the unbounded register). The pattern is the one the tree already uses for its other bounded indices (`barrier_var` ahead of the bound).

**Critical:** none. No map, program, section or attach form is added or removed; no schema change; no `Cargo.toml` edit. The write stays inside the message buffer (`last` ≤ 126 by the check, on the register used to index). What this verdict holds is read from the source; the verifier's own verdict on this head comes from the load-shape gate loading the release-flag object with every shape, as for the previous heads.

**Improvements:** none taken.

**Nits:** none.

**Mechanics:** the diff read as `git diff main..HEAD` in the branch's worktree; the previous head's verifier log read from the gate's guest output; `rustfmt --check` bare on the three Rust files (unchanged by this commit); clippy and the crate's tests are the repository's CI runs on this head, as is the load-shape gate for the BPF object.

josefbacik:
### rust-bpf-code-reviewer verdict — fourth commit (the spilled count clamped to the walk's bound)

Executed in-session by the author over the full diff against `main` (four commits, five files); the three verdicts above stand for the first three commits.

**What changed since the third verdict.** User space only: `with_entry_markers` clamps a frame's entry-frame count to `MAX_ENTRY_FRAMES_PER_SYMBOL` (twice the symbol budget less one, the most one walk can count on a symbol) before expanding it into markers, and one test drives the expansion with `i32::MAX` and with the bound itself. The BPF object is untouched by this commit.

**Why.** The count reaches `with_entry_markers` from the walk and from the stack spill; the spill is a file, so a corrupt record could carry a word up to `i32::MAX` and cost that many marker allocations. The walk itself never counts past its visit budget, so the clamp changes nothing for a count the walk produced.

**Critical:** none. No map, program, section or attach form is added or removed; no schema change; no `Cargo.toml` edit; the spill format is unchanged (the clamp is at the read, not the write).

**Improvements:** none taken.

**Nits:** none.

**Mechanics:** the diff read as `git diff main..HEAD` in the branch's worktree; `rustfmt --check` bare on `stack_walker.rs`; clippy and the crate's tests are the repository's CI runs on this head; the BPF object's load-shape rows stand from the previous head, which this commit does not change.

josefbacik:
### rust-bpf-code-reviewer verdict — fifth commit (the clamp test's names read through the module's helper)

Executed in-session by the author over the full diff against `main` (five commits, five files); the four verdicts above stand.

**What changed since the fourth verdict.** Test code only: `a_count_past_the_walks_bound_is_clamped_to_it` reads the expanded list through the module's `shape` helper (entry markers as `ENTRY`, a frame by its symbol) and asserts on those names, as the sibling tests do; its fourth-commit form compared the last frame's display name (`f (python) [app.py:1]`) against the bare symbol and failed on that comparison alone in the previous run. The clamp and the production code are unchanged.

**Critical:** none. No BPF byte, no schema, no `Cargo.toml` edit.

**Improvements:** none taken.

**Nits:** none.

**Mechanics:** the diff read as `git diff main..HEAD` in the branch's worktree; the previous run's test log read for the failing assertion; `rustfmt --check` bare on `stack_walker.rs`; clippy and the crate's tests are the repository's CI runs on this head; the BPF object's load-shape rows stand from the previous heads, which this commit does not change.

PR #275, opened 2026-09-17
pystacks: witness glibc's thread descriptor before reading through it

josefbacik:
## What this changes

The walker finds a thread's `PyThreadState` through the pthread TLS slot, which it reaches from the thread's glibc descriptor at hard-coded, head-anchored offsets: on x86-64 the descriptor is the thread pointer itself (`fsbase`; `specific_1stblock` at 0x310), on aarch64 / riscv64 it is derived from the clear-child-tid pointer glibc registers with the kernel (`&pthread->tid` at 0xd0). Nothing checked that the thread was laid out that way: a C library with another layout (musl keeps `tid` and the robust list 0x58 apart where glibc keeps them 0x10 apart), or a glibc that moved the fields, was read through the wrong offsets and yielded a garbage thread state.

Both architectures now derive the descriptor in one place (`get_glibc_pthread_descriptor`) and trust it only when the two pointers glibc registers with the kernel — the clear-child-tid pointer (`&pthread->tid`) and the robust-list pointer (`&pthread->robust_head`) — sit at the descriptor's expected offsets. A thread that fails the check reads no slot and reports no thread state, with a new `pthread_id_match` value (`PYSTACKS_PTHREAD_ID_NO_DESCRIPTOR`) saying why; the check reads `task_struct` fields only, never user memory, so it costs nothing on a thread it refuses.

Three diagnostics the walker already computed but could never observe — the global thread state (`_PyThreadState_Current`, 3.11 and older), the descriptor's `header.self` pointer on x86-64, and the pthread that bound the thread state (`PyThreadState.thread`) — were read with `bpf_probe_read_kernel` at user addresses, which fails on both architectures. They are user reads now, through the same helper as every other read of the traced process, so `pthread_id_match` is a live witness: MATCH means the descriptor derivation and the TLS-slot read landed on this thread's own structures (`pthread_self()` of the binding thread is the descriptor's address on glibc). The walker's no-frames returns hand the message header over, so a sample of a thread without frames carries the witness too, and the sampler resets the witness byte beside the frame count for a process the walker did not run on.

With `--pystacks-debug`, the walker's statistics print one line per process of how the witness read on its samples (`match=… mismatch=… no_descriptor=…`), which is what tells a thread without a thread state from a descriptor that was not glibc's.

## What consumers see

Nothing in the frame strings or the output columns changes. On a process whose C library is not glibc, or whose glibc lays the descriptor out otherwise than assumed, the walker used to read through wrong offsets and could emit garbage or `[Frame Error]` frames; it now emits nothing for that thread. The message header of a no-frames sample is copied into the ring event where it was left unset before; no consumer read those bytes.

## How it was verified

- Unit tests on the debug tally (the kinds render in a fixed order, only the kinds seen; the tally is per process; a value past the enum counts as `unknown`).
- The BPF object: the repository's load-shape gate (`bpf-load-shapes.yml`) builds the release object and loads every shape on the CI kernels; the diff adds no program or map — two `task_struct` reads and two compares in the descriptor helper, three reads switched from the kernel helper to the user-task helper, and a header-only copy on the no-frames path.
- Live captures of this head against a control built from the same base without this change, on QEMU guests built from nix (kernel 6.12.104 with BTF, glibc 2.42): x86-64 under KVM and aarch64 under TCG. On each guest, for both builds: `every_shape_loads` 36 of 36; `systing --duration 5 --sw-event --collect-pystacks --pystacks-debug` against a CPython 3.13.15 hot loop and against a workload three C→Python re-entries deep (and, on x86-64, the hot loop under CPython 3.14.7). The witness line read `match` on every sampled glibc thread and no other kind (x86-64: 5,001 / 5,001 / 5,001 samples; aarch64: 4,525 and 4,540 — the first observation of the arm64 robust-list derivation); the share of samples carrying Python frames was the control's on every capture (100 % of the workload's samples on both builds) and `[Frame Error]` was 0 on both. A musl 1.2.6 CPython 3.14.7 process beside a glibc one in the same x86-64 guest read `no_descriptor` on every one of its 4,937 samples and yielded no Python frames; the control yielded none from it either on this guest, so there the change names the reason rather than changing the outcome, and the glibc process beside it was unchanged.

Base: `main` at v1.21.2 (`0da3774e`) — rebased after #274 merged; one test-module block resolved (the retired name-rule test not re-added), no shared code hunk. Five files: `src/pystacks/bpf/include/pthread_helpers.bpf.c`, `src/pystacks/bpf/pystacks.bpf.c`, `src/pystacks/bpf/pystacks.bpf.h`, `src/bpf/systing_system.bpf.c`, `src/pystacks/stack_walker.rs`.

## Risk review

Run on the final diff (one commit). Each answer is tagged verified (read in the diff, the source or a CI log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** none. A revert restores the unchecked descriptor reads whole — verified: the diff changes no schema, table, column, file format, wire format or command-line flag; the new `pthread_id_match` value is one more member of an enum that lives only in the walker's per-run statistics, which are printed and never persisted; the ring event's header bytes were already part of the event's layout (they are now written where they were left unset).
- **Compat changes:** NO, in one respect — on a C library with another layout (musl) a thread now yields no Python frames where it yielded frames read through wrong offsets — verified at musl's own `pthread_impl.h` figures (the refusal is by construction: `tid` and the robust list sit 0x58 apart there against glibc's 0x10), and observed in the musl capture above (`no_descriptor` on every sample); on glibc, a thread whose descriptor passes the witness walks exactly as before — verified: the descriptor address is the one the previous code used, and the witness only refuses; observed on both architectures (the captures above). The native frames of such a thread are unchanged; no default changes; `--pystacks-debug` only adds one line per process to what it already printed. WHY: the frames that disappear were wrong by construction — the thread state was read from a garbage address, so the names and lines came from whatever memory sat there — and no reader can want them; old and new versions coexist (every capture is self-contained), so the change is right as cut.
- **Tradeoffs:** two `task_struct` reads and two compares per walk before any user read, in exchange for never reading a thread state through the wrong layout; the three diagnostics become user reads that can fault like any other, which is the point (they were failing silently) — verified in the diff; their cost is that of the reads the walker already makes (inferred, not measured; the captures' sample counts moved within run noise between the two builds).
- **Blast radius:** every capture with Python stack collection, on every architecture the walker supports (x86-64, aarch64, riscv64) — verified in the diff; on aarch64 the tid equation holds by construction (the descriptor is derived from the tid pointer), so the robust-list equation alone is the witness there — verified at the source and observed in the aarch64 captures above. The BPF object must pass the verifier — verified on the three CI kernels by the load-shape gate (every shape at the head) and on the two guest kernels above.
- **Rollback:** revert the pull request; the next capture runs the previous walker — verified: no state carries across captures.
- **How verified:** the unit tests and Clippy in CI's Test Suite, the load-shape gate on cos-125 / cos-129 / vmtest-6.12 (green at the rebased head), the review pass posted as a comment — verified in the CI runs; the glibc and musl offsets at their own headers (`nptl/descr.h`, `pthread_impl.h`) — verified at the source; the live captures above against a control build on x86-64 (KVM) and aarch64 (TCG) guests — verified in the guests' own logs (the load reports, the witness lines and the per-capture query tables).
- **Not checked:** a capture on real aarch64 hardware (the arm64 read above is under TCG emulation on a 6.12 guest kernel — the same derivation, not the same silicon); a musl process on aarch64, and a musl CPython 3.13 (only a 3.14 build was at hand); a glibc other than 2.42, or one laid out otherwise (the refusal path was exercised by musl only); the cost of the two added `task_struct` reads per walk (not measured); kernels beyond the CI and guest kernels — the first captures after the change on other kernels are the check, their bars in "How it was verified".


josefbacik:
### rust-bpf-code-reviewer verdict — `pystacks: witness glibc's thread descriptor before reading through it`

Executed in-session by the author over the full diff (five files: `src/pystacks/bpf/include/pthread_helpers.bpf.c`, `src/pystacks/bpf/pystacks.bpf.c`, `src/pystacks/bpf/pystacks.bpf.h`, `src/bpf/systing_system.bpf.c`, `src/pystacks/stack_walker.rs`).

**Summary of the change.** The glibc descriptor derivation moves into one helper for both architecture families (`get_glibc_pthread_descriptor`: x86-64 from the thread pointer, aarch64 / riscv64 from the clear-child-tid pointer) and is trusted only when the task's registered clear-child-tid and robust-list pointers sit at the descriptor's expected head-anchored offsets (`GLIBC_PTHREAD_{TID,ROBUST_HEAD,SPECIFIC_1STBLOCK}_OFFSET`: 0x2d0 / 0x2e0 / 0x310 on x86-64, 0xd0 / 0xe0 / 0x110 on the TLS-variant-1 architectures — the same numbers the old code hard-coded piecemeal); a refused descriptor yields no TLS-slot read, no thread state, and the new `PYSTACKS_PTHREAD_ID_NO_DESCRIPTOR` witness value. The three diagnostic reads at user addresses (`_PyThreadState_Current`, `header.self`, `PyThreadState.thread`) go through `bpf_probe_read_user_task` instead of `bpf_probe_read_kernel`; the global-state read is skipped when the process data carries no address. The walker's two no-frames returns hand back the header's length so `pystacks_read_stacks` copies the header alone (its `offsetof(struct pystacks_message, buffer)` bytes) into the ring event, and the sampler resets `pthread_id_match` beside `stack_len` in both event reserves. User space, in debug mode only, tallies the witness byte per process and prints one line per process with the statistics.

**Critical:** none. No map, program, section or attach form is added or removed; no schema change; no `Cargo.toml` edit; no new dependency. Every new memory access is bounded: the descriptor helper reads `task_struct` fields through `BPF_PROBE_READ` and compares scalars; the three converted reads are fixed 8-byte reads through the existing helper (which falls back to `bpf_probe_read_user` for a NULL task, as every other read of the target does); the header-only copy is a constant-size `bpf_probe_read_kernel` from the walker's own per-CPU heap into the ring event. The task-stacks iterator reads `stack_len` after a positive return and still returns 0 for a frameless message, so its behaviour is unchanged by the no-frames return value.

**Improvements:**
- Taken before commit: the `pthread_id_match` byte was reset only where the walker ran; the reset sits in both event reserves now, so a sample of a non-Python process cannot carry a stale witness from an earlier event in the same ring slot.
- Taken before commit: an eighth witness value (`NO_DESCRIPTOR`) instead of folding a refused descriptor into `THREAD_STATE_NULL`, so the debug tally tells a thread without a thread state from a descriptor that was not glibc's.

**Nits:**
- On the TLS-variant-1 architectures the clear-child-tid comparison in the helper is an identity by construction (the candidate is derived from that pointer); it is kept so both arms run the same two checks, and the robust-list comparison is the witness there.

**Mechanics:** the diff read as `git diff` of the staged tree against `main` at v1.21.0 in the branch's worktree; the glibc offsets checked against `struct pthread`'s generic-nptl layout and the musl figures against musl's `pthread_impl.h`; `rustfmt --check` bare on `stack_walker.rs`; clippy and the crate's tests are the repository's CI runs on this head, as is the load-shape gate for the BPF object.

PR #276, opened 2026-09-19
pystacks: walk free-threaded CPython 3.14 (python3.14t)

josefbacik:
Origin: a free-threaded CPython 3.14 (`python3.14t`, a `--disable-gil` build) traced with `--collect-pystacks` yielded no Python frame names — `<unknown python>` markers, and from v1.21.3 mostly `[Frame Error]` — where the default build of the same version resolves names and lines — because the walker picks its offsets table by version alone while the free-threaded build is a different ABI, and because from 3.14 on the frame's code word is a tagged pointer that the walker dereferenced unmasked; a table without the mask reads no names, and the mask with the wrong table reads wrong ones, so the two land together.

## What this changes

A free-threaded build is a different ABI from the default one. Its object header is 32 bytes, so every field behind a header sits 16 bytes further on (24 in a code object, past the `co_tlbc` that only this build has), and its interpreter frame has `tlbc_index` ahead of `owner` (+74 → +78). Walked with the default 3.14 offsets, such a process yields no names.

From 3.14 on a frame's `f_executable` is a `_PyStackRef`: a pointer with a tag in its low bits. The free-threaded build tags every code object (deferred reference counting), the default build the immortal ones (the interpreter's `__init__` shim frame carries one today). `get_code_ptr` returned the raw word as the code pointer; for 3.14 and later it now clears the low two bits first (`Py_TAG_BITS`: 3 in the default build, 1 in the free-threaded one; an object pointer's own low two bits are always zero), inside the existing 3.11-and-later arm — the one change to the BPF programs.

Which build a binary is comes from the binary itself. From 3.13 on `_PyRuntime` starts with a statically initialized `_Py_DebugOffsets` — the cookie `xdebugpy`, the version (`PY_VERSION_HEX`) and a `free_threaded` word — so those bytes are in the ELF file that discovery already opens to find `_PyRuntime` and the version. They are read there once per binary through the same section-to-file-offset lookup the cache-tag read uses (the `object` crate's section API, no hand-rolled ELF walk), cached beside the version and the runtime address, and believed only behind the cookie, a version word of the same `major.minor`, and a value of 0 or 1. No process's memory is read for it: the change asks nothing new of the host — no ptrace access, no capability.

The offsets are then chosen by version and build: `for_build(major, minor, free_threaded)`, with `for_version` as the default build's wrapper (every existing caller unchanged); `py314t()` is `py314()`'s field list over a new generated module, `bindings/v3_14_7t.rs`. A free-threaded build of a version with no table of its own gets none, never the nearest one. Discovery fails closed where a wrong name is possible: a 3.14-or-later binary whose runtime table cannot be trusted is not walked, and neither is a free-threaded 3.13 (or 3.15 and later, until its table is generated) — each said once per binary on stderr, the process then yielding native frames only. A 3.13-or-older default build keeps today's path whatever its file says: its frames hold plain pointers, so nothing there changes.

`scripts/generate_python_bindings.py` gains `--free-threaded` (configures with `--disable-gil`, writes the module with a `t` suffix, refuses a version before 3.13) and `--installed <prefix>` (the same offset program compiled against an installed interpreter's headers, for where there is no CPython checkout; it refuses a tag that disagrees with the headers' own `patchlevel.h`), and each route refuses the other build's headers at compile time. Every 3.13-and-later module it writes from now on also carries where a `_Py_DebugOffsets` keeps its cookie, version and `free_threaded` word; a committed module gains those three lines when it is next regenerated. The test harness follows the repository's own rule for a new interpreter: `scripts/setup-pystacks.sh` installs pyenv's `3.14.6t` beside `3.14.6` (a definition pyenv ships: `PYTHON_BUILD_FREE_THREADING=1` over the `3.14.6` one), `scripts/run-integration-tests.sh` reads that entry out of the list (its version pattern allows the `t` suffix now), and `tests/trace_validation.rs` gains `PYTHON_314T_VERSION` and `test_pystacks_python314t`, its pyenv lookup carrying the `t` from the version to the binary's name.

## What consumers see

On a free-threaded 3.14 process, Python frames with their function names where there were only `<unknown python>` markers or `[Frame Error]`; the frame strings' format is unchanged. On a default 3.14 build, the one frame class whose code word carries the tag — an immortal code object, the interpreter's `__init__` shim frame — is read through the cleared pointer now instead of an odd one (verified at the source; no such frame was sampled in the captures below, so the effect was not observed). Every 3.13-and-older read path is behaviour-identical (see the risk review for what the object diff shows). A 3.14-or-later process whose binary does not say which build it is, and a free-threaded 3.13, go from marker or `[Frame Error]` frames to no Python frames at all.

One thing this does not do: on a free-threaded build a frame's `instr_ptr` points into the bytecode copy of the thread running it (`co_tlbc->entries[tlbc_index]`; only index 0 is `co_code_adaptive` itself), and the walker still takes the instruction index against `co_code_adaptive`, so a frame running a thread's own copy (`tlbc_index != 0`, a worker thread) gets no line, or the function's first line — its NAME is right. The cure needs three more offsets and reads and grows `OffsetConfig`, so it is its own change.

## How it was verified

- Unit tests: the offsets tests pin the free-threaded layout facts (`ob_type` 24, `owner` 78, `co_code_adaptive` 232, the fields `py314` sets all set, no fallback for the flavour) and the `_Py_DebugOffsets` head at 0 in every generated module; the discovery tests pin the parse — a wrong cookie, an older, newer or other-major version word, a value other than 0 or 1, a short table, each refused — the join of what a binary says to its table or the reason there is none, and a built-ELF fixture in four flavours (free-threaded, default, another cookie, another version) taken through `try_python_module` itself, so the table a process gets is checked at the call site (putting the call site back on the default table fails it). 83 of 83 in the crate's pystacks tests (`cargo test --lib -- pystacks::`, 74 at the base plus 9 new).
- The generator, two controls: `--installed` on a default 3.14.7 against the committed `v3_14_0.rs` differs in the header lines, the three new constants and three `PyInterpreterState` compounds (+24 each, CPython's own growth between 3.14.0 and 3.14.7 — no object, frame, thread-state, code-object or runtime constant moves), identical under gcc 13 and 15; the checkout route on the 3.14.7 source against `--installed` on the built interpreters, both flavours: every constant identical.
- Discovery run live (no BPF) on child processes of the three interpreters: `python3.14t` reads "Python 3.14.0 (free-threaded)" with `ob_type` 24 / `owner` 78, the same under `PYTHON_GIL=1`; the default 3.14.7 and 3.13.15 read their unchanged lines; each binary parsed once.
- The BPF object: the repository's load-shape gate (`bpf-load-shapes.yml`) builds the release object and loads every shape on the CI kernels; on QEMU guests built from nix (kernel 6.12.109 with BTF, 2 vCPUs, 4 GiB), `every_shape_loads` 36 of 36 on x86-64 under KVM and 36 of 36 on aarch64 under TCG, for this head and for a control built from the same base without it, with the same program count on every shape (10 on `continuous-pystacks`, 75 on `all-recorders`). The shapes load the main object; the task-stacks recorder's sleepable copy of the walker (`task_stacks.bpf.o`) carries the same change and is built and compared below, but no shape loads it (see "Not checked").
- Live captures on those guests, this head against the control, `systing --duration 5 --sw-event --collect-pystacks --pystacks-debug`: (b) a hot loop under `python3.14t`, the default 3.14.7 and 3.13.15 in one guest — on x86-64 the free-threaded process went from 3,195 samples with `[Frame Error]` on every one and no name (the control) to 3,201 of 3,201 carrying `spin_hot_loop_m4` at `spin.py:5` and `:6` (2,753 / 448, beside the default 3.14's 2,891 / 431 on the same lines) with `[Frame Error]` 0, and discovery's line reading "Python 3.14.0 (free-threaded)"; the two default processes at 100 % of their samples with Python frames on both arms, within run noise (3,244 / 3,322 and 3,314 / 3,310 samples); on aarch64 the same shape — the free-threaded process from 1,980 samples all `[Frame Error]` (the control) to 1,730 of 1,730 named (`spin.py:5` 1,544 / `:6` 186, `[Frame Error]` 0), the two default processes at 100 % on both arms (1,818 / 1,934 and 1,887 / 1,836 samples); (c) the same loop on a worker thread of the free-threaded process — the control `[Frame Error]` on all 5,000 samples, this head the whole chain named (`Thread._bootstrap`, `_bootstrap_inner`, `run`, the worker and the loop) on 5,002 of 5,002 with no line number on any frame and `[Frame Error]` 0 — the limit above, observed; (d) `python3.14t` with `PYTHON_GIL=1` (the process reports the GIL enabled) — the control `[Frame Error]` on all 5,004, this head names and lines on 5,001 of 5,001, as (b): the layout is the build's, not the GIL's.
- The object diff between the control and this head, built with the release compiler (clang 21), the same on both architectures: in each of the three objects that carry the walker, `pystacks_get_frame_data` is the only walker function that differs — it grows from 421 to 431 instructions in `pystacks.bpf.o` (and the linked object) and from 923 to 930 in `task_stacks.bpf.o`, with the same added instruction texts on both architectures and no instruction that exists only in the control; maps, rodata, bss and every program section are equal (in `task_stacks.bpf.o` one 64-bit immediate of `task_stacks_read_user` moves with the address of a local callback the function before it displaced; the callback's bytes are identical). The standalone walker gains one `bpf_probe_read_user` call site, which is a second copy of the existing read of the code word — the 3.11-and-later arm gets its own so the compare and the mask can follow it — and every path still executes exactly that one read: a 3.11 to 3.13 frame runs two instructions more and one jump fewer (the load of the minor version from the walker's own per-CPU state and the compare; no new read of the process, no store), a 3.10-or-older frame and a shadow-frame walk run the same instructions as the control.
- `test_pystacks_python314t` was added per the harness's own rule and not run in this venue (no pyenv here); the captures above are its check.
- The same pair on this head's own tree — v1.21.4 (the base) against v1.21.4 plus this change, x86-64 under KVM: every shape loads on both (36 of 36, the same program count per shape as at v1.21.3); the three-spinner capture reads as before (the free-threaded process from 2,101 samples all `[Frame Error]` on the base to 3,445 of 3,445 named at `spin.py:5` / `:6` (2,865 / 580) with this change; the two default processes at 100 % on both); the base's thread-descriptor witness line reads `match` equal to the sample count on every process of both arms (`no_descriptor` 0, the free-threaded process included); and one capture with the task-stacks recorder on (`--add-recorder task-stacks --task-stacks-frames python`), which loads the sleepable copy of the walker: exit 0 on both arms, 150 snapshots (50 per process, one stack on each), the free-threaded process's 50 all `[Frame Error]` on the base and all named (`spin.py:5` 43 / `:6` 7 under `<module>` at `:12`) with this change, the two default processes the same on both. Thread names read NULL for every process on both arms: the workload imports no `threading`, so there are none to read — the free-threaded thread-name behaviour stays unchecked.

Base: `main` at v1.21.4 (`e0aa865b`); one commit, rebased from v1.21.3 (`b4908540`, the base the guest runs used) with no conflict — the base's own change to `pystacks.bpf.c` between the two tags sits from line 1063 on, the mask's hunk at 661. Nine files: `scripts/generate_python_bindings.py`, `scripts/run-integration-tests.sh`, `scripts/setup-pystacks.sh`, `src/pystacks/bindings/mod.rs`, `src/pystacks/bindings/v3_14_7t.rs` (new, generated), `src/pystacks/bpf/pystacks.bpf.c`, `src/pystacks/discovery.rs`, `src/pystacks/offsets.rs`, `tests/trace_validation.rs`.

## Risk review

Run on the plan and again on the final diff (one commit). Each answer is tagged verified (read in the diff, the source or a log), inferred (follows from a read but was not exercised) or assumed.

- **One-way doors:** none — verified: the diff changes no map, program, rodata field, wire format, trace schema, table, column, file format or command-line flag, and `OffsetConfig`'s layout stays as it is. A revert restores today's behaviour whole.
- **Compat changes:** NO, in two places, both intended. (1) On a default 3.14 build the interpreter's `__init__` shim frame carries a tagged code word today, read as an odd pointer; with the mask it is read right — inferred from CPython's own header (the default build tags an immortal object's reference in bit 0, and `_PyFrame_GetCode` clears the same bits before every use); not observed in a capture (no such frame was sampled in the runs above). WHY: the cleared word is the pointer CPython itself uses; an odd value is never a valid object address, so clearing the bits can turn a miss into a name and never one name into another. (2) A 3.14-or-later process whose binary's runtime table cannot be trusted, and every free-threaded 3.13, go from marker or `[Frame Error]` frames to no Python frames (native frames only), said once per binary on stderr — verified in the diff and the fixture tests. The answer is kept per binary for the session as the existing parse cache keeps it, so a read error on those 24 bytes leaves that binary unwalked for the session — the limit the existing parse already has for an error in the middle of it; under `--pystacks-debug` the caller's existing "libpython/_PyRuntime not found" line also prints for such a process and then reads misleadingly (in a file outside this change; a follow-up). WHY: for such a process the alternative is a walk on a table that cannot be trusted — for a free-threaded 3.13, today's default table reads its 32-byte-header objects at the wrong offsets, the reads that gave wrong names in the free-threaded 3.14 emulation, only untagged — so not walking removes a wrong-name path. Everything else is a yes: every 3.13-and-older read path is behaviour-identical (the object diff above — verified), a default build's stderr is byte-identical to today's (verified in the live discovery run), and old and new versions coexist since every capture is self-contained.
- **Tradeoffs:** at discovery, one 24-byte read of the file per binary at its one parse (the table's cookie, version word and flavour word; a binary without the table is refused at the cookie) and nothing per process — verified in the diff, and measured on a default 3.13 library: 3,491,304 bytes read against 3,491,280 before; in the walker, one AND per frame on 3.14 and later and one compare on 3.11 to 3.13 — verified in the object diff; one more generated module of about 3.6 KB.
- **Blast radius:** the mask reaches every 3.14-or-later process; the table and the not-walked rules reach free-threaded ones only; the BPF object changes, so every load shape is verified again on both architectures (the load-shape gate and the guest loads above). The read of the file's bytes rests on the file discovery opens being the one mapped, the assumption the runtime address and the version already rest on (a replaced file reads "(deleted)" in the maps and is not discovered today either); a binary that controls its own `.PyRuntime` bytes can only make itself not walked or walked with its own build's table, never fault the walker outside it — the BPF side's reads stay bounded by `OffsetConfig` as before (verified in the diff).
- **Rollback:** revert the pull request; the next capture runs the previous walker — verified: no state carries across captures. The mask and the table are never split: the mask with today's table read wrong bytes as names on a free-threaded process (1,316 of 1,316 code objects in the emulation).
- **How verified:** the unit tests and Clippy in CI's Test Suite and the load-shape gate on the CI kernels (green at this head), the review pass posted as a comment, the generator controls, the live discovery run, the guest loads and captures above against a control build on x86-64 (KVM) and aarch64 (TCG), and the same pair on this head's own tree (x86-64 under KVM, with the task-stacks capture) — each in its own log; the per-version integration test added here was not run (no pyenv in the venue), the captures being its check.
- **Not checked:** the task-stacks recorder's sleepable copy of the walker on aarch64 (loaded and run on x86-64 only, above; no load shape covers it — a pre-existing gap of the gate); the mask's effect on a default 3.14 build's `__init__` shim frame (inferred from CPython's header; no such frame sampled); line numbers off a free-threaded process's first thread (known wrong, out of this change; capture (c) observes them); a free-threaded 3.13 (it gets no table and is not walked), and a 3.14 or later binary with an unreadable or mismatched runtime table (the fixture only); a non-PIE embedder whose executable carries a copy relocation of `_PyRuntime` (no table bytes in the executable's file: refused there, so no frames — never a wrong name; none at hand); other 3.14 patch levels (only 3.14.7 was at hand; the three compounds that drift feed a diagnostic byte and a fallback only); names with non-ASCII text on the free-threaded build; musl and the tail-calling interpreter; thread names of a free-threaded process (read as today: `object_offsets_for_build` answers none for that build, and nothing asks for one yet); two pre-existing windows in which a process keeps a stale per-pid table after an exec (the exec program's failed ring-buffer reservation; system-wide mode, which loads no fork or exec program) — a pid registered as a default 3.14 that later execs `python3.14t` would keep the default table, the same class as today's stale table after any exec, and its stale runtime and TLS addresses then normally find no thread state; one observation in the aarch64 captures that is the same on both arms and so not this change's — the `<module>` frame of a 3.14 process carries no line number there (the 3.13 process's does, and every process's does on x86-64), not traced; a capture on real aarch64 hardware (the arm64 read above is under TCG emulation); kernels beyond the CI and guest kernels — the first captures after the change on other kernels are the check.


josefbacik:
Review pass (the repository's reviewer checklist, run on the final diff, one commit on v1.21.4).

**Verdict: PASS — no critical issue, no blocking issue.** What was checked:

- Python version support (the checklist's four items): the new bindings module (`v3_14_7t.rs`, generated) comes with its interpreter in `scripts/setup-pystacks.sh` (`3.14.6t`, a definition pyenv ships) and its per-version test in `tests/trace_validation.rs` (`PYTHON_314T_VERSION`, `test_pystacks_python314t`); the two entries are identical; the integration-test runner's version pattern accepts the `t` suffix, so the missing interpreter triggers the setup. The default build's fallback arm in `for_version` is unchanged (no new default version), and the free-threaded flavour deliberately has no fallback — a table for another layout reads the wrong words.
- ELF parsing: the new read of the `_Py_DebugOffsets` head uses the `object` crate's section API (the same section-to-file-offset lookup the cache-tag read already used, lifted into `bytes_at` and shared); no hand-rolled header or magic constant.
- The BPF change: one hunk, `get_code_ptr`'s 3.14-and-later mask of the two tag bits, inside the existing 3.11-and-later arm; both walker objects carry it; every load shape loads on the three CI kernels and on the x86-64 (KVM) and aarch64 (TCG) guests, with the same program count per shape as the base; the per-function object diff shows `pystacks_get_frame_data` alone changed, growing by ten and seven instructions, nothing that exists only on the base.
- Fail-closed discovery: a 3.14-or-later binary whose runtime table cannot be trusted, and a free-threaded 3.13, are not walked; the fixture test drives the four cases through `try_python_module` itself, and the mutation that puts the call site back on the default table fails it.
- The observed effect: on both architectures a free-threaded 3.14 process goes from `[Frame Error]` on every sample to the hot loop's names and lines on every sample, the default 3.14 and 3.13 processes unchanged within run noise (the body's "How it was verified").

Notes, none a change request:
1. `DEBUG_OFFSETS_COOKIE_BYTES` in `discovery.rs` repeats the private cookie constant `thread_names.rs` keeps; sharing one would touch a seventh source file — a maintainer's preference.
2. The task-stacks recorder's sleepable copy of the walker (`task_stacks.bpf.o`) carries the mask and is built and compared, but no load shape loads it (a pre-existing gap of the gate); the body's "Not checked" names the capture that closes it.
3. Line numbers of a free-threaded worker thread (a per-thread bytecode copy) stay wrong by construction — names right, lines absent; the body sizes the follow-up.
4. For a process this change deliberately does not walk, the caller's existing `--pystacks-debug` line ("python-named exe … but libpython/_PyRuntime not found", in `systing_core.rs`, outside this diff) also prints and then reads misleadingly beside the new per-binary line; a wording follow-up.

